Issue #11, opened 2018-09-20
Implement Pagination Component


InnaAtanasova:
[Pagination](https://sap.github.io/fundamental/components/pagination.html)

InnaAtanasova:
Can we close this issue?

chrismanciero:
There is still the open question about how to handle the ... in the control. This thread discusses some of the issues (https://github.com/SAP/fundamental-react/pull/23) but I still have questions on how this control should work. 

Should this issue be closed and maybe a new one created on implementing the ... once the functionality is better defined?

Some more questions:
- Is there a minimum number of pages needed before displaying the ellipsis?
- If on a page and the next/previous item is ... when clicking next/previous does the ... expand to show more page numbers or just insert the next/previous page before/after the ellipsis. Example:
**1 2 3 ... 35**, after clicking previous does it go to **1 2 3 ... 34 35**
- If the component went with the solution from the question above (**1 2 3 ... 34 35**). If the previous button was hit 30 more times the pagination control will show all pages 1 through 35. How does the component choose where to insert the ... again?

PR #23, opened 2018-10-24
Pagination control

chrismanciero:
Initial commit of the Pagination control. This control will allow users to display a collection of links (page numbers) as well as a Previous and Next button. The total number of pages is determined by the total number of items.

This control is for the issue: https://github.com/SAP/fundamental-react/issues/11

### DIFF
--- a/src/Pagination/Pagination.Component.js
+++ b/src/Pagination/Pagination.Component.js
@@ -1,0 +1,146 @@
+import React from 'react';
+import {
+  DocsTile,
+  DocsText,
+  Separator,
+  Header,
+  Description,
+  Import,
+  Properties,
+  Pagination
+} from '../';
+
+export const PaginationComponent = () => {
+  const handleClick = event => {
+    console.log(`Page clicked - ${event}`);
+  };
+  return (
+    <div>
+      <Header>Pagination</Header>
+      <Description>
+        Pagination is commonly used for tables and tiles. It allows users to see
+        how many pages of content exist, to navigate and highlights which page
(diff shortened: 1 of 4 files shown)


droshev:
@chrismanciero first of all, thank for your contribution. It is very much appreciated 👍 
Regarding the implementation - I guess it is an initial implementation of the component as at the moment you show all the pages as link and no ellipsis when there are many pages per fundamental.
Are you willing to implement that part too?

chrismanciero:
@droshev - You're welcome and Yes this in an initial implementation. I decided to leave out the ellipsis in the control until I have a better understanding of how it should behave.

For example: https://sap.github.io/fundamental/components/pagination.html#in-between-more-than-three-pages 

- I noticed that the ellipsis it not clickable, should it be?

- What should happen when a user clicks on the ellipsis? Should the ellipsis behave like the Contextual Menu control (https://sap.github.io/fundamental/components/contextual-menu.html#with-icon) where when clicked it displays all the pages between

- When navigating using the Previous or Next button, if you go from a displayed number to the ellipsis, what happens?

droshev:
@chrismanciero let me come to you later. I have brought your questions  to the team. We can add you to the contributors so you won't need to fork - what do you think?

droshev:
@chrismanciero I discussed with @jeannewalters and @LeoT7508 your points. Here are the answers:

> No it should not. It should be a gray color instead of blue as it's not clickable.

> When there is an ellipsis, the selected page should always be in the 4th position

 The issue has been raised to @saadmhybris so they can improve it in fundamental-ui (the design and implementation)

Issue #12, opened 2018-09-20
Implement Modal Component 


InnaAtanasova:
[Modal](https://sap.github.io/fundamental/components/modal.html) Component

chrismanciero:
Going to commit a version of the Modal Component later today (10/30)

PR #41, opened 2018-10-30
Modal component

chrismanciero:
Component for https://github.com/SAP/fundamental-react/issues/12


src/Modal/Modal.js
```diff
@@ -0,0 +1,85 @@
...
+    document.addEventListener('keydown', this.handleKeyPress, false);
+  }
+
+  // remove event listener for escape key
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.handleKeyPress, false);
+  }
+
+  render() {
+    const { children, title, secondaryBtnText, primaryBtnText } = this.props;
```
InnaAtanasova: What do you think of adding 'actions' instead of 'secondaryBtnText' and 'primaryBtnText'? 
`const { children, title, actions } = this.props;`

src/Modal/Modal.js
```diff
@@ -0,0 +1,85 @@
...
+              <div className="fd-modal__header">
+                <h1 className="fd-modal__title">{title}</h1>
+                <button
+                  className="fd-button--secondary fd-modal__close"
+                  aria-label="close"
+                  onClick={this.handleCloseClick}
+                />
+              </div>
+              <div className="fd-modal__body">{children}</div>
+              {primaryBtnText ? (
```
InnaAtanasova: Then have something like:
`{actions ? (
                <footer className="fd-modal__footer">
                  <div className="fd-modal__actions">
                   {actions}
                  </div>
                </footer>
              ) : (
                ''
              )}`

This way we can allow the user to enter as many buttons as they wish. Also this way we won't take care of the actions emitted from the elements (buttons).

src/Modal/Modal.Component.js
```diff
@@ -0,0 +1,255 @@
...
+          user.
+        </Description>
+        <DocsTile>
+          <button onClick={this.showHideFormModal}>Show Form Modal</button>
+          {this.state.bShowFormModal ? (
+            <Modal
+              title="Invite user"
+              onClose={this.showHideFormModal}
+              primaryBtnText="Invite"
+              secondaryBtnText="Cancel"
```
InnaAtanasova: If we have 'actions' as props, we can then pass these actions like this:
`<Modal
              title="Invite user"
              onClose={this.showHideFormModal}
              actions={
                    <React.Fragment>
                         <Button type="secondary">Cancel</Button>
                         <Button>Invite</Button>
                      </React.Fragment>
                 }>`

src/Modal/Modal.js
chrismanciero: Are you thinking actions being an object with primary and secondary properties or an array?
actions = {primary: "Yes", secondary: "No" },  actions = ["Yes", "No"]

chrismanciero: Got it, sorry responded to the first comment without reading the others 😄 Going to change the control now to see how it goes

PR #57, opened 2018-11-13
Issue#4: implementation of Panel component

InnaAtanasova:


InnaAtanasova:
> I see 2 eslint warnings for Panel.Component.js
> _Line 11: 'Button' is defined but never used no-unused-vars Line 21: 'PanelContent' is defined but never used no-unused-vars_
> 
> Also I think it would be nice if you can add an example using different components inside PanelBody. What do you think ?

That's a great idea! I will do it now (also fix the warnings). Thanks!

PR #77, opened 2018-11-22
Feature/issue#3 time picker component

CodesOfRa:
Hello 👋 ,

Here's the time picker component :) or at least an initial version for it ( I am sure we could do more improvements)

_Question:_
I see something weird on the **fd-time** CSS class - it has a fixed width(184px) and it makes the smaller time formats to look weird in the pop-up. Any idea if this is a fundamental bug or is the expected behaviour ?

image
 
image

thanks
Ramona ⭐️


InnaAtanasova:
> f

@saadmhybris

InnaAtanasova:
> _Question:_
> I see something weird on the **fd-time** CSS class - it has a fixed width(184px) and it makes the smaller time formats to look weird in the pop-up. Any idea if this is a fundamental bug or is the expected behaviour ?

@saadmhybris

src/TimePicker/TimePicker.js
```diff
@@ -0,0 +1,477 @@
...
+  render() {
+    const {
+      id,
+      ...props
+    } = this.props;
+    const {popoverId, timeId} = this.state;
+    return (
+      <div id={id} className='fd-time-picker'>
+        <div className='fd-popover fd-popover--no-arrow'>
+          <Popover
```
InnaAtanasova: you can add `noArrow` to get rid of the arrow in the dropdown

src/TimePicker/TimePicker.Component.js
```diff
@@ -0,0 +1,102 @@
...
+} from '../';
+
+export const TimePickerComponent = () => {
+  const defaultTimeCode = `<TimePicker />`;
+  const format12hoursTimeCode = `<TimePicker format12Hours={true} />`;
+  const noSecondsTimeCode = `<TimePicker showSecond={false} />`;
+  const disabledStateTimeCode = `<TimePicker disabled={true} />`;
+  return (
+    <div>
+      <Header>Time</Header>
```
InnaAtanasova: Time Picker ;)

Issue #32, opened 2018-10-27
Typescript Support 


giovannidegani:
Any plans to use typescript or support it with typings ?

droshev:
@tiefox could you share your thoughts about using TS in React projects?

tiefox:
Sorry for late reply, but my thoughts are exactly the same as @o0Djeen0o . In my company we are looking to standardize all front-end development on Typescript, and currently looking to base our own internal component library for our internal enterprise apps on a good foundation, and we do already some Fiori + UI5, but also want to provide a standard to other teams working with React and Angular. Would be awesome if it was already in Typescript :)

droshev:
@o0Djeen0o @tiefox thanks for your feedback. I have asked few colleagues writing in React such as @casche. They use https://flow.org/. What are your thoughts for TS vs. Flow?

tiefox:
For me is mostly the traction that TS is getting vs Flow, the ecosystem is booming and we are making a bet on it. See this https://2018.stateofjs.com/javascript-flavors/conclusion/

droshev:
@tiefox I have seen that link. Both of them are close in terms of dev's satisfaction. I know that TS is more popular this year. This may flip quite fast.
Can we try to evaluate both and make a decision? What do you think if we (me, @InnaAtanasova, and @chrismanciero) we use Flow for a component and you use TS for the same component?
What do you think? If you agree which component would you propose?

o0Djeen0o:
Hi again @droshev,

I think that both solutions are valid for **static type checking**. As @tiefox mentionned, typescript has better support from the open source community, but Flow is also used and has Facebook behind, so both solutions valid here.

The real difference is that Typescript is not just about static typings. It can work like that (you have for instance a babel preset to just remove typings info, like for Flow), but it is more a language with:
* A compiler for your code, choosing which type of ES output you would like to address
* Since 3.0, it has project reference, helping to work with mono repos for instance, which is really useful for project with many components, like UI frameworks

I'll let you check what fits your needs best.

droshev:
@o0Djeen0o @tiefox   ok, we will discuss it and keep you posted

tiefox:
@droshev Sure! Lets figure out the right component and do this :)

chrismanciero:
Since the question is still out there as to how/if we should use TypeScript or Flow for our components,
[...]

PR #99, opened 2018-12-12
feat: extend functionality of the tabs component

maylukas:
I tried to use the Tabs component in a small demo / test app and found some aspects that might need changing:
* The tab sizes with the content which is due to rendering the content within the header item.
image
* The state of the tabs component can't be controlled externally and also there is no callback to observe any changes
* The component will always rerender since the tabs are passed as a list (the reference changes every time). If the tabs would be passed as children, the change detection of react would work
* Rendering non-text content fails
* The interface feels a bit non-react-idiomatic :) Passing the tabs as children with their respective content seems to fit better (maybe just personal taste).

I implemented another tabs component as a suggestion and would be happy if you could give feedback on this.

There are also some things I probably didn't get right:
* I probably didn't follow your style guide because I didn't find a reference to it :) If you let me know, I will fix my issues (maybe provide an editorconfig?).
* I used typescript since I saw the issue regarding a possible change of the implementation language to typescript [https://github.com/SAP/fundamental-react/issues/32](https://github.com/SAP/fundamental-react/issues/32) so I figured I could provide an example. If you want me to use javascript I can change that (switching to typescript / providing type definitions would be great though :) )
* The playground doesn't work / needs to be implemented as I didn't quite figure out what was required to control the props of children.
* The tabs component does not support overflow content as I wasn't sure how this should be done according to the Fiori 3 guidelines (is there an officially available reference?). It would be possible to overflow the tabs if the number of tabs exceeds a certain amount (e.g. passed as a prop).
* I didn't include the react-router integration as I am not sure if this is the right place for it. The choice of the routing framework (or a lack thereof) should be up to the downstream application. Let me know what you think about this.


maylukas:
Ah cool stuff! I wasn't aware that this is still in progress. I am going to rebase onto the typescript-conversion branch.

maylukas:
Should I create the pull request against the typescript-conversion branch? Or is it planned to merge the typescript branch into master any time soon?

CodesOfRa:
I think it's best to create a PR against typescript-conversion branch, there is still a debate regarding TypeScript vs Flow .

maylukas:
Okay. I changed the base.

PR #102, opened 2018-12-13
Typescript conversion

chrismanciero:
Just wanted to bring this branch to attention. I updated multiple components to use TypeScript as well as got the ball rolling for future components. All unit tests and coverage stays the same.


CodesOfRa:
awesome stuff @chrismanciero !!! 🎊 🎈 
I will start to convert some too :) pretty excited about this 😄 .

Do you have a list with the converted ones ?

chrismanciero:
@CodesOfRa List of converted components so far:
ActionBar
Alert
Badge
Breadcrumb
Button
Dropdown
Icon
Identifier
Image
InlineHelp
InputGroup
Menu
Modal
MultiInput
Pagination
Popover
SearchInput
Toggle

I also created an interface we can extend from to make sure each component has the default props available.

export interface ICommonProps {
  id?: string;
  children?: ReactNode;
}

InnaAtanasova:
Wow! This is awesome, @chrismanciero !!!

droshev:
Adding @bcullman @jbadan @jeffredodd @mobot11 @jeffhandley in case they can share their experience.

bcullman:
Can you verify `git mv` to create the typescript files? if you used `git add`, you will lose history

src/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,110 @@
+import React from 'react';
```
bcullman: For this and all others, it looks like you `git add [filename].tsx` files and then did `git rm [filename].js` file. This will result in a loss of source control history on the `[filename].js`.

Please un delete the `*.js` files, then perform a `git mv`, such as `git mv src/ActionBar/ActionBar.js src/ActionBar/ActionBar.tsx`

This will allow git associate the file history.

chrismanciero: @bcullman - Do you happen to know the git command to un delete the *.js files? I found a command git checkout remotes/origin/develop -- src/ActionBar/ActionBar.js but didn't know if there is an easier way to retrieve the deleted files.

bcullman: i believe that `git checkout HEAD -- path/to/file/file.cc` will work

chrismanciero: I was able to get the file back by doing, git checkout remotes/origin/develop -- src/ActionBar/ActionBar.js, but when i renamed ActionBar.js to ActionBar.tsx it still shows the .js file as deleted. Is this correct behavior because the history will show it a being renamed?

bcullman: `git log --follow src/ActionBar/ActionBar.tsx` will follow the commit history to the file back from before the rename, but only after you commit, and only if you make no other edits to the file

```
git log --follow src/ActionBar.js
git mv src/ActionBar/ActionBar.js src/ActionBar/ActionBar.tsx
git add src/ActionBar/ActionBar.tsx
git commit -m "rename ActionBar.js -> ActionBar.tsx"
git log --follow src/ActionBar.tsx
...
[further edits to the ActionBar.tsx file]
```

chrismanciero: Ok i made the changes to the ActionBar, Alert, Badge and Breadcrumb files. Does it look correct? Want to make sure my steps are correct before i do all the other files.

bcullman: looks like only ActionBar retained history (see ` git log --follow ActionBar.tsx` vs `git log --follow Badge.tsx`), and even then, the history is now exceptionally noisy. 

Not sure if this repo forces squash commits or not, but if not, this will be messy history.

Honestly, it might be better to copy paste the directory elsewhere to retain your typescript changes. Then, destroy the branch, and then create a fresh branch where the first thing that is done is a proper git mv for files.

Separately, consider my other comment on this branch, with respect to linting.

chrismanciero: Thanks, yes I do see the history getting crazy. The best solution would probably be to create a new branch if we decide to go with TypeScript.

I have read your comments and do agree that maybe TypeScript is not the best solution right now. I will be posting a question to the team to get their feelings on if TypeScript should be implemented yet or try to align with SAP Concur with the linting rules for the components

maylukas:
@bcullman:
I went through the active rules from your eslint pull request and I think there are only 5 needed rules missing:

✅ jsx-a11y/alt-text: 2
  => react-a11y-image-button-has-alt
  => react-a11y-img-has-alt
✅ jsx-a11y/aria-props: 2
  => react-a11y-props
✅ jsx-a11y/aria-proptypes: 2
  => react-a11y-proptypes
✅ #jsx-a11y/aria-role: 2
  => react-a11y-role
✅ jsx-a11y/aria-unsupported-elements: 2
  => react-a11y-aria-unsupported-elements
✅ jsx-a11y/heading-has-content: 2
  => react-a11y-accessible-headings
✅ jsx-a11y/html-has-lang: 2
  => react-a11y-lang
✅ jsx-a11y/role-has-required-aria-props: 2
  => react-a11y-role-has-required-aria-props
✅ jsx-a11y/tabindex-no-positive: 2	
  => react-a11y-tabindex-no-positive
⚠ jsx-a11y/accessible-emoji: 2 
  => unsupported / not needed in a component library
⚠ jsx-a11y/iframe-has-title: 2
  => unsupported / not needed in a component library as the components won't include iframes (I hope :) )
⚠ jsx-a11y/img-redundant-alt: 2
  => unsupported / kind of obvious in a code review and most likely not needed in a component library as images will be created by the downstream application
⚠ jsx-a11y/label-has-for: [2, {'allowChildren': true, 'required': {'every': [ 'id' ]}}]
  => unsupported / not needed in a component library as forms will be built in the downstream application
⚠ jsx-a11y/media-has-caption: 2
  => unsupported / not needed in a component library the fiori component's most likely won't include audio / videos
⚠ jsx-a11y/no-access-key: 2
  => unsupported / not needed in a component library as keyboard shortcuts would be implemented in the downstream application
⚠ jsx-a11y/no-autofocus: 2
  =>  unsupported / not needed in a component library as forms will be built in the downstream application
⚠ jsx-a11y/no-distracting-elements: 2
  => unsupported / not needed in a component library (who t.f. uses blink / marquee elements anyways)
⚠ jsx-a11y/no-redundant-roles: 2
  => unsupported / not severe. just duplication of the browser default
❗ jsx-a11y/mouse-events-have-key-events: 2
  => unsupported
❗ jsx-a11y/no-interactive-element-to-noninteractive-role: 2
  => unsupported
❗ jsx-a11y/no-noninteractive-tabindex: 2
  => unsupported
❗ jsx-a11y/scope: 2
  => unsupported
❗ jsx-a11y/aria-activedescendant-has-tabindex: 2
  => unsupported

If the missing roles are crucial I would suggest adding them to the tslint-microsoft-contrib project. The missing roles are not that complex anyways.

src/ActionBar/ActionBar.tsx
chrismanciero: @bcullman - Can you take a look at this branch (https://github.com/SAP/fundamental-react/tree/typescript-support). First step I did was rename all .js file to .tsx, push those changes to the branch. Then updated all the code to be the TypeScript version and pushed to branch.

Does the Git History look good?

bcullman:
@maylukas: In PR #106, the non-active rules are current set that way because in the first pass, this repo fails to those rules. Those rules are:

```
    #jsx-a11y/anchor-has-content: 2
     #jsx-a11y/anchor-is-valid: 2
     #jsx-a11y/aria-role: 2
     #jsx-a11y/click-events-have-key-events: 2
     #jsx-a11y/interactive-supports-focus: 2
     #jsx-a11y/no-noninteractive-element-interactions: 2
     #jsx-a11y/no-noninteractive-element-to-interactive-role: [2, { ul: ['tablist', 'menu']}]
     #jsx-a11y/no-onchange: 2
     #jsx-a11y/no-static-element-interactions: 2
     #jsx-a11y/role-supports-aria-props: 2
```

greg-a-smith:
I added my comments to #108.

chrismanciero:
Since the topic of if/when the transition to TypeScript is still ongoing and this branch was more created to show the changes needed to switch to TypeScript i have decided to close it to prevent any unwanted merges to the develop branch

PR #106, opened 2018-12-18
fix: init linting

bcullman:
based on conversation in #102, wanted to init linting, to show how we can normalize code style and enforce best practices.

eslint is the standard linting tool for javacript all modern editors (atom/sublime/vscode, etc) have plugins that can read a repo's .eslint file and provide dynamic code hints. lint can also be run via `npm run lint`, as well as be part of the CI build process to ensure quality.

Process:
* The first commit will be to init lint infrastructure. 
* The next commits will, one at a time, fix issues that are automatically fixable with the `lint --fix` command
* issues that cannot be automatically fixed by `lint --fix`will be left commented, for a future PR
* in the event that travis fails to run lint, or is misconfigured, we may need additional help


.eslintrc
```diff
@@ -0,0 +1,140 @@
+env:
+    browser: true
+    node: true
+    es6: true
+    jest: true
+
+globals:
+    $: true
+    widgets: true
```
greg-a-smith: I don't think `widgets` is needed here.  I think that's left over from `nui-widgets`.

bcullman:
remaining lint rules that are still commented out will will require code changes that cannot be automatically fixed with `lint --fix`

bcullman:
@greg-a-smith. Yes. all remaining lint rules require hand manual intervention, and should be done in separate PRs

PR #107, opened 2018-12-18
fix: eject from create react app; prep for linting

bcullman:
prep for #106 

linting requires eslint, which cannot be declared until we eject from create-react-app. 

Ejecting in this PR.

Note: ALL code was generated automatically by running `npm run eject`, the standard way to remove the react training wheels. documented here: https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#npm-run-eject


greg-a-smith:
@bcullman So, just so I understand this.  The `eject` option just makes a copy of the boilerplate JS scripts (from create-react-app) and stashes them in the `config` and `scripts` directories, right?  I'm guessing this PR review isn't to specifically review the code in those files, right?

bcullman:
@greg-a-smith: correct.

PR #117, opened 2018-12-20
fix: Add Polyfill for IE11 support

### LINKED ISSUE #110: Add IE11 Support
SAP Concur will continue to support IE11 until Microsoft stops supporting IE11 (target date Oct 14, 2025)

we need to fundamental-react to support IE11 until then as well.

We will need to change this as well as their babel config, they are currently using the create-react-app base config.,

this will require ejecting from create-react-app

CORUI-6052

jeffredodd:
Hello Fundamental React!

This PR is rather simple and focuses on supporting IE11 and will resolve #110.

- Added Babel-Polyfill for Documentation site
- Removed `Object.includes` from Calendar.js


bcullman:
can you add backing lint rule to avoind having `Object.includes` sneak back in?

https://www.npmjs.com/package/eslint-plugin-ie11

jeffredodd:
@bcullman I brought in these linting rules except for `ie11/no-loop-func` as this is a common practice in react with the linting rule aiming to improve performance across all frameworks.

bcullman:
i contributed a bit here, so I'll give a partial 🚢

PR #118, opened 2018-12-20
fix: move lint extend react-app to eslintrc. comment out due to failing rules

bcullman:
once a .eslintrc was made available, it invalidated

```
"eslintConfig": {	
     "extends": "react-app"	
   }
```
in the package.json.  

Moved the extension to eslintrc, where i found out that rules under this extension are failing (and have always been - lint was never a limiting factor to the repo before), so it is commented out for now, like other failing rules that need to be addressed.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,4 +3,5 @@
 *.opts
 *.scss
 *.snap
+*.svg
 node_modules/**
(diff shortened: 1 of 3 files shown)


.eslintignore
```diff
@@ -3,4 +3,5 @@
 *.opts
 *.scss
 *.snap
+*.svg
```
bcullman: dont lint .svg files

Issue #114, opened 2018-12-18
Enable code coverage output to Travis


bcullman:
Travis claims that code coverage is run, but there is no output to look at.

it is *possible* that this simply requires a flag, or that we need to eject from create-react-app

CORUI-6061

PR #119, opened 2018-12-21
fix: init coverage

bcullman:
code coverage was available, but not reported in Travis. This corrects that

fixes: #114

see:
image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,9 @@ jobs:
   include:
     - stage: "Fundamental-react: Lint"
       script: npm run lint
-    - stage: "Fundamental-react: Build"
+    - stage: "Fundamental-react: Test"
       name: "Unit tests"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -8,13 +8,14 @@ branches:
...
   - "/^feature\\/.*$/"
+  - "/^fix\\/.*$/"
 jobs:
   include:
     - stage: "Fundamental-react: Lint"
       script: npm run lint
-    - stage: "Fundamental-react: Build"
+    - stage: "Fundamental-react: Test"
       name: "Unit tests"
-      script: "./ci-scripts/unit-tests.sh"
```
bcullman: this script is no longer used. it could be removed.

@InnaAtanasova - looks like this was meant to gate checkin to make sure code coverage was above a threshold, but I am not sure it was working. Can you confirm?

PR #122, opened 2018-12-21
Issue#121: refactor search input

InnaAtanasova:


InnaAtanasova:
Tests will be added in a separate ticket

chrismanciero:
@InnaAtanasova - A couple of questions on the new implementation of the control.

- When typing in the search box and the autocomplete list show, if the user presses the Enter key the top item in the list is not selected. I used to in the past, should this functionality be put back in?
- Previously there was 2 modes, one with autocomplete and on without. Is a non-autocomplete version available?
- I noticed that pressing Enter in search box calls a method on the parent, but selecting an item calls an item's callback property. Would it be better to just have one method in the parent handle the Enter key and item selection?

InnaAtanasova:
> * I noticed that pressing Enter in search box calls a method on the parent, but selecting an item calls an item's callback property. Would it be better to just have one method in the parent handle the Enter key and item selection?

@chrismanciero , this is also a valid point. My idea of attaching a callback property to each item and a separate method to handle the Enter key is to give more freedom to the developers and make the component more generic. But I get your point. I will raise the question to the Fiori Fundamentals team and we can discuss it after the holidays.

InnaAtanasova:
@chrismanciero , thanks for the review and for the valid points!

PR #131, opened 2019-01-02
Update webpack dev server

chrismanciero:
Because of security vulnerabilities in npm, an update to webpack-dev-server are needed

https://www.npmjs.com/advisories/725

### DIFF
--- a/src/Calendar/Calendar.test.js
+++ b/src/Calendar/Calendar.test.js
@@ -31,9 +31,7 @@ describe('<Calendar />', () => {
       ]} />
   );
   const disabledWeekDay = <Calendar disableWeekday={['Monday', 'Tuesday']} />;
-  const rangeSelect = (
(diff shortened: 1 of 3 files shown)


src/Calendar/Calendar.test.js
```diff
@@ -135,7 +133,7 @@ describe('<Calendar />', () => {
 
     // check that April was selected
     const currentDateDisplayed = wrapper.state('currentDateDisplayed');
-    expect(currentDateDisplayed.getFullYear()).toEqual(2021);
+    expect(currentDateDisplayed.getFullYear()).toEqual(2022);
```
InnaAtanasova: :D I also fixed this in my PR

chrismanciero: I actually have a real fix, so that this problem doesn't happen next year, in my update unit test branch that i will push after your changes are merged 😄

PR #136, opened 2019-01-03
FOR DISCUSSION - Enhancement/improvements for search input component

InnaAtanasova:
There is a question raised by @aerfio about the Search Input: 

> Would you consider giving user the possibility to use onChange props, and option to hide button on right-hand side? It would be quite useful :D

Let me know what you think about having this functionality added to our component.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,5 @@
+/node-modules
+/scripts
+/src
+/config
+/ci-scripts
--- a/package.json
+++ b/package.json
@@ -1,154 +1,154 @@
 {
-  "name": "fundamental-react",
-  "version": "0.0.6-beta",
-  "private": false,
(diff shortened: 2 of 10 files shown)


chrismanciero:
@CodesOfRa - If you wanted to change a sister component on search, should that be handled by the parent (lifting the state up - [https://reactjs.org/docs/lifting-state-up.html](https://reactjs.org/docs/lifting-state-up.html))? You have the parent component with the search component and the sister component, the Search component passes the search term back up to the parent and then the parent passes the search value to the sister component. Using the example you gave, then the sister component will handle the filtering of the list.

CodesOfRa:
@chrismanciero yes that it is correct .

aerfio:
Yes, our use case would be like in the example provided by @CodesOfRa.

Issue #115, opened 2018-12-18
Remove react router as a dependency


bcullman:
This should allow consumers to use React-Router 4 which will allow for greater code splitting of applications.

CORUI-5993

droshev:
@bcullman the link to your JIRA requires user

InnaAtanasova:
@bcullman , we are using react-router for the documentation components. If we separate the documentation, do you think we still need to remove react router as a dependency?

jiawang1:
some components depends on react router: Breadcrum, Menu, tabs....

bcullman:
@jiawang1 they will need to be re-written so that they operate using props and not a direct dependency on react-router.

Issue #105, opened 2018-12-17
Consider separating the library code from the examples (Playground)


maylukas:
Currently the library is mixed with the example page. The index.tsx for example both exports the components and bootstraps the demo application.

The storybook project ([https://github.com/storybooks/storybook](https://github.com/storybooks/storybook)) does pretty much the same than what the Playground is designed for in my opinion.
Using Storybook, the documentation is written in so called stories. They are stored aside the components they describe and are then picked up by the storybook scripts. Additionally stoybook provides a UI that showcases the components and provides functionality such as event logging, knobs (changing props), responsive preview, generating the interface documentation from typescript code and much more.
Example:
[React official storybook](https://storybooks-official.netlify.com/?selectedKind=Components%7CForm%2FButton&selectedStory=with%20knobs&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook-addon-viewport%2Faddon-panel)

When the library is published the stories can be easily filtered using a .npmignore file. 

If you consider this a valid option I would be happy to provide an example for this project.

<details>
<summary>Example for a story:</summary>

```typescript

import *  as React from 'react';
import { storiesOf } from '@storybook/react';

import { wInfo } from '../utils';
import { Tabs } from './tabs';
import { Tab } from './tab';
import { action } from '@storybook/addon-actions';
import { text } from '@storybook/addon-knobs';
import { Story } from '../story';

const info = `
### Notes 
This control is used to navigate between frequently accessed, distinct content categories.
They allow for navigation between two or more content views.
When the header is clicked, the section's content is displayed.

### Do
* Use on content-heavy pages that would otherwise require a significant 
  amount of scrolling to access the various sections.
* Be concise on the navigation labels, ideally one or two words rather than a phrase.

### Usage
* Don’t use the Pivot to link to a new page.
* Don’t use the Pivot to link to hidden content.
~~~js
<Tabs />
~~~`;

(storiesOf('Components/Tabs', module) as any).addWithJSX(
    'Tabs',
    wInfo(info)(() => {
        return (
            <Story>
                <div>
                    <Tabs
                        onSelectTab={action('on-select-tab')}
                        selectedTab={
                            text('selectedTab', null) !== null && text('selectedTab', null) !== ''
                                ? text('selectedTab', null)
                                : undefined
                        }
                    >
                        <Tab key={'tab-1'} title={{id: 'dummy', defaultMessage: 'Tab 1'}}>
                            Content for Tab 1
                        </Tab>
[...]

PR #163, opened 2019-01-09
fix: Updating Dependencies

### LINKED ISSUE #111: Cleanup dependencies
Looking through the dependencies, it appears some of the dependencies can be moved to devDependencies and some can be moved to peerDependencies.  Organizing dependencies properly can dramatically reduce the size of the package.

CORUI-5995

jeffredodd:
The goal of this PR is to migrate the documentation/build-time dependencies (Webpack, babel, PostCSS, and their various plugins etc) to `devDependencies` while the `dependencies` should only directly relate to the required libraries for consumers of the widgets. 

Also in this PR I have introduced `react` and `react-dom` as peerDependencies. This will reduce our dependencies while giving our consumers some flexibility. 

This resolves #111 and is a pre-requisite for #105 and #115.


jeffredodd:
@mobot11 also in this files defence, I think the create-react-app ejection is responsible for the majority of these dependencies being in the `dependencies` section.

jeffredodd:
@droshev I think for now a large list of devDeps is ok since our future consumers won't be required to install the devDeps. In a future PR, #115, our deps will get even smaller as we won't need react router for the documentation except for local development and the demo site publishing. 😄

jeffredodd:
❓ I think we are pretty close to merging this, anyone have thoughts around merging? ❓

droshev:
@jeffredodd i am fine

PR #168, opened 2019-01-09
fix: update webpack-dev-server to non-vulnerable version

mobot11:
There is a security alert for `webpack-dev-server`. Updating to version 3.1.11 or higher fixes the issue.
For more information see https://github.com/SAP/fundamental-react/network/alert/package.json/webpack-dev-server/open.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2873,6 +2873,7 @@
         "anymatch": "2.0.0",
         "async-each": "1.0.1",
         "braces": "2.3.2",
+        "fsevents": "1.2.4",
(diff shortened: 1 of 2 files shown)


jbadan:
I believe @droshev tried this, but there's a problem with something related to react-scripts.

mobot11:
@droshev I was able to build and start the demo with the updated version of react-dev-server. Is there another reason this version cannot be updated?

droshev:
@mobot11 I think react-scripts released yesterday a new version addressing the issue. I was thinking of updating our repo.

Issue #171, opened 2019-01-10
Update components to use arrow function syntax


greg-a-smith:
Since the repository is already using ES6, we can take advantage of the arrow function syntax.  When these are part of a React component, they will autobind removing the need for statements like this in the constructor:
```
this.updateDate = this.updateDate.bind(this);
```
Instead of writing component methods like:
```
updateDate(date) {
    ...
}
```
The functions can be written as such:
```
updateDate = (date) => {
    ...
}
```

**NOTE:** Standard React lifecycle functions do not need to be converted as React already handles the binding.

**NOTE:** There are likely some existing codemods that should make this easier.

jbadan:
fixed by https://github.com/SAP/fundamental-react/pull/172

PR #172, opened 2019-01-10
fix: update components to use arrow function syntax

chrismanciero:
PR for issue #171


chrismanciero:
thanks @greg-a-smith, i missed that file. For the indent spacing i setup my VS Code to Prettier extension to look at the eslint file for formatting on save. Here is the rule in the file **#indent: [2, 4, {"SwitchCase": 1}]** which i think is commented out.

greg-a-smith:
@chrismanciero Yes, that's the lint rule.  It's still commented out because there are a number of places that fail when enabled.  Our plan was to circle back and make the fixes necessary to satisfy the rules as we enable them.  It will likely be a number of PRs until they are all enabled.  I'm fine leaving the changes in this PR and then everything will be addressed when the work to enable that rule is done.

mobot11:
To piggyback off of what @greg-a-smith  said, I opened https://github.com/SAP/fundamental-react/issues/174 to create a .editorconfig so we can don't have to think about our editor settings when making contributions to the repo.

bcullman:
consider adding a backing eslint rules for this:

https://eslint.org/docs/rules/arrow-body-style

chrismanciero:
Sure no problem, the merge is still blocked until @greg-a-smith resolves his request, which i am assuming he is waiting until the automated version is resolved 😄

greg-a-smith:
@chrismanciero Do you want to take a shot at adding a https://eslint.org/docs/rules/arrow-body-style lint rule while this PR is on hold?

chrismanciero:
@greg-a-smith sure i can take a look today, is there a specific option we want set by default for this lint rule?

greg-a-smith:
@chrismanciero Actually, looking at that rule, I think I misunderstood it.  I initially took it as a rule that would enforce the use of arrow functions for component methods, but this rule is simply for the syntax of the function body.  Although we could maybe look into rules like this in the future, I don't think this rule relates to this PR.  We have not used any of these rules in our SAP Concur repo so we don't have any experience with them yet.

PR #184, opened 2019-01-12
chore: Prop spreading unit tests

### LINKED ISSUE #167: create backing unit tests for prop spreading
we will unit test everything, but will `xit` for ones that do not pass, and add future stories to address them.

CORUI-6086

greg-a-smith:
I have added a number of unit tests to verify that props can be spread through each component and end up in the emitted HTML.  The tests use `data-sample` as the HTML attribute (prop).  There are a number of components and elements that do not allow these props to be spread and I tried to put a skipped test with a description of the test needed so it helps the developer(s) adding the additional prop spreading functionality.

**NOTE:** Most of the `.test` files were also changed from 2 space indents to 4 so my apologies for the massive diffs in these files.  I didn't change _any_ existing tests.  The only thing I did was add a `describe` block called "Prop spreading" that contains the prop spreading tests.

Fixes #167.


jbadan:
Looks good 🚢  (once we get the publishing working ;) )

bcullman:
@greg-a-smith So, to be clear the added `xtest` in this PR (44 by my count) demonstrate locations in the code that prop spreading is not yet complete.

greg-a-smith:
@bcullman That's correct, for the most part.  There are a couple places where it might get fairly complicated so I just left a generic comment that more tests will need to be added as functionality is added.

PR #197, opened 2019-01-15
fix: eslint no empty

greg-a-smith:
Enabled the `no-empty` eslint rule and fixed lint errors.

### DIFF
--- a/src/Time/Time.js
+++ b/src/Time/Time.js
@@ -217,7 +217,6 @@ class TimeItem extends Component {
       aux = event.target.value.replace(/\D/, '');
       this.updateStyle(style, aux, max);
       this.setState({ value: aux });
-    } else {
     }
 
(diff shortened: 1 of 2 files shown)


src/Time/Time.js
```diff
@@ -316,6 +315,7 @@ export class Time extends Component {
     super(props);
     const { time } = this.props;
     if (time.hour !== '00') {
+        // do nothing
```
bcullman: This can just be flipped to `if (time.hour == '00')`

greg-a-smith: You know ... I hmmm'd and hah'd over this.  I ended up doing this because of my fear of the undefined, null, etc. values.  But, I may have just been worrying over nothing.  I guess undefined and null would "not equal" `00` so maybe this will be just a quick change.

PR #202, opened 2019-01-15
fix: eslint react in jsx scope

greg-a-smith:
Enabled the `react-in-jsx-scope` eslint rule.


greg-a-smith:
@bcullman Correct, other changes must have already addressed this.

PR #203, opened 2019-01-15
fix: eslint valid jsdoc

greg-a-smith:
Enabled the `valid-jsdoc` eslint rule and fixed lint errors.


bcullman:
not sure about this one… we should discuss the merits of JS docs, if there are any.

bcullman:
looks like most of the JSDocs lines came from @CodesOfRa, it woul dbe best if she chimes in before removing this.

it *does* appear that we are manually documenting props in the *.Components.js files. If we *were* to use these JSDocs capabilities, instead, we should do it universally, and in a way that can be consumed automatically by ghpages.

PR #205, opened 2019-01-15
fix: eslint no unused vars

greg-a-smith:
Enabled the `no-unused-vars` eslint rule and fixed lint errors.

### DIFF
--- a/src/MultiInput/MultiInput.test.js
+++ b/src/MultiInput/MultiInput.test.js
@@ -38,7 +38,7 @@ describe('<MultiInput />', () => {
 
     let wrapper;
 
-    const getListStatus = (wrapper, bIsShown) => {
+    const getListStatus = (bIsShown) => {
         const combobox = wrapper.find(
             `div.fd-combobox-control[aria-expanded=${bIsShown}]`
         );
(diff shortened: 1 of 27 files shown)


src/MultiInput/MultiInput.test.js
```diff
@@ -55,23 +55,23 @@ describe('<MultiInput />', () => {
     });
 
     // create a default multi-input control
-    test('create multi-input', () => {
+    xtest('create multi-input', () => {
```
bcullman: @greg-a-smith: why are these test being `x`-ed?

src/Shellbar/Shellbar.test.js
```diff
@@ -51,101 +50,6 @@ describe('<Shellbar />', () => {
...
-            copilot
-            searchInput={searchInput}
-            actions={actions}
-            notifications={notifications}
-            profile={profile}
-            profileMenu={profileMenu}
-            productSwitcher={productSwitcher}
-            productSwitcherList={productSwitcherList} />
-    );
-
```
jbadan: Were these made with the intention of using them in the snapshot tests?

greg-a-smith: That's a good question.  There were a bunch of variables defined to create it, but then it was never used.  I will check the blame on this to see if someone wants to retain all this code in a separate branch, but it should all be removed from `master`.

chrismanciero: My bad, i think i forgot to make the snapshot call.  I will create a chore: PR to add the call in there. This snapshot call bumps up the code coverage from 30% to about 85%

greg-a-smith: @chrismanciero Do you want to fix it in this PR or create a separate one?

PR #210, opened 2019-01-15
fix: update slack location for travis notifications

bcullman:
repoint travis to our new public slack workspace: ui-fundamentals


greg-a-smith:
**NOTE:** Builds are currently _NOT_ being picked up by Travis.

PR #217, opened 2019-01-16
fix: eslint valid typeof

greg-a-smith:
Enabled the `valid-typeof` eslint rule and fixed lint errors.

### DIFF
--- a/src/DatePicker/DatePicker.js
+++ b/src/DatePicker/DatePicker.js
@@ -62,12 +62,7 @@ export class DatePicker extends Component {
         secondDateYear;
       } else {
       //Checks if the type of date doesn't match those types and that it doesn't contain any special character symbols
-          if (
-              (typeof date !== 'array' ||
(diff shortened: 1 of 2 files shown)


src/DatePicker/DatePicker.js
```diff
@@ -62,12 +62,7 @@ export class DatePicker extends Component {
         secondDateYear;
       } else {
       //Checks if the type of date doesn't match those types and that it doesn't contain any special character symbols
-          if (
-              (typeof date !== 'array' ||
-          typeof date !== 'date' ||
-          typeof date !== 'object') &&
-        date.toString().search(regex) !== 1
-          ) {
```
greg-a-smith: I could not figure out how the block:
```
(typeof date !== 'array' ||
 typeof date !== 'date' ||
 typeof date !== 'object')
```
could ever return anything but `true` so I removed it and just based the `if` condition on the regex.

PR #228, opened 2019-01-16
fix: Manual version bump to sync the repo [ci skip]

greg-a-smith:
Version got out of sync so this should reset it.


greg-a-smith:
The ci skip prevents the build so will redo this.

Issue #234, opened 2019-01-17
Update `dismissable` prop in Alert component to `dismissible`


mobot11:
Currently, the prop is named `dismissable`, however the correct spelling is `dismissible` which is how the class is spelt as well `fd-alert--dismissible`.

PR #235, opened 2019-01-17
fix: rename dismissable to dismissible in Alert component

mobot11:
Fixes https://github.com/SAP/fundamental-react/issues/234


greg-a-smith:
Minor nit: The PR should be prefixed as `fix:` rather than `feat:`.

PR #237, opened 2019-01-17
fix: update doc, add pagination spread props

chrismanciero:
include displayTotalProps and linkProps in the documentation for Pagination component

### DIFF
--- a/src/Pagination/Pagination.Component.js
+++ b/src/Pagination/Pagination.Component.js
@@ -46,6 +46,21 @@ export const PaginationComponent = () => {
                             'number (default: page 1) - Initial page to be selected.'
                     },
                     {
+                        name: 'displayTotalProps',
+                        description:
+                            'object - additional props to be spread to the display total section'
+                    },
+                    {
+                        name: 'linkProps',
+                        description:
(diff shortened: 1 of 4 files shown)


src/Pagination/Pagination.Component.js
```diff
@@ -45,6 +45,21 @@ export const PaginationComponent = () => {
                         description:
                             'number (default: page 1) - Initial page to be selected.'
                     },
+                    {
+                        name: 'displayTotalProps',
+                        description:
+                            'object - additional props to be spread to the display total section'
+                    },
+                    {
+                        name: 'linkProps',
```
greg-a-smith: I missed the original PR (#224), but should we have two separate props for the "previous" and "next" links?

chrismanciero: Separate props as in a prevNextProps which is different than the linkProps or as in prevProps and nextProps

greg-a-smith: I was asking about `prevProps` and `nextProps`.

chrismanciero: ok, i just added a prevProps and nextProps values to the pagination control

PR #246, opened 2019-01-17
fix: Prop spreading for Forms

greg-a-smith:


src/Forms/Forms.test.js
```diff
@@ -312,7 +313,7 @@ describe('<Forms />', () => {
 
         xtest('should allow props to be spread to the FormRadio component', () => {
```
chrismanciero: should this be **test** and not **xtest**?

greg-a-smith: Sorry, I was busy creating an issue to link in my annotation.  I will add that now.

greg-a-smith:
**NOTE:** I left the `FormRadio` tests `x`'d out because I actually think the component itself should be refactored to better allow for prop spreading.  See #247 .

PR #248, opened 2019-01-17
fix: Prop spreading for SideNavigation

cjskillingstad:
- Spread props to SideNavigation's elements
- Update unit tests

The documentation for this component didn't include the props for the component that I added prop spreading to so I didn't update docs.

### DIFF
--- a/src/SideNavigation/SideNavigation.js
+++ b/src/SideNavigation/SideNavigation.js
@@ -159,12 +159,12 @@ SideNavList.propTypes = {
 };
 
 export const SideNavGroup = props => {
-    const { title, children, className, headerProps, ...rest } = props;
+    const { title, children, className, titleProps, ...rest } = props;
     return (
         <div
             className={`fd-side-nav__group${className ? ' ' + className : ''}`}
             {...rest}>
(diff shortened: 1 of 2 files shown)


src/SideNavigation/SideNavigation.js
```diff
@@ -159,12 +159,12 @@ SideNavList.propTypes = {
...
 
 export const SideNavGroup = props => {
-    const { title, children, className, ...rest } = props;
+    const { title, children, className, headerProps, ...rest } = props;
     return (
         <div
             className={`fd-side-nav__group${className ? ' ' + className : ''}`}
             {...rest}>
-            <h1 className='fd-side-nav__title'>{title}</h1>
+            <h1 {...headerProps} className='fd-side-nav__title'>{title}</h1>
```
greg-a-smith: @jbadan had a similar change to a different component and used `titleProps` as the name, which seems to fit a little better since the content of this element is the `title` prop.

greg-a-smith: On an unrelated note, this element should assume the use of `<h1>`.  There should only be one `<h1>` per page and assuming it will exist in the side navigation seems wrong.  I created #249 to address that in a future story.

PR #253, opened 2019-01-18
fix: prop spreading to Calendar

jbadan:
spreading props to Calendar component


src/Calendar/Calendar.js
```diff
@@ -459,39 +459,57 @@ export class Calendar extends Component {
...
+              monthListProps,
+              yearListProps,
+              tableProps,
+              tableHeaderProps,
+              tableBodyProps,
+              ...props
+          } = this.props;
 
           return (
               <div className={`fd-calendar${className ? ' ' + className : ''}`} {...props}>
```
chrismanciero: probably want to move (...props} in front of className

chrismanciero:
🚢

Issue #262, opened 2019-01-18
Enabling caching on Travis builds


jbadan:
Enabling caching to be good code citizens and share Travis resources more effectively. 

https://docs.travis-ci.com/user/caching/#fetching-and-storing-caches

PR #263, opened 2019-01-18
chore: enable caching of node_modules

jbadan:
enable caching of node_modules in travis (per branch)

fixes https://github.com/SAP/fundamental-react/issues/262

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 node_js: 10.15.0
+cache: npm
 dist: trusty
 git:
   depth: false


bcullman:
dumb ?:

if a module is updated, I take it that it knows the cache in invalid?

jbadan:
@bcullman not a dumb question! According to the docs (sometimes unreliable to be honest), this should only cache node_modules on the same branch. So, it should just speed up our secondary “pr” build. 

If we include this, I want to keep an eye on it for any weird behaviors.

greg-a-smith:
That said, should we wait on this merge until the other open PRs are merged and a release is done?  Seems like it would be a nice first story for the next rc version chain.  This way, we can finish this `0.1.0` release cycle with a known and proven process?

PR #264, opened 2019-01-18
fix: Prop spreading for Tree

greg-a-smith:
**NOTE:** The API doesn't easily allow for prop spreading to elements below the header and list so I did a few of them in this PR and created #265 for the follow-up work.

### DIFF
--- a/src/Tree/Tree.js
+++ b/src/Tree/Tree.js
@@ -146,7 +146,7 @@ export class Tree extends Component {
                               role='treeitem'>
                               <div className='fd-tree__row'>
                                   {parent}
-                                  {<Dropdown isContextual />}
+                                  {<Dropdown />}
                               </div>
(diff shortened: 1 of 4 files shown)


src/Tree/Tree.js
```diff
@@ -146,7 +146,7 @@ export class Tree extends Component {
                               role='treeitem'>
                               <div className='fd-tree__row'>
                                   {parent}
-                                  {<Dropdown isContextual />}
+                                  {<Dropdown />}
```
bcullman: did we ever get an answer on this `isContexual` prop?

greg-a-smith: Yes, @chrismanciero found a commit where it was removed from `Dropdown`, but apparently just missed in a few of the places that called it.

PR #273, opened 2019-01-23
fix: update publishing script for releases

jbadan:
Change logic for publishing releases.

`npm run release` will checkout a new branch on your local machine called `automated_travis_release_do_not_use` and use `git push` to trigger a travis build. Travis will then run a deployment based off branch names. Merges to master will still result in a rc release, while builds from `automated_travis_release_do_not_use` will result in a release.
`automated_travis_release_do_not_use` will then be deleted both locally and remotely to ensure this works for multiple releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ jobs:
   - stage: "Fundamental-react: Test & Lint"
     script: npm run lint && npm run test:coveralls && npm run size
   - stage: Deploy
-    if: (branch = master OR branch = automated_travis_release_do_not_use) AND type = push
+    if: (branch = master OR branch = tmp_branch_for_automated_release_do_not_use) AND type = push
     before_deploy:
     - bash ./ci-scripts/setup_npm.sh
     deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,13 +10,15 @@ jobs:
   - stage: "Fundamental-react: Test & Lint"
     script: npm run lint && npm run test:coveralls && npm run size
   - stage: Deploy
-    if: branch = master AND type != pull_request
+    if: (branch = master OR branch = automated_travis_release_do_not_use) AND type = push
     before_deploy:
     - bash ./ci-scripts/setup_npm.sh
     deploy:
       - provider: script
         script: bash ./ci-scripts/publish.sh $TRAVIS_BRANCH $TRAVIS_BUILD_NUMBER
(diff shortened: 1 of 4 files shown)


ci-scripts/publish_release.sh
```diff
@@ -1,3 +1,4 @@
+
```
bcullman: unnecessary whitespace here.

package.json
```diff
@@ -26,7 +26,6 @@
     "size:debug": "npm run build && size-limit --why",
     "start-js": "node scripts/start.js",
     "start": "node scripts/start.js",
-    "std-version:release": "standard-version -m \"chore(release): version %s\"",
     "std-version": "standard-version -m \"chore(release): version %s build ${TRAVIS_BUILD_NUMBER} [ci skip]\"",
```
bcullman: so run `/ci-scripts/publish_release.sh' to release?

.travis.yml
```diff
@@ -10,13 +10,15 @@ jobs:
   - stage: "Fundamental-react: Test & Lint"
     script: npm run lint && npm run test:coveralls && npm run size
   - stage: Deploy
-    if: branch = master AND type != pull_request
+    if: (branch = master OR branch = automated_travis_release_do_not_use) AND type = push
```
bcullman: perhaps rename to `tmp_branch_for_automated_release_do_not_use`

package.json
jbadan: Yeah, `npm run release` calls `/ci-scripts/publish_release.sh`, which calls `/ci-scripts/publish.sh` inside of travis so that we have the correct variables. We can reuse the same `npm run std-version` now that we don't need a `[ci-skip]` and non-`[ci-skip]`

jbadan:
@greg-a-smith The problem with publishing on tags, is that we need to create the tags first. Standard-version does this, but we cannot push from local because of all the branch protections. It makes a logic nightmare trying to separate out the tag from the other builds then proceeding that way. This is not a great explanation, but we can zoom if you'd like! Easier to show than than explain.

greg-a-smith:
@jbadan Although I think/hope there's a way to do this with tags, I think we should postpone that research until we circle back on deployments when we try and get the "Latest release" badge to show up correctly.  There's gotta be a way to sequence our flow from the GitHub UI or from the GitHub Rest API -- both of which will set that "Latest release" badge accordingly.

PR #274, opened 2019-01-23
fix: import path on docs

### LINKED ISSUE #239: Fix Import component on playground
The `Import` component for the playground is showing imports as:
```
import { ActionBar } from '/fundamental-react/src/';
```
This is not correct or how a consumer would truly import the component.

Consider showing tree-shakable imports that import directly from the path in addition to named imports (like shown above).

greg-a-smith:
**NOTE:** The essence of the change is in `Import.js`, which unfortunately shows up at the bottom of the "Files changed" tab's diff.  Please start there and the rest of the changes will make more sense.

Fixes #239

### DIFF
--- a/src/_playground/documentation/Import/Import.js
+++ b/src/_playground/documentation/Import/Import.js
@@ -1,16 +1,17 @@
+import arraySort from 'array-sort';
+import packageJson from '../../../../package.json';
 import PropTypes from 'prop-types';
 import React from 'react';
 
-export const Import = (props) => {
-
-    const { module, path } = props;
+export const Import = ({ sourceModule }) => {
 
     return (
-        <code>import &#123; { module } &#125; from '{ path }';</code>
+        <code>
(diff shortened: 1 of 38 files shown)


src/_playground/documentation/Import/Import.js
```diff
@@ -1,16 +1,17 @@
...
 
-export const Import = (props) => {
-
-    const { module, path } = props;
+export const Import = ({ sourceModule }) => {
 
     return (
-        <code>import &#123; { module } &#125; from '{ path }';</code>
+        <code>
+            import &#123; { arraySort(Object.keys(sourceModule)).join(', ') } &#125; from '{packageJson.name}';
```
greg-a-smith: In addition to fixing the path shown in the example `import` on each component's playground page, let's also remove hard-coded value props and just scan the code to find exported components and the name of the package.

PR #278, opened 2019-01-23
chore: add gh deploy script to publish

jbadan:
Adds gh-pages release to publish script. Tested that this does run predeploy and deploy at the same time.


ci-scripts/publish.sh
```diff
@@ -19,6 +19,8 @@ if [[ "$TRAVIS_BRANCH" = "tmp_branch_for_automated_release_do_not_use" ]]; then
   
     npm publish
 
+    npm run deploy
```
bcullman: this is covered in the if statement above right? only push on actual releases, not RCs?

jbadan: Yeah, only publish on `npm run release`

PR #281, opened 2019-01-24
fix: fixes #279 Add `event` argument to onChange props in SearchInput

### LINKED ISSUE #279: Add missing functionality to SearchInput
Right now we have no access to `event` in `onChange` prop.
https://github.com/SAP/fundamental-react/blob/370325509cfea75d5277b96fe99dbe982e444158/src/SearchInput/SearchInput.js#L37
If you just changed this line to `this.props.onChange(event)` it would work 👍

aerfio:
Resolves #279

### DIFF
--- a/src/SearchInput/SearchInput.js
+++ b/src/SearchInput/SearchInput.js
@@ -38,9 +38,7 @@ export class SearchInput extends Component {
         } else {
             if (this.state.searchList) {
                 let filteredResult = this.state.searchList.filter(item =>
-                    item.text
-                        .toLowerCase()
-                        .startsWith(event.target.value.toLowerCase()),
+                    item.text.toLowerCase().startsWith(event.target.value.toLowerCase())
                 );
                 this.setState({
                     filteredResult: filteredResult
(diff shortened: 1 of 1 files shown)


src/SearchInput/SearchInput.js
```diff
@@ -34,11 +34,13 @@ export class SearchInput extends Component {
             value: event.target.value
         });
         if (this.props.onChange) {
-            this.props.onChange();
+            this.props.onChange(event);
         } else {
             if (this.state.searchList) {
                 let filteredResult = this.state.searchList.filter(item =>
-                    item.text.toLowerCase().startsWith(event.target.value.toLowerCase())
+                    item.text
```
chrismanciero: @aerfio - adding the event argument to the this.props.onChange method looks good. But there seems to be some autoformatting applied to the file. Can you either remove the auto-formatting or if you do have a formatter installed in the IDE have it apply the eslint rules that are available in the repo (.eslintrc)?

aerfio: Doesn't it do it automatically on before-commit hook? I've even run `npm run build:lint:fix` and `npm run lint`

aerfio: OK, my prettier config somehow took precedence before eslint, my bad

Issue #245, opened 2019-01-17
fix remaining a11y eslint issues


greg-a-smith:
see commented out rules in https://github.com/SAP/fundamental-react/blob/master/.eslintrc

they are:

    #jsx-a11y/anchor-has-content: 2
    #jsx-a11y/anchor-is-valid: 2
    #jsx-a11y/aria-role: 2
    #jsx-a11y/click-events-have-key-events: 2
    #jsx-a11y/interactive-supports-focus: 2
    #jsx-a11y/no-noninteractive-element-interactions: 2
    #jsx-a11y/no-noninteractive-element-to-interactive-role: [2, { ul: ['tablist', 'menu']}]
    #jsx-a11y/no-onchange: 2
    #jsx-a11y/no-static-element-interactions: 2
    #jsx-a11y/role-supports-aria-props: 2

jbadan:
Completed

PR #286, opened 2019-01-24
fix: Playground component to use onBlur and onChange

jbadan:
Add onBlur to playground component's <select> elements. Another option is to remove the onChange events from these <select> elements entirely. 

#245

### DIFF
--- a/src/_playground/documentation/Playground/Playground.js
+++ b/src/_playground/documentation/Playground/Playground.js
@@ -191,7 +191,7 @@ export class Playground extends Component {
                               <input
                                   className='form-control'
                                   name={item.attribute}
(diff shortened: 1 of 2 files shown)


src/_playground/documentation/Playground/Playground.js
```diff
@@ -179,6 +179,7 @@ export class Playground extends Component {
                               <select
                                   className='form-control'
                                   name={item.attribute}
+                                  onBlur={this.updateComponent}
```
bcullman: dumb q: does this double fire because it is on both onBlur and onChange?

jbadan: I thought it would too, but I played around with it and everything seems fine.

Issue #269, opened 2019-01-21
Refactor certain components that set values that set css classes with dynamic values to have prop types


mobot11:
For example the Identifier.js component sets the class `fd-identifier--${size}` however, the size variables prop type validation is `size: PropTypes.string`. Ideally this would be an explicit range of values so that consumers are warned if they provide a value to to size that isn't valid css. There are a few options to do this, but at SAP Concur we have abstracted out these values into arrays in their own file like so ```...
export const MESSAGEBAR_TYPES = [
    'success',
    'info',
    'warning',
    'error'
];

export const MESSAGEBOX_TYPES = [
    'confirm',
    'error',
    'info',
    'success',
    'warning'
];

export const BUTTON_TYPES = [
    'default',
    'primary',
    'secondary',
    'create',
    'success',
    'info',
    'warning',
    'danger',
    'link'
];
...```

and then we import them into components where they are needed.

chrismanciero:
👍 Sounds like a good idea.

greg-a-smith:
Should we pull together a list of the components and the targeted props in order to scope this?  Or, do we combine this work with another issue (_e.g._ #244) and then create stories to do one component at a time?

mobot11:
@greg-a-smith I don't think there are a ton of places where this happens. I can go ahead and make a list in this issue.

mobot11:
Done: 
- [ ] Alert
- [ ] Badge
- [ ] Button
- [ ] Forms
- [ ] Icon
- [ ] Identifier
- [ ] Image
- [ ] InlineHelp
- [ ] InputGroup
- [ ] Panel
- [ ] Popover
- [ ] Tile
- [ ] Toggle

droshev:
@mobot11 this is a good idea but fundamental-ui is still playing with the exact set of names. There is no an easy way to track any time they change those types.

PR #287, opened 2019-01-24
feat: move explicit prop types into constants file

mobot11:
Fixes https://github.com/SAP/fundamental-react/issues/269


src/Tile/Tile.js
```diff
@@ -39,9 +40,9 @@ Tile.propTypes = {
     backgroundColor: PropTypes.number,
     className: PropTypes.string,
     colorAccent: PropTypes.number,
-    columnSpan: PropTypes.number,
+    columnSpan: PropTypes.oneOf(GRID_COLUMNS),
     disabled: PropTypes.bool,
-    rowSpan: PropTypes.number
+    rowSpan: PropTypes.oneOf(GRID_ROWS)
```
bcullman: should this be `TILE_GRID_COLUMNS` and should this be `TILE_ROW_SPAN`?

src/utils/constants.js
```diff
@@ -0,0 +1,138 @@
...
+];
+
+export const IMAGE_SIZES = [
+    's',
+    'm',
+    'l'
+];
+
+export const IMAGE_TYPES = [
+    'circle'
```
bcullman: missing ` `

that said, I don't like that ` ` is a valid value. lets give it a name (default? rectangular? none?) and operate on that.

src/utils/constants.js
```diff
@@ -0,0 +1,138 @@
...
+    4,
+    5,
+    6,
+    7,
+    8,
+    9
+];
+
+export const IDENTIFIER_MODIFIERS = [
+    '',
```
bcullman: I don't like that ` ` is a valid value. lets give it a name (default? rectangular? none?) and operate on that.

bcullman:
this is a `chore` or `fix`; not a `feat`

src/Tile/Tile.js
greg-a-smith: I don't believe `rowSpan` should have a `oneOf` list of values.  Columns, yes, but not rows.  And I also agree with @bcullman about `TILE_GRID_COLUMNS`.

src/utils/constants.js
```diff
@@ -0,0 +1,138 @@
...
+    1,
+    2,
+    3,
+    4,
+    5,
+    6,
+    7,
+    8,
+    9
+];
```
greg-a-smith: This one seems a little silly to me.  I know `PropTypes` doesn't have a numeric range validator, but what if we create a custom one?  Something that throws an error if the number is less than 1 or greater than 9.  Seems a little easier to maintain especially if this list of colors choices ever gets big.

src/utils/constants.js
greg-a-smith: I'd actually say let's just leave it out.  The value `''` is going to evaluate as falsy anyway.  Besides, if we added a different constant and made it the default value, it would generate a CSS classname that doesn't exist.

src/utils/constants.js
greg-a-smith: I like that `''` isn't here.  The value `''` is going to evaluate as falsy anyway.  Besides, if we added a different constant and made it the default value, it would generate a CSS classname that doesn't exist.

src/utils/constants.js
```diff
@@ -0,0 +1,138 @@
...
+
+export const GRID_ROWS = [
+    1,
+    2,
+    3,
+    4,
+    5,
+    6
+];
```
greg-a-smith: Similar to `IDENTIFIER_COLORS`, `GRID_COLUMNS` and `GRID_ROWS` (assuming `GRID_ROWS` survives based on comment above) could use a custom prop type validator for a numeric range?  Since we've got more than one instance of this now, we could make a `util` function for it that takes `minVal` and `maxVal` as parameters.

src/utils/constants.js
```diff
@@ -0,0 +1,138 @@
...
+
+export const POPOVER_ALIGNMENTS = [
+    'right'
+];
+
+export const TOGGLE_SIZES = [
+    'xs',
+    's',
+    'l'
+];
```
greg-a-smith: Am I the only one that finds it interesting there's no `m` in this list?

mobot11: Yes. The docs say the sizes are `--xs` `--s` normal and `--l`. https://sap.github.io/fundamental/components/toggle.html. We could just make `--m` map to the default class. What do you think?

package-lock.json
```diff
@@ -10,16 +10,16 @@
       "integrity": "sha512-bfna97nmJV6nDJhXNPeEfxyMjWnt6+IjUAaDPiYRTBlm8L41n8nvw6UAqUCbvpFfU246gHPxW7sfWwqtF4FcYA==",
```
greg-a-smith: Looks like you're using `npm` v5.x.x.  For this project, we should be using v6.x.x.  If you install and use `nvm` (Node Version Manager), it makes it pretty easy flipping back and forth.  A `package-lock.json` file shouldn't change without changes to `package.json`.

src/utils/CustomPropTypes/CustomPropTypes.js
```diff
@@ -0,0 +1,35 @@
...
+
+        if (value >= start && value <= end) {
+            return null;
+        } else {
+            return new Error(`${componentName}: ${propName} must be within the range ${start} to ${end}`);
+        }
+    };
+};
+
+export default {range};
```
greg-a-smith: I'll just include some comments for this file without targeting each individual line.

1. Would `start` and `end` be better as `minVal` and `maxVal`?
1. Any validation that occurs on the `start` and `end` parameter values must be in the outer function -- basically meaning that we throw the error (or console log it) and then don't return the validation function.
1. Personally, I would probably just throw `Error` for everything rather than using `TypeError`.  I guess it technically doesn't matter, but not quite sure what you get with `TypeError` that `Error` doesn't give you.
1. The `isNaN` checks are correct, but I'd reword the error messages.  Error messages should be easy-to-read English sentences (see examples below).
1. The convention within `PropTypes` for error messages is to put relevant data (prop names, component names, data types, values, etc.) in backticks so they stand out in the message.

Below are a couple example prop validation errors coming from `PropTypes`:
```
The prop `className` is marked as required in `ActionBar`, but its value is `undefined`.

Invalid prop `glyph` of type `number` supplied to `Identifier`, expected `string`.
```
Maybe these should be something like this:
```
Invalid prop `color` of type `string` supplied to `Identifier`, expected `number`.

The prop `color` supplied to `Identifier` has a value of `10`, which is not in the range `1` to `9` (inclusive).
```

package-lock.json
mobot11: I'm on node 10 and npm 6.4.1. I'll just take the package-lock from master.

src/utils/CustomPropTypes/CustomPropTypes.js
mobot11: I was following https://github.com/airbnb/prop-types/blob/master/src/between.js as an example. I think TypeError is more explicit about why the validation is failing. But I do agree that we can make the messages easier to read and more descriptive.

greg-a-smith: The Airbnb one is a custom extension library whereas `prop-types` (from Facebook) is the defacto standard.  I'd rather pattern after `prop-types` so all prop type validation messages look similar.

greg-a-smith:
@bcullman There are such tests for the new `CustomPropTypes`, but I don't really think it's necessary for the others since all we'd really be testing is `prop-types` since it owns the code that's doing the prop validation.

bcullman:
Let's ship 🚢

Issue #266, opened 2019-01-18
Consider refactoring Shellbar into multiple components


greg-a-smith:
Currently, `Shellbar` is a very complex component and it's very difficult to unit test.  I think this should be refactored into multiple sub-components (`Profile`, `ProductSwitcher`, `Actions`, `Search` etc.) so each can have a concise API and thus be much more testable.  Then Shellbar can take these sub-components and construct the final markup.

AdamFedor:
CORUI-6126 for our side of the work.

arminnaimi:
@greg-a-smith I think this seems to be modelled after the SAPUI5 control that I got to work with just yesterday. Is this project following similar composition choices to SAPUI5?

PR #288, opened 2019-01-24
fix: add roles to static element interactions

jbadan:
Adds role='button' to static element interactions.

Adds unit tests for Icon, Menu, Popover new roles. Skipped unit test on Shellbar per plan to refactor (https://github.com/SAP/fundamental-react/issues/266).

Updated snapshots.

#245

### DIFF
--- a/src/ComboboxInput/__snapshots__/ComboboxInput.test.js.snap
+++ b/src/ComboboxInput/__snapshots__/ComboboxInput.test.js.snap
@@ -11,6 +11,7 @@ exports[`<ComboboxInput /> create combobox input 1`] = `
       aria-expanded={false}
       className="fd-popover__control"
       onClick={[Function]}
+      role="button"
     >
       <div
         className="fd-combobox-control"
@@ -93,6 +94,7 @@ exports[`<ComboboxInput /> create combobox input 2`] = `
       aria-expanded={false}
       className="fd-popover__control"
       onClick={[Function]}
+      role="button"
     >
       <div
         className="fd-combobox-control"
--- a/src/Icon/Icon.js
+++ b/src/Icon/Icon.js
@@ -15,7 +15,9 @@ export const Icon = ({ glyph, size, clickHandler, className, ...props }) => {
         <span
             {...props}
             className={iconClasses}
-            onClick={clickHandler} />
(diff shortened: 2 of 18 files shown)


src/ComboboxInput/__snapshots__/ComboboxInput.test.js.snap
```diff
@@ -11,6 +11,7 @@ exports[`<ComboboxInput /> create combobox input 1`] = `
       aria-expanded={false}
       className="fd-popover__control"
       onClick={[Function]}
+      role="button"
```
bcullman: should this just be a button instead?

src/Icon/Icon.js
```diff
@@ -15,7 +15,9 @@ export const Icon = ({ glyph, size, clickHandler, className, ...props }) => {
         <span
             {...props}
             className={iconClasses}
-            onClick={clickHandler} />
+            onClick={clickHandler}
+            role='button'
```
bcullman: here (and elsewhere) should these just *be* buttons, rather than have role of button?

greg-a-smith: Agreed.  I don't think we're setting a very good precedent by assigning `<span>` or `<div>` elements a role of `button`.  Because now we've have to wire up our own custom keyboard nav and potentially even focus management.  I know it makes this a more difficult (and time consuming) change.  But similar to parents, I think we need to set a good example.  😄

src/Menu/Menu.js
```diff
@@ -63,7 +63,8 @@ export const MenuItem = ({ url, link, isLink, separator, addon, children, onclic
                 }
                 {(!url && !link) && <a {...linkProps}
                     className='fd-menu__item'
-                    onClick={onclick}>{children}</a>}
+                    onClick={onclick}
+                    role='button'>{children}</a>}
```
greg-a-smith: Although `<a>` elements with a role of `button` "works", is it the best choice?  @jeffredodd What have we typically done with menu items in the past?  Are they anchors or buttons?

bcullman: actually, i think they are supposed to have the role of `menuitem`

I think for things like this, fundamentals should show in their example, the proper roles.

PR #295, opened 2019-01-25
chore: disable anchor a11y rule on documentation pages

jbadan:
### Description
Disabling lint rule on component documentation pages. This is prework for changes needed in ShellBar and Pagination before the lint rule can be fully enabled.

#245


src/Alert/Alert.Component.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
```
bcullman: this is not a documentation page. whats up with that?

PR #307, opened 2019-01-25
fix: gh-pages deployment process

jbadan:
### Description
In order to get proptypes working, we need to be development mode when we deploy to github pages. This is a stop gap measure for the meantime and will be addressed by a future story - either converting to production or lerna.

### DIFF
--- a/config/webpack.config.prod.js
+++ b/config/webpack.config.prod.js
@@ -43,9 +43,9 @@ const env = getClientEnvironment(publicUrl);
 
 // Assert this just to be safe.
 // Development builds of React are slow and not intended for production.
-if (env.stringified['process.env'].NODE_ENV !== '"production"') {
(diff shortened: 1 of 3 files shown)


config/webpack.config.prod.js
```diff
@@ -43,9 +43,9 @@ const env = getClientEnvironment(publicUrl);
 
 // Assert this just to be safe.
 // Development builds of React are slow and not intended for production.
-if (env.stringified['process.env'].NODE_ENV !== '"production"') {
-  throw new Error('Production builds must have NODE_ENV=production.');
-}
+// if (env.stringified['process.env'].NODE_ENV !== '"production"') {
+//   throw new Error('Production builds must have NODE_ENV=production.');
+// }
```
jbadan: I don't love commenting this out, but we need to use the prod config for webpack with NODE_ENV set to development.

PR #309, opened 2019-01-25
fix: update Shellbar to 100% code coverage

chrismanciero:
### Description
Updated Shellbar component's unit test to 100% code coverage


bcullman:
I agree with @greg-a-smith, but it appears right now that this practice is common in this repo. Let's look at refactoring them in a follow up story.

🚢

PR #316, opened 2019-01-29
Fix/date picker code refactor, update unit tests

chrismanciero:
### Description
Updated unit test code coverage for DatePicker component. While creating/updating unit tests also noticed duplicate code being used in the DatePicker component. Refactored duplicate code to new method called **validateDates**. Code coverage is now 98% for DatePicker component. Some code was also removed due to unnecessary condition checks.

### DIFF
--- a/src/DatePicker/DatePicker.js
+++ b/src/DatePicker/DatePicker.js
@@ -31,7 +31,7 @@ export class DatePicker extends Component {
 
     //This is being used only when the user presses enter or clicks outside of the component
     formatDate = (date) => {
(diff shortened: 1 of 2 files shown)


bcullman:
not sure how others feel, but I would prefer unit tests in one PR first, and then a followup PR to refactor.

chrismanciero:
I like that. Should I close this PR and make 2 separate ones. Make a refactor one first and then the unit tests.

greg-a-smith:
@chrismanciero Actually, the first PR would be the unit tests so that there was a testing baseline to help ensure the refactoring doesn't produce unexpected results.

PR #318, opened 2019-01-29
chore: update docs pages with preamble, update nav

### LINKED ISSUE #298: Improve Docs: Add preamble pages (Home / Installation / Usage)
Description

Follow design seen here:

image

CORUI-6167

jbadan:
### Description

Adds home and usage pages to new secondary Getting Started nav.
Adds new MarkdownImporter components that allows us to write documentation pages entirely in markdown.
Updates styles so markdown files will be automatically formatted the same as existing component pages.
A11y changes to left hand navigation (precursor for adding search functionality)
Reformats routes array for consistency
Fixes small typo in Readme

fixes #298


src/_playground/documentation/Markdown/Home.md
```diff
@@ -0,0 +1,38 @@
+# Fundamental React
```
bcullman: hrm.. this looks suspiciously like our readme. ;)

Is there anyway we can this content can live in just one location?

alternatively, perhaps different content here?

src/_playground/Routes.js
```diff
@@ -33,155 +37,309 @@ import { TimePickerComponent } from '../TimePicker/TimePicker.Component';
...
 import React, { Component } from 'react';
 
+
+
 export default class Routes extends Component {
     constructor(props) {
         super(props);
         this.state = {
             routes: [
+                {
```
greg-a-smith: Is there a reason why these are in state?  I don't believe they ever morph so couldn't it just be a simple `const`?

src/_playground/documentation/Markdown/Home.md
greg-a-smith: If it's going to be this content, could we use `node-fetch` (or something like it) to get the `https://github.com/SAP/fundamental-react/blob/master/README.md`?  We could extend the functionality of `MarkdownImporter` to do this fetching.

jbadan: @greg-a-smith love this idea - I'll take a shot at implementing it now

src/_playground/Routes.js
jbadan: no reason, I just left them where I found them. I'll change it

jbadan:
@greg-a-smith @bcullman  I implemented a slightly different approach so that we wouldn't get a readme that corresponds to an rc version on the github pages site if we deploy it out of sync.

Instead of using node-fetch, I have the build script (the one for gh-pages), copy the readme into another file in the `src` directory and then import from there. It is also run on `npm run start` so that it exists in development mode. 

What do you think of this solution?

src/_playground/Routes.js
```diff
@@ -39,139 +41,278 @@ import React, { Component } from 'react';
...
+                                                    key={route.url}
+                                                    to={{ pathname: route.url }}>
+                                                    {route.name}
+                                                </NavLink>
+                                            </li>
+                                        );
+                                    }
+                                })}
+                            </ul>
+                        </nav>
```
greg-a-smith: It looks like you're already using the `sort-by` package to sort the routes by section and then by name so I think you could wrap the whole thing in one `.map` loop.  Start with the first section (`<ul>`), render each route within that section, then close the section each time the `section` value changes and then finish with one last closing of the section.

Also, I don't think you want two separate `<nav>` elements.  I think the whole left nav is one `<nav>` element with a separate list (`<ul>`) for each section.

Issue #300, opened 2019-01-25
Improve Docs: Add dark vs light mode for code examples


bcullman:
### Description

Follow design seen here:
https://sap.github.io/fundamental-ngx/docs/action-bar

![image](https://user-images.githubusercontent.com/5056972/51769968-5171ac00-2099-11e9-8ec3-beffdb9f5933.png)

CORUI-6166

PR #319, opened 2019-01-29
fix: refine toggle component

jbadan:
### Description

The Toggle component was missing an `onChange` callback that was needed for work on #300. 

Toggle was also changed from an uncontrolled element to controlled element, this was also fixed here. Checked prop was previously being spread to the wrong component (the parent div).

Added tests and updated snapshots.

### DIFF
--- a/src/Toggle/Toggle.test.js
+++ b/src/Toggle/Toggle.test.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import renderer from 'react-test-renderer';
+import sinon from 'sinon';
 import { Toggle } from './Toggle';
 import { mount, shallow } from 'enzyme';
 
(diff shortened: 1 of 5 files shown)


src/Toggle/Toggle.test.js
```diff
@@ -53,6 +54,32 @@ describe('<Toggle />', () => {
...
+        test('should have truthy checked state when passed checked prop', () => {
+            const element = mount(<Toggle checked />);
+
+            expect(element.state(['checked'])).toBe(true);
+        });
+    });
+
+    describe('onChange handler', () => {
+        test('should dispatch the onChange callback with the event', () => {
+            let spy = sinon.spy();
```
cjskillingstad: If we don't want to add sinon as a dependency jest has a built in mock function: `jest.fn()` https://jestjs.io/docs/en/mock-functions

PR #322, opened 2019-01-29
fix: CORUI-6120: Add required and default props

### LINKED ISSUE #244: Mark appropriate props as required and use default props
There are a number of locations where props are necessary for functionality, but are not currently marked as required. 

Default props should also be leveraged where appropriate.

Include tests to verify

CORUI-6120

cjskillingstad:
### Description
- Mark props as required that are required for functionality
- Set default props
    - I didn't set default strings as `''`, funcs as `() => {}`, numbers as `null`, objects as `{}`, etc. as it seemed somewhat unnecessary. I did do the bools as false because I think it will make the documentation more readable. Let me know if you think I should do those other cases.
- Update snapshots where necessary

fixes #244


bcullman:
there are a number of props that moved to required here - it would be good to kick out a new release to get updated docs to go along with this PR.

src/Forms/Forms.js
```diff
@@ -98,7 +107,7 @@ export const FormMessage = ({ type, children, className, ...props }) => {
 FormMessage.propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
-    type: PropTypes.oneOf(['', 'error', 'warning', 'help'])
+    type: PropTypes.oneOf(['error', 'warning', 'help'])
```
bcullman: @mobot11 : looks like this was missed in the earlier const PR

bcullman:
really nice cleanup here, @cjskillingstad!

src/Dropdown/Dropdown.js
```diff
@@ -26,6 +26,10 @@ Dropdown.propTypes = {
     standard: PropTypes.bool
 };
 
+Dropdown.defaultProps = {
+    standard: false
+};
+
```
greg-a-smith: We shouldn't need to set `false` for boolean prop types.  The `Properties` component for the documentation pages will automatically show a default of `false` for any boolean prop type, if not set to `true`.

src/Forms/Forms.js
greg-a-smith: We should either fix this or create a follow-up issue.

src/Time/Time.js
```diff
@@ -5,6 +5,32 @@ const INVALID = 'is-invalid';
...
+    };
+    static defaultProps = {
+        arialabel: '',
+        disabled: false,
+        format12hours: false,
+        id: '',
+        spinners: false,
+        value: null
+    };
+
```
greg-a-smith: I believe these were moved to a static reference outside the class to match all other components.

src/TimePicker/TimePicker.js
```diff
@@ -7,6 +7,39 @@ const CLOCK = ['am', 'pm'];
...
+        style: '',
+        isValid: false,
+        length: 0,
+        disabled: false,
+        format12Hours: false,
+        showHour: false,
+        showMinute: false,
+        showSecond: false
+    };
+
```
greg-a-smith: I believe these were moved to a static reference outside the class to match all other components.

src/ActionBar/ActionBar.js
```diff
@@ -73,10 +78,10 @@ export const ActionBarHeader = ({ className, description, descriptionProps, titl
 };
 
 ActionBarHeader.propTypes = {
+    description: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
cjskillingstad: I think it could be optional, but if it is then we shouldn't render that empty `<p>` tag at all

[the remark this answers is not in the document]
greg-a-smith: Agree 100%.  I think it would be safer to leave it optional and then just refactor the component to either write (or not write) the `<p>` element.

src/ActionBar/ActionBar.js
```diff
@@ -27,6 +27,11 @@ ActionBar.propTypes = {
     width: PropTypes.string
 };
 
+ActionBar.defaultProps = {
+    mobile: false,
```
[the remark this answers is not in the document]
jbadan: I think we'll need to add back some of the logic that defaults to the props to false if we revert this (like in Toggle).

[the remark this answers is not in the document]
cjskillingstad: Removed all false default prop types in af7bd50. Also added defensive logic in `Toggle` to fix that.

src/Calendar/Calendar.js
```diff
@@ -579,7 +579,14 @@ Calendar.propTypes = {
     tableBodyProps: PropTypes.object,
     tableHeaderProps: PropTypes.object,
     tableProps: PropTypes.object,
-    yearListProps: PropTypes.object
+    yearListProps: PropTypes.object,
+    onChange: PropTypes.func
```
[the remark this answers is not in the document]
cjskillingstad: Added default props for `func`'s and removed defensive coding surrounding them if it existed in bdbca88

src/ActionBar/ActionBar.js
greg-a-smith: Looks like there are just a couple default props set to false remaining:

* `ActionBar`: `mobile`
* `TimeItem`: `disabled`, `format12hours `, `spinners `

src/Pagination/Pagination.js
```diff
@@ -89,7 +89,7 @@ export class Pagination extends Component {
         // calculate number of pages based on item total and items per page
         // check to make sure itemsPerPage != 0
         this.numberOfPages = Math.ceil(
-            itemsTotal / (itemsPerPage ? itemsPerPage : 10)
+            itemsTotal / (itemsPerPage || 10)
```
greg-a-smith: With a default value, we could remove the `|| 10`.

cjskillingstad: I thought about this, but if the consumer passes in a `0` then this will break

greg-a-smith: Fair enough.  If it would break with certain values, then we should consider adding a custom prop type validator that would prevent that.  But, that would be a follow-up enhancement if we wanted to do that.

cjskillingstad:
@mobot11 @greg-a-smith Using const's for the last couple places we missed in the propTypes in e09cb71

PR #327, opened 2019-01-30
fix: issue #181 DatePicker disabled date selected

### LINKED ISSUE #181: Datepicker: disabled dates still selectable
Dates are still selectable via the calendar overlay when disabled.

See playground example:

chrismanciero:
### Description
Add to check to see if Day being added to calendar should be disabled or not. If disabled, remove onClick functionality so the disabled date cannot be selected.

fixes #181


jbadan:
Should we add a test for this as well?

chrismanciero:
@jbadan - unit test added for selecting a disabled date and the current selected date does not change

src/Calendar/Calendar.js
```diff
@@ -465,21 +465,16 @@ export class Calendar extends Component {
...
-                    </td>);
-                } else {
-                    days.push(<td className={dayClasses} key={copyDate}
-                        onClick={() => this.dateClick(copyDate, enableRangeSelection)} role='gridcell'><span className='fd-calendar__text'>{dateFormatted}</span>
-                    </td>);
-                }
+                days.push(<td className={dayClasses} key={copyDate}
+                    onClick={!this.displayDisabled(day) ? () => this.dateClick(copyDate, enableRangeSelection) : null
+                    } role='gridcell' > <span className='fd-calendar__text'>{dateFormatted}</span>
+                </td >);
```
greg-a-smith: I like this much better -- less chance that a change to calendar days get missed.

Minor nit:  from a code readability standpoint, I think I'd reformat and put a prop on each line like:
```
days.push(
    <td
        className={dayClasses}
        key={copyDate}
        onClick={!this.displayDisabled(day) ? () => this.dateClick(copyDate, enableRangeSelection) : null}
        role='gridcell'>
        <span className='fd-calendar__text'>{dateFormatted}</span>
    </td>
);
```

chrismanciero: 👍 yes it does look better, done

Issue #317, opened 2019-01-29
Refactor .babelrc to compile production builds correctly


mobot11:
### Description
The code in `dist` after running a build still contains import statements.

```
import _defineProperty from "@babel/runtime/helpers/esm/defineProperty";
import _objectWithoutProperties from "@babel/runtime/helpers/esm/objectWithoutProperties";
import _classCallCheck from "@babel/runtime/helpers/esm/classCallCheck";
import _createClass from "@babel/runtime/helpers/esm/createClass";
import _possibleConstructorReturn from "@babel/runtime/helpers/esm/possibleConstructorReturn";
import _getPrototypeOf from "@babel/runtime/helpers/esm/getPrototypeOf";
import _inherits from "@babel/runtime/helpers/esm/inherits";
import { ALERT_TYPES } from '../utils/constants';
import classnames from 'classnames';
import React, { Component } from 'react';
```

This will break in browsers that don't support the import syntax.

Ideally post-babel our imports would look something more like this: 
```
"use strict";

var _interopRequireDefault = require("@babel/runtime-corejs2/helpers/interopRequireDefault");

exports.__esModule = true;
exports.default = void 0;

var _values = _interopRequireDefault(require("@babel/runtime-corejs2/core-js/object/values"));

var _extends3 = _interopRequireDefault(require("@babel/runtime-corejs2/helpers/extends"));

var _objectWithoutPropertiesLoose2 = _interopRequireDefault(require("@babel/runtime-corejs2/helpers/objectWithoutPropertiesLoose"));

var _inheritsLoose2 = _interopRequireDefault(require("@babel/runtime-corejs2/helpers/inheritsLoose"));

var _classnames = _interopRequireDefault(require("classnames"));

var _react = _interopRequireDefault(require("react"));

var _propTypes = _interopRequireDefault(require("prop-types"));

var _bootstrapUtils = require("./utils/bootstrapUtils");

var _StyleConfig = require("./utils/StyleConfig");

var _CloseButton = _interopRequireDefault(require("./CloseButton"));
```

### Versions
**fundamental-react:**

---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

mobot11:
[...]

PR #330, opened 2019-01-31
chore: update babel config and build

mobot11:
### Description
Library now compiles to the `lib` directory. Also compiles to commonjs. 

fixes https://github.com/SAP/fundamental-react/issues/317


README.md
```diff
@@ -29,14 +29,28 @@ For an existing react application, follow the steps bellow:
...
+    ```scss
     @import '~fiori-fundamentals/dist/fiori-fundamentals.min.css';
     ```
 
 1. Import components as needed. See [Component Documentation](https://sap.github.io/fundamental-react/) for examples and API details.
-    ```
-    import { Alert } from 'fundamental-react';
-    ```
+```javascript
+import { Alert } from 'fundamental-react/lib/Alert/Alert';
```
jeffredodd: Shouldn't `import Alert from 'fundamental-react/lib/Alert/Alert'` also work in this case?

babel.config.js
```diff
@@ -0,0 +1,48 @@
...
+    env: {
+        production: {
+            presets: [
+                [
+                    '@babel/preset-env',
+                    {
+                        targets: {
+                            ie: '11'
+                        }
+                    }
```
jeffredodd: We should probably be using browserlist to set `@babel/preset-env`. https://github.com/SAP/fundamental-react/blob/master/package.json#L130 or maybe removal it all together if its not used anywhere but for babel. This was inherited from Create React App.

README.md
greg-a-smith: @jeffredodd As of now anyway, there are no default exports in any of the source modules so named exports are all we have to work with.

mobot11: I don't think so since they aren't default exports.

babel.config.js
```diff
@@ -0,0 +1,48 @@
...
+                ]
+            ],
+            plugins: productionPlugins
+        }
+    }
+};
```
greg-a-smith: I have basically no idea what any of this is doing. 😏 I'd love a brief walkthrough of this config.

PR #332, opened 2019-01-31
fix: DatePicker update unit tests

chrismanciero:
### Description
Update unit tests for DatePicker component

### DIFF
--- a/src/DatePicker/DatePicker.test.js
+++ b/src/DatePicker/DatePicker.test.js
@@ -42,22 +42,11 @@ describe('<DatePicker />', () => {
         // check to make sure calendar is shown
         expect(wrapper.state('hidden')).toBeFalsy();
 
-        // click to hide calendar
+        // clicking on input text should keep calendar displayed
         wrapper.find('input[type="text"]').simulate('click', { type: 'input' });
 
         // check to make sure calendar is shown
         expect(wrapper.state('hidden')).toBeFalsy();
-
-        wrapper.instance().componentWillMount();
-
-        // click on body element
-        let event = new MouseEvent('mousedown', {
-            target: document.querySelector('body')
-        });
-        document.dispatchEvent(event);
-
-        // check to make sure calendar is hidden
-        expect(wrapper.state('hidden')).toBeTruthy();
 
         // click to show calendar
         wrapper.find('input[type="text"]').simulate('click', { type: '' });
@@ -134,11 +123,14 @@ describe('<DatePicker />', () => {
         expect(wrapper.state('formattedDate')).toEqual(switchFormattedDate);
         expect(wrapper.state('arrSelectedDates').length).toEqual(2);
 
-        // click on body element
+        // click on body element to hide calendar
         let event = new MouseEvent('mousedown', {
             target: document.querySelector('body')
         });
         document.dispatchEvent(event);
+
+        // check to make sure calendar is hidden
+        expect(wrapper.state('hidden')).toBeTruthy();
     });
 
     test('entering invalid range dates', () => {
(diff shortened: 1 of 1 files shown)


src/DatePicker/DatePicker.test.js
```diff
@@ -33,27 +33,36 @@ describe('<DatePicker />', () => {
...
+        // check to make sure calendar is hidden
         expect(wrapper.state('hidden')).toBeTruthy();
 
+        // click to show calendar
         wrapper.find('input[type="text"]').simulate('click', { type: 'input' });
 
+        // check to make sure calendar is shown
         expect(wrapper.state('hidden')).toBeFalsy();
 
+        // click to hide calendar
```
greg-a-smith: This _seems_ like the wrong comment.  It's simulating a click in the input, which based on the next assertion, should _not_ hide the calendar.

src/DatePicker/DatePicker.test.js
```diff
@@ -62,33 +71,57 @@ describe('<DatePicker />', () => {
...
         let arrDates = [startRangeDate, endRangeDate];
         wrapper.instance().updateDate(arrDates);
 
+        expect(wrapper.state('arrSelectedDates').length).toEqual(2);
+
+        // click on body element
+        let event = new MouseEvent('mousedown', {
+            target: document.querySelector('body')
+        });
         document.dispatchEvent(event);
```
greg-a-smith: I don't understand why the added `mousedown` event, but then no additional assertions.  I'm probably just missing something.

src/DatePicker/DatePicker.test.js
```diff
@@ -101,10 +134,34 @@ describe('<DatePicker />', () => {
         expect(wrapper.state('formattedDate')).toEqual(switchFormattedDate);
         expect(wrapper.state('arrSelectedDates').length).toEqual(2);
 
+        // click on body element
+        let event = new MouseEvent('mousedown', {
+            target: document.querySelector('body')
+        });
         document.dispatchEvent(event);
```
greg-a-smith: I don't understand why the added `mousedown` event, but then no additional assertions.  I'm probably just missing something.

src/DatePicker/DatePicker.test.js
```diff
@@ -175,20 +232,25 @@ describe('<DatePicker />', () => {
 
         wrapper.find('input[type="text"]').simulate('keypress', { key: 'Enter' });
 
+
+        // press Esc key
+        wrapper.find('input[type="text"]').simulate('keypress', { key: 'Esc' });
```
greg-a-smith: Shouldn't each `test` block end with an assertion?

src/DatePicker/DatePicker.test.js
chrismanciero: @greg-a-smith - thanks for noticing, i just updated the comment

src/DatePicker/DatePicker.test.js
chrismanciero: removed MouseEvent code since it was not necessary

src/DatePicker/DatePicker.test.js
chrismanciero: added assertion to check that calendar is hidden

src/DatePicker/DatePicker.test.js
chrismanciero: Added assertions at the end of test

src/DatePicker/DatePicker.test.js
chrismanciero: sorry, missed that mouseDown event. Just added an assertion to check that calendar is hidden.

PR #294, opened 2019-01-25
feat: Remove React Router from Breadcrumbs

jeffredodd:
### Description
In order to remove the dependency on React Router from Fundamentals-React (#115), we must refactor the the following components...

- Breadcrumb (#294)
- Menu (#293)
- Tabs (Coming Soon)
- SideNavigation (Coming Soon)

This PR is focused on **Menu** and does introduce breaking changes such as...

- Removal of Link prop

**React Router Breadcrumb Example (NEW)**
```
  <Breadcrumb>
    <BreadcrumbItem>
      <Link to='#'>Link Text</Link>
    </BreadcrumbItem>
    <BreadcrumbItem>
      <Link to='#'>Link Text</Link>
    </BreadcrumbItem>
    <BreadcrumbItem>
      <Link to='#'>Link Text</Link>
    </BreadcrumbItem>
  </Breadcrumb>
```

**Breadcrumbs via Anchors children (NEW)**
```
  <Breadcrumb>
    <BreadcrumbItem>
      <a href='#'>Link Text</a>
    </BreadcrumbItem>
    <BreadcrumbItem>
      <a href='#'>Link Text</a>
    </BreadcrumbItem>
    <BreadcrumbItem>
      <a href='#'>Link Text</a>
    </BreadcrumbItem>
  </Breadcrumb>
```

**Breadcrumbs via Props (Classic)**
```
<Breadcrumb>
    <BreadcrumbItem name='Link Text' url='#' />
     <BreadcrumbItem name='Link Text' url='#' />
     <BreadcrumbItem name='Link Text' url='#' />
</Breadcrumb>
```

PR #293, opened 2019-01-25
feat: Remove React Router from Menu

jeffredodd:
### Description
In order to remove the dependency on React Router from Fundamentals-React (#115), we must refactor the the following components...

- Breadcrumb (#294)
- Menu (#293)
- Tabs (Coming Soon)
- SideNavigation (Coming Soon)

This PR is focused on **Menu** and does introduce breaking changes such as...

- Removal of Link prop
- Removal of LinkProps prop

**React Router Menu Example (NEW)**
```
<Menu>
    <MenuList>
        <MenuItem>
            <Link to='/'>Option 1</Link>
        </MenuItem>
        <MenuItem>
            <Link to='/'>Option 2</Link>
        </MenuItem>
        <MenuItem>
            <Link to='/'>Option 3</Link>
        </MenuItem>
    </MenuList>
</Menu>
```

**Anchor via children (NEW)**
```
<Menu>
    <MenuList>
        <MenuItem>
            <a href='/'>Option 1</a>
        </MenuItem>
        <MenuItem>
            <a href='/'>Option 2</a>
        </MenuItem>
        <MenuItem>
            <a href='/'>Option 3</a>
        </MenuItem>
    </MenuList>
</Menu>
```

**Anchor via Props (Classic)**
```
<Menu>
    <MenuList>
        <MenuItem url="/">Option 1</MenuItem>
        <MenuItem url="/">Option 2</MenuItem>
        <MenuItem url="/">Option 3</MenuItem>
        <MenuItem url="/">Option 4</MenuItem>
    </MenuList>
</Menu>
```

Issue #315, opened 2019-01-29
Tabs component not suitable for use


magicmatatjahu:
### Description

At the moment `Tabs` component not suitable for use (for dev and prod) -> see the attachment. 
**Hint from me**: remove react-router dependency from this component and change code to something like this: 
https://github.com/kyma-project/console/blob/master/components/react/src/components/Tabs/index.js

If you want, I can create PR with fix (and/or new architecture of component) to Sunday.

### Versions
**fundamental-react:** 0.3.0-rc.0
**fiori-fundamentals:** 1.4.1

---

_**NOTE:** Where applicable, please include uncropped screen captures._

![image](https://user-images.githubusercontent.com/20404945/51899414-3cf02500-23b3-11e9-9252-bc57d8b543f7.png)

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

jeffredodd:
There is an upcoming change related to #115 that will improve the Tab API and separate its dependecy on React Router.

magicmatatjahu:
@jeffcollier Thank you for you comment! In what form will the new api be and when will it be realized? I do not hide that tabs are needed in my project. If I created PR with a new tabs architecture, would you mind?

jeffredodd:
This is resolved by #336. See #344 for issues with sizing.

PR #336, opened 2019-01-31
fix: Refactor tabs to remove React-Router dependency 

jeffredodd:
### Description
In order to remove the dependency on React Router from Fundamentals-React (#115), we must refactor the the following components...

- Breadcrumb (#294)
- Menu (#293)
- Tabs (#336)
- SideNavigation (Coming Soon)

This PR is focused on **Tab** and does introduce breaking changes such as...

- Refactoring of JSX components
- `<TabComponent>` now encapsulates `<Tab>` 
- `<Tab>` no longer creates its own `<Link>` but instead allows it to be added via an HOC pattern. 

Since we needed to support Tabs ingesting `<Link>` via a higher order component, there was a major change to the `<JSX>` API. I'm open to renaming any component, just give a shoutout!

**Tab Component**
```
<TabComponent>
        <Tab
            content='Hello world'
            id='1'
            url='#'>
            Tab 1
        </Tab>
        <Tab
            content='Hello world 2'
            id='2'
            url='#'>
            Tab 2
        </Tab>
        <Tab
            content='Hello world 3'
            disabled
            id='3'
            url='#'>
            Tab 3
        </Tab>
    </TabComponent>
```

**Tab Component w/ Anchors**
```
 <TabComponent>
        <Tab
            content='Hello world'
            id='1'>
            <a
                href='#'>
                Tab 1
            </a>
        </Tab>
        <Tab
            content='Hello world 2'
            id='2'>
            <a
                href='#'>
                Tab 2
            </a>
        </Tab>
        <Tab
            content='Hello world 3'
            disabled
            id='3'>
            <a
                href='#'>
                Tab 2
            </a>
        </Tab>
    </TabComponent>
```

**Tab Component w/ `<Link>`**
```
<MemoryRouter>
        <TabComponent>
            <Tab
                content='Hello world'
                id='1'>
                <Link
                    to='/'>
                    Tab 1
                </Link>
            </Tab>
            <Tab
                content='Hello world 2'
                id='2'>
                <Link
                    to='/'>
                    Tab 2
                </Link>
            </Tab>
            <Tab
                content='Hello world 3'
                disabled
                id='3'>
                <Link
                    to='/'>
                    Tab 2
                </Link>
            </Tab>
        </TabComponent>
    </MemoryRouter>
```


src/_playground/documentation/Playground/Playground.js
```diff
@@ -434,9 +434,9 @@ export class Playground extends Component {
               break;
           case 'tabs':
               componentToGenerate = (
-                  <Tabs>
+                  <Tab>
                       <TabComponent ids={this.state.childs.children.props.ids} />
-                  </Tabs>
+                  </Tab>
```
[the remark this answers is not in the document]
jeffredodd: Missed this, this was a playground example that we are no longer using. I've moved this for now.

jeffredodd:
@greg-a-smith, all feedback should be up to date now 😄

src/Tabs/Tabs.js
```diff
@@ -1,111 +1,129 @@
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
-import { BrowserRouter, Link } from 'react-router-dom';
 import React, { Component } from 'react';
 
-export const Tabs = props => {
-    const { children, className, ...rest } = props;
+export const Tab = props => {
+    const { children, className, content, disabled, id, onClick, selectedId, tabContentProps, tabLinkProps, url, ...rest } = props;
```
cjskillingstad: Looks like some of these are missing in the `propTypes` definition. This could be an argument to use classes rather than SFC's.

src/Tabs/Tabs.js
```diff
@@ -1,111 +1,129 @@
...
+        if (url) {
+            return (
+                <a
+                    {...tabLinkProps}
+                    aria-disabled={disabled}
+                    className={className}
+                    disabled={disabled}
+                    href={url}
+                    onClick={(e) => {
+                        if (!disabled && onClick) {
```
cjskillingstad: If we default `onClick: () => {}` we won't have to check if its defined.

src/Tabs/Tabs.js
```diff
@@ -1,111 +1,129 @@
...
+                </a>
+            );
+        } else if (children && React.isValidElement(children)) {
+            return React.cloneElement(children, {
+                ...tabLinkProps,
+                'aria-disabled': disabled,
+                className: className,
+                disabled: disabled,
+                onClick: (e) => {
+                    if (!disabled && onClick) {
```
cjskillingstad: Same as above.

src/Tabs/Tabs.js
greg-a-smith: It looks like there are a number of these props that are not defined in the `propTypes` or `propDescriptions`.  The `Tab` component is only showing `id`, `children`, `className` and `url`.

src/Tabs/Tabs.js
```diff
@@ -1,111 +1,129 @@
...
-        'fd-tabs-container',
-        className
-    );
+    const renderLink = () => {
+        if (url) {
+            return (
+                <a
+                    {...tabLinkProps}
+                    aria-disabled={disabled}
+                    className={className}
```
greg-a-smith: Does the `className` prop go here or should that go on the `<li>`?  If they really wanted a `className` on the `<a>` or `Link`, they could pass it through `tabLinkProps`.

src/Tabs/Tabs.js
```diff
@@ -1,111 +1,129 @@
...
+                        }
+                    }}>
+                    {children}
+                </a>
+            );
+        } else if (children && React.isValidElement(children)) {
+            return React.cloneElement(children, {
+                ...tabLinkProps,
+                'aria-disabled': disabled,
+                className: className,
```
greg-a-smith: Same as the comment above on the `<a>` for when `url` is passed.

src/Tabs/Tabs.js
```diff
@@ -1,111 +1,129 @@
...
             }
         );
     }
 
+    renderTabs = (children) => {
+        return React.Children.map(children, (child) => {
+            const classes = this.getLinkClasses(child.props.id, child.props.className);
+
+            return React.cloneElement(child, {
+                selectedId: this.state.selectedId,
```
greg-a-smith: Instead of passing the `selectedId` to each child `Tab`, what if the comparison of `child.props.id` to `selectedId` occurred here and the `Tab` component was simply passed a `selected` prop?  Keep more of the decision making in this component (that is carrying state) and leave the `Tab` as a dumb component.

src/Tabs/Tabs.js
```diff
@@ -1,111 +1,129 @@
...
 };
 
-TabComponent.propDescriptions = {
-    ids: 'An array of objects with keys \'id\', \'url\', \'name\', \'hasChild\', \'child\', and \'glyph\' defining each tab.',
-    tabContentProps: 'Additional props to be spread to the tab\'s content.',
-    tabLinkProps: 'Additional props to be spread to the tab\'s link.',
-    tabProps: 'Additional props to be spread to the tab.'
+TabGroup.propDescriptions = {
+    children: 'One or more Tab components to render within the component.',
+    selectedId: 'Initial selected `<Tab>`'
```
greg-a-smith: Maybe reword to "The `id` of the selected `Tab`."

src/Tabs/Tabs.js
```diff
@@ -1,111 +1,129 @@
...
+    selectedId: PropTypes.string
 };
 
-TabComponent.propDescriptions = {
-    ids: 'An array of objects with keys \'id\', \'url\', \'name\', \'hasChild\', \'child\', and \'glyph\' defining each tab.',
-    tabContentProps: 'Additional props to be spread to the tab\'s content.',
-    tabLinkProps: 'Additional props to be spread to the tab\'s link.',
-    tabProps: 'Additional props to be spread to the tab.'
+TabGroup.propDescriptions = {
+    children: 'One or more Tab components to render within the component.',
```
greg-a-smith: "Tab" should be backticked.

greg-a-smith:
@jeffredodd There is an issue (#315) where the tab content appears misaligned for Tab 2.  I am seeing this in the documentation examples.  Is this an easy fix that could be added to this?

src/Tabs/Tabs.js
jeffredodd: This functionality was not present in the last version of tabs so lets put it on the `<li>` and if a user needs it that can add it to `tabLinkProps` to get it to the anchor.

jeffredodd:
@greg-a-smith So when doing research it appeared that the [Fundamental Tabs](https://sap.github.io/fundamental/components/tabs.html) don't currently document how they support content. 

The jittering I think we are observing is due to the content of a tab being loaded into the `<li>` with a `<p>` which is not the right place for it. I didn't change it in this PR as I was trying to alter as little outputted markup as possible.

PR #349, opened 2019-02-04
fix: Added FocusTrap to Modal component

### LINKED ISSUE #345: Modal does not contain Focus Cycling mechanism
### Description
In order to be WCAG 2.1 compliant, our Modal's with the `role='dialog'` must contain a mechanism to enforce focus cycling. A users focus must be contained within a modal to be compliant. See https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal for more information.  

### Reproduction
1) Open a modal
2) Attempt to tab

Expected:
- Focus should remain within modal
- Focus should be redirected into modal when opened

Actual:
- Focus is not altered and left in last known state. 

_Note: This issue is seen in all 3 libraries_
Fundamental-React: SAP/fundamental-react#345
Fundamental-NGX: SAP/fundamental-ngx#477
Fundamental-Vue: SAP/fundamental-vue#136

chrismanciero:
### Description
Added [FocusTrap](https://github.com/davidtheclark/focus-trap-react) react library to Modal component to handle Focus trapping when cycling through Modal box. 

Added attributes to Modal dialog box:
- aria-label={title}
- aria-modal='true'
- role='dialog'

fixes #345

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
   "dependencies": {
     "classnames": "^2.2.6",
     "fiori-fundamentals": "^1.4.0",
+    "focus-trap-react": "^6.0.0",
     "react-router": "^4.3.1",
     "react-router-dom": "^4.3.1",
     "react-syntax-highlighter": "^9.0.1"
@@ -81,7 +82,6 @@
     "eslint-plugin-loosely-restrict-imports": "^0.1.15",
     "eslint-plugin-react": "7.11.1",
     "file-loader": "2.0.0",
-    "focus-trap-react": "^6.0.0",
     "fork-ts-checker-webpack-plugin-alt": "0.4.14",
     "fs-extra": "7.0.0",
     "gh-pages": "^2.0.1",
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "eslint-plugin-loosely-restrict-imports": "^0.1.15",
     "eslint-plugin-react": "7.11.1",
     "file-loader": "2.0.0",
+    "focus-trap-react": "^6.0.0",
     "fork-ts-checker-webpack-plugin-alt": "0.4.14",
     "fs-extra": "7.0.0",
     "gh-pages": "^2.0.1",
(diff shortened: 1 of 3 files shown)


src/Modal/Modal.js
```diff
@@ -43,43 +44,49 @@ export class Modal extends Component {
         }
 
         return ReactDOM.createPortal(
-            <div
-                {...rest}
-                className={modalClasses}>
-                <div className='modal-demo-bg'>
-                    <div className='fd-modal'>
+            <FocusTrap>
```
bcullman: @chrismanciero: does focusTrap generate an additional layer of `<div>`?

package.json
```diff
@@ -81,6 +81,7 @@
     "eslint-plugin-loosely-restrict-imports": "^0.1.15",
     "eslint-plugin-react": "7.11.1",
     "file-loader": "2.0.0",
+    "focus-trap-react": "^6.0.0",
```
bcullman: @droshev, @jeffredodd :

focus-trap-react is a light-weight wrapper around `focus-trap` https://github.com/davidtheclark/focus-trap

Angular implementation: https://github.com/hassam7/ngx-focus-trap
Vue implementation: https://github.com/wovue/focus-trap

jeffredodd: Yeah I would agree that consistency across the difference platforms will help us keep the A11y focus trapping the same in all 3 libraries.

jeffredodd:
Just one general thing to call out, it looks like this library allows focus to be trapped and cycle around a component, what about when a dialog has no actionable elements say something like a modal that says loading. 

One thing we did with SAP Concur's NUI-Widgets was added the ability to focus on a blank `<span index='-1'>` which was able to get focus, wasn't read out-loud, and redirected the focus of the consumer into the modal without disruption. it also eliminated the issue of a focus outline being initially triggered on the first actionable element it found. 

Is this something we could support with this library? @chrismanciero @bcullman thoughts?

chrismanciero:
@jeffredodd Yes Focus-trap-react has the functionality that you can set the starting tab element. You can see from my PR I created an element <span className='startTabIndex' tabIndex='-1' /> and set the initialFocus to this element. Now when the modal loads the Close (X) is not outlined until the user hits the tab key to cycle through the Modal component.

package.json
bcullman: should this be a `Dep`, rather than a `devDep`?

chrismanciero: Moved to dependencies

Issue #297, opened 2019-01-25
Improve Docs: Add Link to Github


bcullman:
### Description
Add a link (with octocat icon) to github

Follow design seen here:
https://sap.github.io/fundamental-ngx/docs/home

![image](https://user-images.githubusercontent.com/5056972/51769792-c2fd2a80-2098-11e9-82bf-7d3a1d4f1b6d.png)

CORUI-6168

PR #355, opened 2019-02-05
chore: Documentation - add link to GitHub

mjaycub:
### Description
Added the Github logo which links to the fundamental-react repo to the documentation.

Issue: #297


src/_playground/documentation/Header/Header.js
```diff
@@ -4,10 +4,21 @@ import React from 'react';
...
 
+    const goToGithub = () => {
+        window.location.href = 'https://github.com/SAP/fundamental-react';
+    };
+
     return (
-        <h1 className='frDocs-Content__header'>
-            {children}
-        </h1>
+        <div>
```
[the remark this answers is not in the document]
mjaycub: Yes of course 👍

src/_playground/App.scss
```diff
@@ -73,6 +73,17 @@ $fd-fonts-path: '../../node_modules/fiori-fundamentals/dist/fonts/';
...
+            .frDocs-Content__githubLogo {
+                float: right;
+                z-index: 100;
+                cursor: pointer;
+            }
+
+            .frDocs-Content__githubLogo svg {
+                width: 2rem;
+                height: 2rem;
+            }
```
greg-a-smith: You can write this cleaner since it's a Sass file.
```scss
.frDocs-Content__githubLogo {
    float: right;
    z-index: 100;
    cursor: pointer;

    svg {
        width: 2rem;
        height: 2rem;
    }
}
```

greg-a-smith:
@mjaycub I like the link you added.  I just think this should be lined up better in relation to the H1 text (see screenshot below).

screen shot 2019-02-06 at 11 09 07 am

I think this image should be centered vertically like the other text.

mjaycub:
> @mjaycub I like the link you added. I just think this should be lined up better in relation to the H1 text (see screenshot below).
> 
> screen shot 2019-02-06 at 11 09 07 am
> 
> I think this image should be centered vertically like the other text.

This was a little more awkward than I thought it would be. I moved the SVG out of the `<h1>` tag so that we can hide the `<h1>` on the Home page by using the `:empty` selector which avoids the difference in alignment between home page and component page.

I also had to set the `line-height` for the `.frDocs-Content__header` - I'd like to hear other opinions on this or alternative cleaner ways to align the svg in the header if this is not ideal.

src/_playground/App.scss
```diff
@@ -73,10 +73,29 @@ $fd-fonts-path: '../../node_modules/fiori-fundamentals/dist/fonts/';
...
 
             .frDocs-Content__header, .frDocs-markdown > h1 {
                 margin: 0 0 12px;
                 font-size: 2.2rem;
+                line-height: 2.2rem;
+            }
+
+            .frDocs-Content__header:empty {
+                display: none;
```
greg-a-smith: The display looks good now.  My only remaining question is what is the purpose of this `:empty` block?  If I remove it, the logo still appears correct.

src/_playground/App.scss
```diff
@@ -73,10 +73,29 @@ $fd-fonts-path: '../../node_modules/fiori-fundamentals/dist/fonts/';
...
+                svg {
+                    width: 2rem;
+                    height: 2rem;
+                }
+            }
 
             .frDocs-Content__header, .frDocs-markdown > h1 {
                 margin: 0 0 12px;
                 font-size: 2.2rem;
+                line-height: 2.2rem;
```
greg-a-smith: TIP: If you are desiring the line height to match the font size, you could just enter it as `line-height: 1;` which is a ratio setting so if the font size ever changes, the line height will change as well.  Makes for easier maintenance going forward.

src/_playground/App.scss
mjaycub: On the homepage, since there is no header text for the `<h1>` tag, the logo will be misaligned (see below). The first title on homepage comes from the markdown.

screen shot 2019-02-07 at 23 10 59


By using the `:empty` we can hide the `<h1>` and the space it takes on the homepage only (when empty) and have the logo line up with the markdown title.

screen shot 2019-02-07 at 23 06 15

greg-a-smith: Ok, I see now.  Actually, there are two `<h1>` tags -- that's what was confusing me.  The empty one with the class `frDocs-Content__header` and then the `<h1>` that gets pulled in from the markdown file (with no class on it).

src/_playground/documentation/Header/Header.js
```diff
@@ -5,9 +5,18 @@ export const Header = (props) => {
     const { children } = props;
 
     return (
-        <h1 className='frDocs-Content__header'>
-            {children}
-        </h1>
+        <React.Fragment>
+            <a className='frDocs-Content__githubLogo' href='https://github.com/SAP/fundamental-react'
```
greg-a-smith: If you really wanted to get fancy, you could import the `package.json` file and use `repository.url` for the value in the anchor's `href` attribute.

mjaycub: Done!

src/_playground/App.scss
mjaycub: Yep exactly, so we're hiding the empty `frDocs-Content__header`

PR #363, opened 2019-02-05
fix: update currentYear on date selection

### LINKED ISSUE #357: Calendar Component: currentYear should be updated on date selection for consistentcy
### Description

Related to: https://github.com/SAP/fundamental-react/issues/334

The same problem is present with the currentYear state. 

### Versions
**fiori-fundamentals:**

---

_**DISCLAIMER:**

chrismanciero:
### Description
Set the currentYear value to the selected date object. 

fixes #357

### DIFF
--- a/src/Calendar/Calendar.js
+++ b/src/Calendar/Calendar.js
@@ -139,6 +139,7 @@ export class Calendar extends Component {
         if (!this.props.enableRangeSelection) {
             this.setState({
                 currentDateDisplayed: date,
+                currentYear: date,
                 selectedDate: date,
                 dateClick: true
             }, function() {
@@ -147,6 +148,7 @@ export class Calendar extends Component {
         } else {
             this.setState({
                 currentDateDisplayed: date,
+                currentYear: date,
                 dateClick: true
             });
         }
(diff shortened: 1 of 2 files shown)


src/Calendar/Calendar.js
```diff
@@ -139,6 +139,7 @@ export class Calendar extends Component {
         if (!this.props.enableRangeSelection) {
             this.setState({
                 currentDateDisplayed: date,
+                currentYear: date,
                 selectedDate: date,
```
greg-a-smith: Is it strange that we have 3 different state variables (`currentDateDisplayed`, `currentYear`, `selectedDate`) that are all being set to the same value?  Seems like they should either be set to different values or the code that depends on them should just reference a single stored value.

jbadan: @greg-a-smith I agree, but I think it might be better to save that work for an overall refactoring effort instead of putting more time into this. Thoughts?

greg-a-smith: @jbadan Agreed.  That's probably a much larger effort.

chrismanciero: I agree we should refactor this control. as a side not i did remove the currentYear property and updated the code and tests to work off the currentDateDisplayed property instead 😄

PR #367, opened 2019-02-05
chore: Refined release process more and implemented github-assistant

greg-a-smith:
### Description
As I was beginning to implement `create-release.js` and `release-notes.js` in other repositories, it became apparent that those should be part of their own package that can be reused across multiple repos.  There is now a `github-assistant` package with this functionality in it.  There are also a few enhancements to the release note generation such as continuing to fetch commits until a matching previous version was found despite the 30 commit limit per API endpoint call.


.npmrc
```diff
@@ -0,0 +1,3 @@
+username=${NPM_USERNAME}
+email=${NPM_EMAIL}
+//registry.npmjs.org/:_authToken=${AUTH_TOKEN}
```
greg-a-smith: Eliminates the need for another script to generate this file.

.travis.yml
```diff
@@ -8,22 +8,32 @@ script:
...
+    before_deploy:
+    - git config --add remote.origin.fetch +refs/heads/*:refs/remotes/origin/*
+    - git fetch
+    - git checkout master
+    deploy:
+    - provider: script
+      script: bash ./ci-scripts/publish.sh
+      skip_cleanup: true
+      on:
+        all_branches: true
```
greg-a-smith: Changes to allow separate scripts for publishing rc vs. non-rc version -- keeps the scripts much cleaner.

package.json
```diff
@@ -85,6 +85,7 @@
     "fs-extra": "7.0.0",
     "gh-pages": "^2.0.1",
     "github-api": "^3.0.0",
+    "github-assistant": "^0.2.0",
```
greg-a-smith: New home for `create-release.js` and `release-notes.js`.

.travis.yml
```diff
@@ -8,22 +8,32 @@ script:
...
 notifications:
   email:
     on_failure: always
     on_success: change
   slack: 
     rooms:
-      - ui-fundamentals:zVRyxlqkYnQFD7p8QHkWM8Fd#react-automated
\ No newline at end of file
+      - ui-fundamentals:zVRyxlqkYnQFD7p8QHkWM8Fd#react-automated
```
bcullman: unneeded diff here.

there should be a lint rule for this?

greg-a-smith: I just added a new line at the end of the file.  I don't believe we have linted `yml` files before.

.travis.yml
```diff
@@ -8,22 +8,32 @@ script:
   - npm run build
 jobs:
   include:
-  - stage: "Fundamental-react: Test & Lint"
+  - stage: Lint and Test
     script: npm run lint && npm run test:coveralls && npm run size
-  - stage: Deploy
-    if: (branch = master OR branch = tmp_branch_for_automated_release_do_not_use) AND type = push
+  - stage: Pre-release
+    if: branch = master
```
jbadan: We should  add `AND type = push` here so that it doesn't do a pre-release when the PR travis run builds.

.travis.yml
bcullman: true. no biggie.

.travis.yml
jbadan: Nevermind! Travis is smarter than I thought it was 🤓

bcullman:
@greg-a-smith - have you already tried this updated workflow in a lab repo?

.npmrc
bcullman: does this need to be added to one of the `.ignore` files?

greg-a-smith: I don't _think_ so.  We don't want it in `.gitignore` since it's a versioned resource.  The `.eslintignore` shouldn't matter because we're only linting in the `src` and `scripts` directories.  It would seem strange if `.npmignore` ignored `.npmrc`.

greg-a-smith:
@bcullman I did use this workflow in the `github-assistant` repo so I'm fairly confident that this should work.

PR #369, opened 2019-02-05
chore: add playground start script

jbadan:
### Description

Adds a new script `npm run start:playground` that has linting disabled. This can be used to get stop the dev server from stopping because of lint failures while debugging.

`npm start` still functions as expected.


config/webpack.config.dev.js
```diff
@@ -40,50 +39,69 @@ const sassModuleRegex = /\.module\.(scss|sass)$/;
...
+            }
+        }
+    ];
+    if (preProcessor) {
+        loaders.push(require.resolve(preProcessor));
+    }
+    return loaders;
+};
+
+const getLintingRule = () => {
```
jbadan: This file got messy - it needed indents. This is the only real change besides line 199 where `getLintingRule()` is called.

config/webpack.config.dev.js
```diff
@@ -76,6 +75,25 @@ const getStyleLoaders = (cssOptions, preProcessor) => {
     return loaders;
 };
 
+const getLintingRule = () => {
+    return process.env.LINT_ENV !== 'false' ?
+        {
```
bcullman: this is an anti pattern here - a return that depending on bool value, may return nothing.

jbadan:
Closing to reopen in new PR to get around broken CLA agreement.

PR #371, opened 2019-02-06
chore: H1 margin; normalize content padding

bcullman:
### Description
margin for block-level elements should always be applied to the bottom. cleaning up H1 here.

image

also cleaning up margin/padding for content area. gave no space on the right side

image

### DIFF
--- a/src/_playground/App.scss
+++ b/src/_playground/App.scss
@@ -75,8 +75,7 @@ $fd-fonts-path: '../../node_modules/fiori-fundamentals/dist/fonts/';
             padding: 2rem;
 
             .frDocs-Content__header, .frDocs-markdown > h1 {
-                margin: 12px 0;
-                margin-top: 0;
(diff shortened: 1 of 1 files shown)


src/_playground/App.scss
```diff
@@ -71,16 +71,12 @@ $fd-fonts-path: '../../node_modules/fiori-fundamentals/dist/fonts/';
...
-            margin-right: auto;
-            margin-left: 12px;
-            padding-left: 1rem;
-            padding-right: 1rem;
+            padding: 2rem;
 
             .frDocs-Content__header, .frDocs-markdown > h1 {
                 margin: 12px 0;
-                margin-top: 2rem;
+                margin-top: 0;
```
greg-a-smith: Should we just rewrite these lines as `margin: 0 0 12px;`?

bcullman: yup. done.

PR #372, opened 2019-02-06
fix: Refactor SideNavigation to remove React-Router dependency

### LINKED ISSUE #115: Remove react router as a dependency

jeffredodd:
### Description
This PR aims to resolve #115. In order to remove the dependency on React Router from Fundamentals-React (#115), we must refactor the the following components...

- ~~Breadcrumb (#294)~~
- ~~Menu (#293)~~
- ~~Tabs (#336)~~
- SideNavigation (#372)

This PR is focused on **SideNav** and does introduce breaking changes such as...

- Complete Refactor of JSX components for `SideNav`
- Introduction of `SideNavGroup`, `SideNavItem`, and `SideNavSubItems` components
- `<SideNavList>` now takes `<SideNavItem>` instead of accepting an object through the `items` prop.
- `<SideNavItem>` no longer creates its own `<Link>` but instead allows it to be added via an HOC pattern. 
- Demotion of `react-router` and `react-router-dom` to  `devDependencies`.

See the documentation for examples of how to use `<SideNav>`.


src/SideNavigation/SideNavigation.js
```diff
@@ -1,204 +1,286 @@
...
+    onItemSelect: PropTypes.func
+};
+
+SideNavList.propDescriptions = {
+    expandedIds: 'An array of strings for expanded `SideNavItems`\'s.',
+    selectedId: 'The id of the selected `SideNavItem`.',
+    onItemSelect: 'Callback function when user selects an item .'
+};
+
+export const SideNavItem = ({children, expandedIds = [], glyph, id, isSubItem, name, onClick, onItemSelect, selected, selectedId, url, ...props}) => {
```
bcullman: Let's rename to `SideNavListItem`

src/SideNavigation/SideNavigation.js
```diff
@@ -1,204 +1,286 @@
...
 };
 
-export const SideNavGroup = props => {
-    const { title, children, className, titleProps, ...rest } = props;
+SideNav.propDescriptions = {
+    icons: 'Set to **true** enables side navigation collapsed with icons.',
+    selectedId: 'The id of the selected `SideNavItem`.'
+};
 
+export const SideNavGroup = ({children, className, expandedIds, onItemSelect, selectedId, title, titleProps, ...props}) => {
```
bcullman: Lets remove `SideNavGroup`, and add title prop to `SideNavList`

src/SideNavigation/SideNavigation.js
jeffredodd: Done!

src/SideNavigation/SideNavigation.js
jeffredodd: Done!

src/SideNavigation/SideNavigation.js
```diff
@@ -1,204 +1,263 @@
...
-                    })}
-                </ul>
-            </BrowserRouter>
+            <div
+                className={sideNavHeaderlasses}>
+                <h1 {...titleProps} className='fd-side-nav__title'>
+                    {title}
+                </h1>
+                {sideNavList}
+            </div>
```
greg-a-smith: Do we need this `<div>` wrapper?  It seems odd that it's included when a title exists, but not otherwise.  It appears both styles tied to `fd-side-nav__group` already exist on `fd-side-nav__list` and `fd-side-nav__title` does not require a specific parent class.  Maybe just `<React.Fragment>` here instead?

src/SideNavigation/SideNavigation.js
```diff
@@ -1,204 +1,263 @@
...
     }
-}
+
+    return sideNavList;
+};
+
 SideNavList.propTypes = {
-    items: PropTypes.array.isRequired,
-    className: PropTypes.string
+    children: PropTypes.node,
```
greg-a-smith: Same comment as above regarding specific instances of children.

src/SideNavigation/SideNavigation.js
```diff
@@ -1,204 +1,263 @@
...
+                } else {
+                    return child;
+                }
+            })}
+        </li>
     );
+};
 
+SideNavListItem.propTypes = {
+    children: PropTypes.node,
```
greg-a-smith: Same comment as above regarding specific instances of children.

src/SideNavigation/SideNavigation.js
```diff
@@ -1,204 +1,263 @@
...
+        </ul>
     );
 };
 
-SideNavGroup.propTypes = {
-    title: PropTypes.string.isRequired,
-    className: PropTypes.string,
-    titleProps: PropTypes.object
+SideNavSubList.propTypes = {
+    children: PropTypes.node,
```
greg-a-smith: Same comment as above regarding specific instances of children.

src/SideNavigation/SideNavigation.js
```diff
@@ -1,204 +1,263 @@
...
+    selected: 'Set to **true** to display as selected.',
+    selectedId: 'The id of the selected `SideNavListItem`.',
+    url: 'Creates an internal anchor when a child anchor is not provided.'
+};
+
+SideNavListItem.defaultProps = {
+    onClick: () => {}
+};
+
+export const SideNavSubList = ({children, id, onItemSelect, open, selectedId, ...props}) => {
```
greg-a-smith: Do we need this extra component?  Semantically, isn't a sub list just another list?  Maybe have an additional prop like `hasParent` (that could be passed in the `cloneElement` calls) that allows you to add different classes or do other conditional logic.

src/SideNavigation/SideNavigation.js
```diff
@@ -1,204 +1,263 @@
...
-                'is-selected': this.state.selectedItem === id
-            }
+        return (
+            <nav {...rest} className={sideNavClasses}>
+                {React.Children.map(children, (child) => {
+                    if (React.isValidElement(child)) {
+                        return React.cloneElement(child, {
+                            expandedIds: this.state.expandedIds,
+                            onItemSelect: this.handleSelect,
+                            selectedId: this.state.selectedId
```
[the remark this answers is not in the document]
jeffredodd: Chatted with @greg-a-smith about this one. We will have to continue tracking selectedId state however we can track expanded collapsed state of the Sub-`SideNavList`'s at the `SubNavListItem` level.

src/SideNavigation/SideNavigation.js
jeffredodd: In this case I think we should try to match the Fundamental HTML which does include this `<div>`.

src/SideNavigation/SideNavigation.js
jeffredodd: This component has been refactored out.

bcullman: it's still being exported. should it?

src/index.js
```diff
@@ -50,7 +50,7 @@ export { ProductTile } from './Tile/Tile';
 export { ProductTileContent } from './Tile/Tile';
 export { ProductTileMedia } from './Tile/Tile';
 export { SearchInput } from './SearchInput/SearchInput';
-export { SideNav, SideNavList, SideNavGroup } from './SideNavigation/SideNavigation';
+export { SideNav, SideNavList, SideNavListItem, SideNavSubList } from './SideNavigation/SideNavigation';
```
bcullman: should `SideNavSubList` still be exported?

src/SideNavigation/SideNavigation.js
```diff
@@ -1,204 +1,251 @@
...
+    open: PropTypes.bool,
+    selectedId: PropTypes.string,
+    title: PropTypes.string,
+    titleProps: PropTypes.object,
+    onItemSelect: PropTypes.func
 };
 
 SideNavList.propDescriptions = {
-    items: 'An array of objects with keys \'id\', \'url\', \'name\', \'hasChild\', \'child\', and \'glyph\' setting the attributes of the items.'
+    hasParent: 'Automatically set to **true** when item has a parent `SideNavListItem`',
```
bcullman: if this is set automatically, should it be exposed as a prop type?

src/SideNavigation/SideNavigation.js
```diff
@@ -1,204 +1,251 @@
...
+export class SideNavListItem extends React.Component {
+    constructor(props) {
+        super(props);
 
-    const sideNavGroupClasses = classnames(
-        'fd-side-nav__group',
-        className
-    );
+        this.state = {
+            expanded: this.props.expanded ? this.props.expanded : false
```
greg-a-smith: I think we will need a `getDerivedStateFromProps` (previously known as `componentWillReceiveProps`) to keep the internal state in sync since `expanded` is a prop.

src/SideNavigation/SideNavigation.js
```diff
@@ -1,204 +1,251 @@
...
-
-                obj[id] = false;
-                initialState.push(obj);
-            }
-        });
-
         this.state = {
-            selectedItem: 'item_2',
-            itemStates: initialState
+            selectedId: props.selectedId
```
greg-a-smith: I think we will need a `getDerivedStateFromProps` (previously known as `componentWillReceiveProps`) to keep the internal state in sync since `selectedId ` is a prop.

src/SideNavigation/SideNavigation.js
jeffredodd: Corrected in ca228a2

src/SideNavigation/SideNavigation.js
jeffredodd: Corrected in ca228a2

src/SideNavigation/SideNavigation.js
jeffredodd: We have an issue where internal props have to be documented or they are left with a blank description. @greg-a-smith and I chatted about this and decided to write that these props are used for internal use only. I've opened #393 which will aim to correct this issue and would remove these prop descriptions.

src/index.js
jeffredodd: Fixed

src/SideNavigation/SideNavigation.js
jeffredodd: After trying to tackle this, there is not a great way in React to enforce what components are passed as children. For now we are ignoring this issue per a conversation with @greg-a-smith.

src/SideNavigation/SideNavigation.js
jeffredodd: After trying to tackle this, there is not a great way in React to enforce what components are passed as children. For now we are ignoring this issue per a conversation with @greg-a-smith.

src/SideNavigation/SideNavigation.js
```diff
@@ -1,209 +1,276 @@
...
+            return (
+                <a
+                    className={getClasses()}
+                    href={url}
+                    onClick={(e) => {
+                        onClick(e);
+                        onItemSelect(e, id, hasChild);
+                        if (hasChild) {
+                            this.handleExpand();
+                        }
```
greg-a-smith: I found what I _think_ is a bug, but wanted to get thoughts on it first before creating an issue.  It would be a follow-up task if it was a bug since the behavior existed previously.  When click on a nav item that is the expand/collapse item, it changes the `selectedId`, but these items themselves shouldn't actually be "selectable".  They are simply used to expand and collapse the children within it.

jeffredodd: I agree this was an issue. Pushed 33e7b51 to fix this issue and 0935d11 to fix the corresponding test.

jeffredodd:
I can toss on a 🚢 for @greg-a-smith's contributions! Just have to wait for Coveralls 😄

PR #375, opened 2019-02-06
chore: Release process cleanup

greg-a-smith:
### Description
In the last PR to refine the release process, I moved the NPM credentials into a `.npmrc` file.  The problem is that those environment variables need to be present (even in a local development environment) in order to run any npm scripts.  This PR removes that file and restores a script file that will setup the `.npmrc` only during CI server deployments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,6 @@
     "fork-ts-checker-webpack-plugin-alt": "0.4.14",
     "fs-extra": "7.0.0",
     "gh-pages": "^2.0.1",
-    "github-api": "^3.0.0",
     "github-assistant": "^0.2.0",
     "group-array": "^0.3.3",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -122,8 +121,7 @@
     "webpack": "4.19.1",
     "webpack-dev-server": "^3.1.11",
     "webpack-manifest-plugin": "2.0.4",
-    "workbox-webpack-plugin": "3.6.3",
-    "yargs": "^12.0.5"
```
greg-a-smith: The `github-api` and `yargs` are no longer necessary since that functionality was offloaded to `github-assistant`.

chrismanciero:
👍🚢

bcullman:
🚢

Issue #376, opened 2019-02-07
Remove propDescriptions from prod build


mobot11:
### Description
Currently our custom propDescriptions we use to build our docs make it into the production build. We should leverage https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types to remove them in the production build.

### Versions
**fundamental-react:**

---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

PR #377, opened 2019-02-07
chore: add babel plugin to remove prop descriptions from prod build

mobot11:
### Description
Leverages code from https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types/tree/master/src to remove prop descriptions from production build.

fixes https://github.com/SAP/fundamental-react/issues/376


greg-a-smith:
I like the premise of this.  I'm just a little unsure about the volume of code we are ingesting to make it happen (and code we basically didn't write).  So, I guess I'm neither in favor of it nor against it.  I'd be interested in hearing thoughts from others.

jbadan:
Same as @greg-a-smith, just dropping a note that we need to coordinate as this pr will not play nicely with https://github.com/SAP/fundamental-react/pull/378

mobot11:
We get about a 5KB size decrease by using this. While not a whole lot, it could end up being significant as the library grows in size.

mobot11:
Closing

PR #378, opened 2019-02-07
chore: adding module to package.json

jbadan:
### Description

Setting modules to false to ensure that import statements are left as is (opposed to transpiling them to require) to create new `lib/es` folder . This should give us a tree shakeable icon on bundlephobia.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "private": false,
   "license": "Apache-2.0",
   "main": "lib/index.js",
+  "module": "lib/es/index.js",
   "homepage": "https://sap.github.io/fundamental-react",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -15,7 +16,9 @@
     "build-js": "node scripts/build.js",
     "build:lint:fix": "eslint --fix 'src/**' --ext .js,.jsx --env browser,node",
     "build:lint": "eslint 'src/**' --ext .js,.jsx --env browser,node",
-    "build": "rm -rf lib && NODE_ENV=production babel src --out-dir lib --ignore \"src/**/*.spec.js\",\"src/**/*.test.js\",\"src/**/*.Component.js\",\"src/_playground/*\"",
+    "build": "npm run build:lib && npm run build:es",
```
bcullman: we need to build both?

mobot11:
My question is do we want to provide an es modules version of the library when IE 11 support is a first-class requirement of fundamental-react? Even though Material provides both CJS and ESM distributions they "strongly discourage library authors from using" the ESM distribution https://material-ui.com/guides/minimizing-bundle-size/#ecmascript.

bcullman:
Their documentation starts a bit above that anchor. 

Try https://material-ui.com/guides/minimizing-bundle-size/#how-to-reduce-the-bundle-size

greg-a-smith:
Random thoughts time. 😄 Take these with a grain of salt since I really haven't spent much time at all on tree shaking stories and research.

It seems strange to transpile the library twice.  If our goal is to support tree shaking and we want import statements to remain simple (_e.g._ `import { Alert } from 'fundamental-react';`), then shouldn't we just do the es2015 style and make the consumer do the commonJS transpiling?

Or, do we just find a way to document the specific import statements better (_e.g._ `import { Alert } from 'fundamental-react/lib/Alert/Alert`) which, if I understand it correctly, already just provides the specific components they are using and thus makes tree shaking a mute point.

I'm open to any and all feedback/thoughts and please correct me if my understandings are invalid.

package.json
jbadan: Yes

jbadan:
@greg-a-smith I don't have a great answer for this. I was following the lead of all the other bigger libraries like react-bootstrap, material-ui, grommet, etc. Having both does provide more flexibility for consumers.

@mobot11 That was the point of putting it inside of the` lib/es` folder. Nobody would use it by accident. I'm not 100% what changes you're asking for, I understand some documentation but what exactly are you looking for? Another section on the readme? An import example?

mobot11:
@jbadan A section in the readme detailing that we provide an ES Modules distribution for evergreen browsers would suffice. If we are going to provide the ESM build we should document it somewhere.

@greg-a-smith I think providing both makes the most sense. If IE 11 support is a requirement of this library, we shouldn't make consumers take extra steps to make our library IE 11 compatible.

jbadan:
Will reopen at a future date once we have a more clear direction with this.

PR #380, opened 2019-02-07
chore: add eslint plugin for react hooks

### LINKED ISSUE #379: eslint support for react hooks
### Description
With React 16.8, React Hooks are available in a stable release. An [eslint plugin](https://www.npmjs.com/package/eslint-plugin-react-hooks) is available to validate the use of hooks within a component.

---

_**DISCLAIMER:**

chrismanciero:
### Description
Adds eslint checking for use of react hooks within components

Reference sites:

- [https://reactjs.org/blog/2019/02/06/react-v16.8.0.html](https://reactjs.org/blog/2019/02/06/react-v16.8.0.html)
- [https://reactjs.org/docs/hooks-rules.html](https://reactjs.org/docs/hooks-rules.html)
- [https://www.npmjs.com/package/eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks)

fixes #379

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,6 +17,7 @@ plugins:
     - jsx-a11y
     - loosely-restrict-imports
     - react
+    - react-hooks
 
 rules:
     brace-style: 2
@@ -134,6 +135,7 @@ rules:
     react/react-in-jsx-scope: 2
     react/self-closing-comp: 2
     react/sort-prop-types: [2, { 'callbacksLast' : true, 'ignoreCase' : true, 'requiredFirst' : true, } ]
+    react-hooks/rules-of-hooks: 2
     semi-spacing: 2
     semi: 2
     sort-imports: [2, { 'ignoreCase': true, 'ignoreMemberSort': false, 'memberSyntaxSortOrder': ['none', 'single', 'multiple', 'all'] }]
--- a/package-lock.json
+++ b/package-lock.json
@@ -1828,7 +1828,7 @@
         },
         "chalk": {
           "version": "1.1.3",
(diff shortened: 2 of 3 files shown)


.eslintrc
```diff
@@ -17,6 +17,7 @@ plugins:
     - jsx-a11y
     - loosely-restrict-imports
     - react
+    - react-hooks
```
greg-a-smith: Not that this is a bad idea, but are we using React Hooks?

bcullman:
i'd like this in our back-pocket, but I suggest closing until hooks lands in fundamental-react. that will be when a) code is written, and b) this

https://github.com/SAP/fundamental-react/blob/master/package.json#L127-L128

is bumped to `^16.8.0`

.eslintrc
chrismanciero: I added this more as a future check. Hooks was just released with react v16.8. A new developer can check out our library and make/change a component using hooks.
Obviously we have to approve it :)

chrismanciero:
Doesn't `"react": "^16.6.3",` in our package.json file mean that if someone was to do an npm install of our library it will install react version > 16.6.3 (potentially 16.8) or does the npm install limit it to just 16.6.* versions?

greg-a-smith:
@chrismanciero The hat (`^`) means the consumer is willing to take any additional minor or patch-level version bumps, but not major versions (breaking changes).  That said, the `package-lock.json` file may prevent this since its job is to "lock" the version so the environment is repeatable.  You can remove that file, run `npm install` and have it be regenerated at which time, it would bump to the newest non-breaking version.  Hope that helps.

chrismanciero:
Yes it does thanks. closing the pr.

PR #381, opened 2019-02-07
chore: remove unused test script

bcullman:
### Description
not seeing an advantage of using the test.js script that was ejected by created react app. removing.

`npm run test` takes the place of `npm run test:dev`


package.json
```diff
@@ -33,8 +33,7 @@
     "std-version": "standard-version -m \"chore(release): version %s build ${TRAVIS_BUILD_NUMBER} [ci skip]\"",
     "test:coverage": "jest --coverage",
     "test:coveralls": "jest --coverage && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
-    "test:dev": "jest",
-    "test": "node scripts/test.js",
```
greg-a-smith: The `test.js` script provided a nice "watch" mode for unit tests.  I'm not quite sure if it was the script or if it's something in `jest` or a combination of both.  I'm okay removing the script if we could restore that watch functionality.

bcullman:
all right sounds like there is some usefulness here. I'll close my PR.

PR #382, opened 2019-02-08
Fix/popover remove style

chrismanciero:
### Description
Popover component was applying a CSS class (example: `fd-popover--right`) that does not exist in the Fiori Fundamental CSS Style guide. This PR removes that applying of the class to the Popover component wrapper div.

### DIFF
--- a/src/Popover/Popover.js
+++ b/src/Popover/Popover.js
@@ -1,5 +1,5 @@
 import classnames from 'classnames';
-import {POPOVER_ALIGNMENTS} from '../utils/constants';
+import { POPOVER_ALIGNMENTS } from '../utils/constants';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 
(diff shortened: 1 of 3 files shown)


src/Popover/Popover.js
```diff
@@ -1,5 +1,5 @@
 import classnames from 'classnames';
-import {POPOVER_ALIGNMENTS} from '../utils/constants';
+import { POPOVER_ALIGNMENTS } from '../utils/constants';
```
bcullman: unnecessary whitespace change. If you prefer this style, please open a separate PR, with a lint rule enforcing this style

```
object-curly-spacing: [2, always]
```

chrismanciero: The shift again happened from some VS Code setting on a TypeScript extension that I removed. Will look into why it's being applied or will add the lint rule.

PR #386, opened 2019-02-08
chore: Add eslint rule object-curly-spacing

chrismanciero:
### Description
Add and applies the [object-curly-spacing](https://eslint.org/docs/rules/object-curly-spacing) eslint rule


chrismanciero:
And here i am thinking no one would want this because it wasnt already there 😄

PR #388, opened 2019-02-08
fix: add displayName for each component

### LINKED ISSUE #339: Components need static display names
### Description

Each component needs a set display name for easier debugging when minified. When display names are not set, the component "names" become "t", "o", "c", etc.  See: https://github.com/facebook/react/issues/4915

ex: 
`static displayName = 'Button';`

alternatively, we could expect consumers to use https://www.npmjs.com/package/babel-plugin-add-react-static-displayname if they want this feature.

### Versions
**fiori-fundamentals:**

---

_**DISCLAIMER:**

chrismanciero:
### Description
Add displayName value to all component per discussion on https://github.com/facebook/react/issues/4915

fixes #339

### DIFF
--- a/src/ActionBar/ActionBar.js
+++ b/src/ActionBar/ActionBar.js
@@ -20,6 +20,7 @@ export const ActionBar = ({ mobile, width, children, className, ...props }) => {
         </React.Fragment>
     );
 };
+ActionBar.displayName = 'ActionBar';
 
 ActionBar.propTypes = {
     className: PropTypes.string,
--- a/src/Alert/Alert.js
+++ b/src/Alert/Alert.js
@@ -70,6 +70,7 @@ export class Alert extends Component {
         );
     }
 }
+Alert.displayName = 'Alert';
 
 Alert.propTypes = {
     buttonProps: PropTypes.object,
--- a/src/Badge/Badge.js
+++ b/src/Badge/Badge.js
@@ -20,6 +20,7 @@ export const Badge = ({ type, modifier, children, className, ...props }) => {
         </span>
     );
 };
+Badge.displayName = 'Badge';
 
 Badge.propTypes = {
     className: PropTypes.string,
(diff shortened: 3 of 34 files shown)


mobot11:
@jbadan @chrismanciero does it make sense for us to do this manually or should we do this programmatically during build time via https://babeljs.io/docs/en/babel-plugin-transform-react-display-name. I like the idea of automating this but curious to hear your thoughts.

chrismanciero:
I’m all for automation. I’m not too familiar with Babel plugin, but if they have one that does it, lets use it.

jbadan:
@mobot11 The reason we couldn't use that plugin in nui-widgets was the way our playground was structured. I think as long as our docs/examples still work, it should be fine: For Dylan: https://github.concur.com/coreui/nui-widgets/pull/471 @chrismanciero sorry you can't see that link, but it's the pr with a discussion about the babel plugin and why it didn't work for us in nui-widgets.

chrismanciero:
@mobot11 - I just created a new branch to try this change out and after installing @babel/plugin-transform-react-display-name and making the change in babel.config.js when I do a `npm start` I get the following error

screen shot 2019-02-08 at 4 24 23 pm

jbadan:
@chrismanciero @mobot11 I get the same error unless I add it to default plugins here: https://github.com/SAP/fundamental-react/blob/master/babel.config.js#L16

Does that still accomplish what we want?

chrismanciero:
@jbadan - if you update the babel.config.js to be like this:
const productionPlugins = [
    [
        'transform-react-remove-prop-types',
        {
            mode: 'unsafe-wrap'
        }
    ],
    [
        '@babel/plugin-transform-react-inline-elements'
    ],
    [
        '@babel/plugin-transform-react-display-name'
    ],
    path.resolve(__dirname, './devtools/remove-prop-descriptions/index.js')
];

The application loads, but after I do a npm run build and search a component in the /lib folder I don't see displayName.

mobot11:
I can't seem to get this plugin to do what we want either. I'm fine doing it manually.

jbadan:
@mobot11 @chrismanciero I think that plugin meant for the consumer end, not the library end. That might be the problem.

PR #394, opened 2019-02-11
fix: CORUI-6127: Refactor Tree component

### LINKED ISSUE #265: Consider refactoring Tree component
Prop spreading was added to `Tree` for the header and list portions.  However, the current API does not easily support prop spreading to the additional elements.

cjskillingstad:
### Description
Complete refactor of the Tree component.

Going from passing a data object to the Tree component to allowing the consumer to iterate through their own data object and render the Tree as JSX elements.

The expand/collapse states are all held in the outer most `Tree` component. Each `TreeRow` is given a unique id from the `shortid` package, with a boolean value representing expanded/collapsed.

Updated unit tests, snapshots, and documentation.

fixes #265


package.json
```diff
@@ -43,7 +43,8 @@
     "fiori-fundamentals": "^1.4.0",
     "react-router": "^4.3.1",
     "react-router-dom": "^4.3.1",
-    "react-syntax-highlighter": "^9.0.1"
+    "react-syntax-highlighter": "^9.0.1",
+    "shortid": "^2.2.14"
```
bcullman: shortid can generate invalid HTML id. Use a util function to build HTML valid shortid values

see: https://github.concur.com/coreui/nui-widgets/pull/432/files

src/Tree/Tree.js
```diff
@@ -1,238 +1,330 @@
-import { Dropdown } from '../Dropdown/Dropdown';
+import classnames from 'classnames';
 import PropTypes from 'prop-types';
+import shortid from 'shortid';
 import React, { Component } from 'react';
 
-export class Tree extends Component {
+export class TreeCell extends Component {
```
bcullman: I believe that Treeview and SideNav have a lot in common.

Let's try and standardize on naming conventions for subcomponents, where it makes sense.

package.json
cjskillingstad: Fixed in c25e564

src/Tree/Tree.js
cjskillingstad: I agree. I think the `Tree` component has one extra layer though for the nested elements, as well as the header elements. What do you think of:
```
<Tree>
    <TreeHeader>
        <TreeListItem>
     <TreeList>
         <TreeRow>
            <TreeListItem>
```
Where `SideNav` is:
```
<SideNav>
    <SideNavList>
        <SideNavListItem>
```

src/Tree/Tree.js
```diff
@@ -1,238 +1,330 @@
...
+            buttonProps,
+            children,
+            isExpanded,
+            onExpandAll,
+            ...rest
+        } = this.props;
+
+        return (
+            <div {...rest} className='fd-tree fd-tree--header'>
+                <div className='fd-tree__row fd-tree__row--header'>
```
greg-a-smith: I'm assuming this isn't its own component (or a reuse of `TreeRow`) because there is only one row for headers, right?  That said, if `TreeRow` is necessary, then shouldn't we be able to reuse it here?
```jsx
<Tree>
    <TreeHeader>
        <TreeRow>
            <TreeCell />
            <TreeCell />
            <TreeCell />
        </TreeRow>
    </TreeHeader>
</Tree>
```
Also, same question about not allowing custom CSS classes on this element.  And prop spreading?

src/Tree/Tree.js
```diff
@@ -1,238 +1,330 @@
...
+                            const className = classnames({
+                                'fd-tree__col--control': isFirstTreeCell
+                            });
+
+                            // Add expand button to first TreeCell element
+                            const newChildren = isFirstTreeCell ? (
+                                <div>
+                                    <button
+                                        {...buttonProps}
+                                        aria-label='expand'
```
greg-a-smith: This should be extracted into a localized prop although that's probably a whole other story in itself since none of the components support that right now.  Also, the label should probably also toggle between "expand" and "collapse".

src/Tree/Tree.js
```diff
@@ -1,238 +1,330 @@
...
+                                        aria-pressed={isExpanded}
+                                        className='fd-tree__control'
+                                        onClick={onExpandAll} />
+                                    {child.props && child.props.children}
+                                </div>
+                            ) : child.props && child.props.children;
+
+                            return React.cloneElement(child, {
+                                children: newChildren,
```
greg-a-smith: This seems dangerous.  If we're providing semantic components for them to use, we probably shouldn't be messing around with `children`.

src/Tree/Tree.js
```diff
@@ -1,238 +1,330 @@
...
-            expandAllClicked: false,
-            numberOfElements: 0
-        };
+
+        const {onExpandClick} = this.props;
+
+        // Generate unique id for row to manage expand/collapse state in parent
+        const id = shortid.generate();
+
+        this.state = {id};
```
greg-a-smith: Storing an `id` in state seems wrong.  I understand not including it in the `render` function so it's not just regenerated each time, but how about:
```js
this.rowId = shortid.generate();
```
in the constructor?

src/Tree/Tree.js
```diff
@@ -1,238 +1,330 @@
...
+                    {child.props && child.props.children}
+                </div>
+            ) : child.props && child.props.children;
 
-        this.setState({
-            iStates: modifiedStates,
-            expandAllClicked: !this.state.expandAllClicked,
-            numberOfElements: numberOfElements
+            return isTreeCell ?
+                React.cloneElement(child, {className, children: newChildren}) :
```
greg-a-smith: Similar to comment above, messing with `children` _seems_ wrong.

src/Tree/Tree.js
```diff
@@ -1,238 +1,330 @@
...
+    level: 0
+};
+
+export class Tree extends Component {
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            expandData: {},
+            isExpandAll: false
```
greg-a-smith: I would caution against tracking too much in state (if any) in this component.  It is very easy to either get out of sync it prevents the consumer from making any alterations with props.  Because this is a component library, in large part, these components should be "dumb".  Their job is simply to render what the consumer tells them to and have some callbacks exposed that tell the consumer if anything changed.  They should look for the consumer to maintain any sort of "state" because they likely have a redux store to leverage.

src/Tree/Tree.js
cjskillingstad: Correct, it doesn't look like you should ever have more than one `TreeHeader`

src/utils/shortId.js
```diff
@@ -0,0 +1,8 @@
+import shortId from 'shortid';
+
+// A passthrough method to prepend cnqr- to every automatically generated ID.
+function generate() {
+    return 'cnqr-' + shortId.generate();
```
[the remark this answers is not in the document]
cjskillingstad: Changed to `fd-` in 9ce9d6a

src/Tree/Tree.js
```diff
@@ -1,238 +1,330 @@
...
+                {children}
+            </div>
+        );
+    }
+}
+
+TreeCell.propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string
+};
```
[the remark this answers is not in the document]
cjskillingstad: Just about all of these are internal props, so I added the _INTERNAL USE ONLY._ descriptions in 46de12d

src/Tree/Tree.js
```diff
@@ -1,238 +1,330 @@
...
+        const {
+            buttonProps,
+            children,
+            isExpanded,
+            onExpandAll,
+            ...rest
+        } = this.props;
+
+        return (
+            <div {...rest} className='fd-tree fd-tree--header'>
```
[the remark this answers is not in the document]
cjskillingstad: Fixed in 7d6f561

src/Tree/Tree.js
cjskillingstad: I will leave it hard coded for now so we can tackle those localized labels at once. Toggling the labels to "expand"/"collapse" in a2bd47e

src/Tree/Tree.js
cjskillingstad: I don't like it either. But with the first cell in each row, there needs to be the expand/collapse button within the cell div, and I'm not sure how to achieve that without setting children. I tried pulling the button outside of the cell div as a sibling element and the styles didn't look right.

src/Tree/Tree.js
cjskillingstad: Fixed in fc864e3

src/Tree/Tree.js
cjskillingstad: I agree, but I feel like the expand/collapse state is something that the consumer shouldn't have to manage through application state. It does feel like something that the component should handle and the application shouldn't have to worry about.

src/Tree/Tree.js
cjskillingstad: The html is different for a row in the header vs. a row in the list part of the tree. For rows in the List part of the tree, they are rendered as `<li>` tags with `<div>`'s inside of them, but in the header they are just `<div>`'s. I could pass a prop to the `TreeRow` to determine if it's a header row, then add logic to determine what html elements to render.

src/Tree/Tree.js
greg-a-smith: I don't disagree, however a consumer _can't_ control it now even if they wanted to.  If we're going to track this type of state, then we need to expose a prop and add a `getDerivedStateFromProps` (formerly `componentWillReceiveProps`) function to sync state with incoming props.  This way, a consumer _could_ provide an initial or updated state.  This would probably also mean exposing a callback prop for when the expansion state changes.

I was thinking that maybe it would be easier to just have `TreeHeader` and `TreeList` maintain their own expanded states, but the expand/collapse-all may not play nice with that so we can leave it where it is.

I still think this type of state shouldn't be in this component, but the `Tree` component is still under review and I've heard there may actually be multiple tree-type components coming so there will be time to revisit this again.

This is certainly _MUCH_ better than it was before! 👍

src/Tree/Tree.js
bcullman: Can you have a header without a row? Can a Header have multiple tree rows? if not, TreeHeader 
 could just be a proptype `headerText` on TreeRow.

greg-a-smith:
@SAP/fundamental-react-team Now that I've had time to digest these changes and look over the Fiori Fundamentals and WAI specs a bit, I will throw out some thoughts on the component structure and component names.

Let me preface this by saying the example shown is a very complicated tree view.  Typically, tree views do not have headers (in fact, they are not actually part of the [Tree View](https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView) spec in the WAI) nor are they used to display tabular data.  That said, I think we can come up with a structure that will support simple tree views and more complicated ones (like the example).

Using the HTML/CSS spec from [Fiori Fundamentals](https://sap.github.io/fundamental/components/tree.html) as inspiration, how about this for a component structure?

```jsx
<TreeView>
    <TreeHeader>
        <TreeCol />
    </TreeHeader>
    <Tree>
        <TreeItem>
            <TreeRow>
                <TreeCol />
            </TreeRow>
            <Tree>
                <TreeItem>
                    <TreeRow>
                        <TreeCol />
                    </TreeRow>
                </TreeItem>
            </Tree>
        </TreeItem>
    </Tree>
</TreeView>
```

It's fairly verbose, but it does match the structure of the spec.

The `TreeView` component would not render anything, but would simply be used for wrapping everything together and handling state tracking with all the children -- especially for the expand/collapse-all button in the header (which is also not part of the WAI spec).  I'm still not a huge fan of tracking all this state, but if we expose props for setting the expanded state of nodes and callbacks for changes, then it could work.

The `TreeHeader` component is optional and typically only used when displaying tabular data.

Since the `Tree` component is no longer the top-level component, it can easily be reused as both a root node (with role of "tree") and as a branch node (with a role of "group").

The `TreeItem` component provides a semantic wrapper around a row and another branch node.  This would only have either 1 or 2 children.  It would always have a `TreeRow` child and optionally a `Tree` child.  I think this provides a little cleaner structure than having 'n' number of `TreeCell` children.

There can be any number of `TreeCol` component children for either `TreeHeader` or `TreeRow` components.

---

On a separate, but somewhat related note, I think this story should also include the addition of a new "simple" tree view example.  Something with just one column of data for each row.  A directory structure with folders and files comes to mind.

src/Tree/Tree.js
greg-a-smith: Also, if we expose props to let consumers manage expand states, then we likely need to expose `id` props for the items.  The component could use `shortid` to generate ones if none are provided, but consumers will need to provide these if they want to work with them in their own state management.

src/Tree/Tree.js
cjskillingstad: The header html element always contains a row html element. I don't see a situation where a consumer would want multiple rows within a header.

We could accept a `headerText` prop as an array of strings for the `TreeRow` component and render a `TreeCell` for each value. The reason I left it to the consumer to render the `TreeCell`'s is for consistency between the Header and the List

jeffredodd:
I like @greg-a-smith suggestions! I would perhaps call `Tree` a `TreeBranch` instead. Mostly because the idea of nesting trees seems like a weird concept, but branches coming from another tree branch makes more sense. Just a minor suggestion. 😄

bcullman:
how about `TreeHead` (vs `TreeHeader`) and `TreeBody`?

src/TreeView/TreeView.js
```diff
@@ -0,0 +1,353 @@
...
+            [`fd-tree__group--sublevel-${level}`]: level > 0,
+            'is-hidden': !isExpanded
+        });
+
+        return (
+            <ul
+                {...rest}
+                aria-hidden={!isExpanded}
+                className={className}
+                role={level === 0 ? 'tree' : 'group'}>
```
bcullman: i see this role for the tree. is there some sort of role relationship for the `<treehead>` with the tree? @jeffredodd?

cjskillingstad:
@bcullman Renamed TreeHeader to TreeHead in 152e772
@greg-a-smith Refactored JSX API in 3668242
@greg-a-smith Added simple tree view in docs 2aaf596

cjskillingstad:
@jeffredodd Renamed child `Tree` components to `TreeBranch` in 5f49033

src/Tree/Tree.js
cjskillingstad: Allowed the consumer to override the expand/collapse state of both the `isExpandAll` variable and `isExpanded` for `TreeItem`s in 62ccf9c.
`isExpandAll` Ex: 
```
<TreeView isExpandAll>
    <TreeHead>
        <TreeCol>Header 1</TreeCol>
    </TreeHead>
    <Tree>
        .....
    </Tree>
</TreeView>
```

`isExpanded` Ex:
```
<TreeView isExpandAll>
    <TreeHead>
        ...
    </TreeHead>
    <Tree>
        <TreeItem isExpanded>
            <TreeRow>
                <TreeCol />
            </TreeRow>
            <TreeBranch>
                <TreeItem>
                    <TreeRow>
                         <TreeCol />
                    </TreeRow>
                </TreeItem>
            </TreeBranch>
        </TreeItem>
    </Tree>
</TreeView>
```

src/TreeView/TreeView.js
greg-a-smith: After looking at the WAI again, there is no concept of a header (`TreeHead`) for a `TreeView`.  However, there is a Tree Grid spec that defines a role of `treegrid` on a `<table>` element.  It appears to require standard table markup and uses some additional roles and aria attributes as well.  It really seems like this should be a separate component, but would need to be spec'd at the Fiori Fundamentals first.

See https://www.w3.org/TR/wai-aria-practices-1.1/examples/treegrid/treegrid-1.html

src/TreeView/TreeView.js
```diff
@@ -0,0 +1,471 @@
...
+            'fd-tree--header',
+            className
+        );
+
+        return (
+            <div {...rest} className={headerClassName}>
+                <div className='fd-tree__row fd-tree__row--header'>
+                    {
+                        React.Children.map(children, (child, index) => {
+                            const isFirstTreeCol = index === 0 && child.type && child.type.name === 'TreeCol';
```
greg-a-smith: Should these types of checks now be using `child.type.displayName` instead of `child.type.name`?  If so, these exist in many of these components.

src/TreeView/TreeView.js
```diff
@@ -0,0 +1,471 @@
...
+
+TreeHead.propTypes = {
+    buttonProps: PropTypes.object,
+    children: PropTypes.node,
+    className: PropTypes.string,
+    isExpanded: PropTypes.bool,
+    onExpandAll: PropTypes.func
+};
+
+TreeHead.propDescriptions = {
```
greg-a-smith: I know that `children` is one of the defaulted prop descriptions, but the message is pretty generic.  Since these components now essentially depend on certain components being children of other components, maybe we should add a `children` prop description for each component specifying which child component(s) are expected (_e.g._ `TreeView` expects `TreeHead` and/or `Tree`, `TreeHead` expects `TreeCol`, etc.)

src/TreeView/TreeView.js
```diff
@@ -0,0 +1,471 @@
...
+TreeItem.propDescriptions = {
+    expandData: '_INTERNAL USE ONLY._',
+    isExpanded: 'Set to *true* for expanded tree item. This variable is handled internally, but can be overridden by the consumer through this prop',
+    level: '_INTERNAL USE ONLY._',
+    onExpandClick: '_INTERNAL USE ONLY._'
+};
+
+export class TreeBranch extends Component {
+    render() {
+        return <Tree {...this.props} />;
```
greg-a-smith: I'm wondering if `Tree` should call `TreeBranch` instead and then move all the conditional class logic and children map to `TreeBranch`.  Or, maybe an internal-only component is created (_e.g._ `_Tree`) that both `Tree` and `TreeBranch` call passing the necessary props?

The reason I say this is because `Tree` is level 0 (by definition) and won't ever be anything other than that.  In fact, `level` shouldn't even be a prop on `Tree` and it should probably be a required prop on `TreeBranch` and _not_ have a default value.

src/TreeView/TreeView.js
```diff
@@ -0,0 +1,471 @@
...
+
+        this.state = {
+            expandData: {},
+            isExpandAll: false
+        };
+    }
+
+    static getDerivedStateFromProps(props, state) {
+        const {
+            isExpandAll
```
greg-a-smith: It seems like this should also accept `expandData` as a prop so the user can control which items may be initially expanded.  That would also mean that prop would be checked in `getDerivedStateFromProps `.  If that gets added, then I think there are a few more things that would need to be added:

* Add a callback prop to `TreeView` that gets called whenever the expand/collapse state changes on any row id.
* Add a `rowId` prop to `TreeItem` so a consumer could set their own id value (since they are the keys in the `expandData` object).

src/TreeView/TreeView.js
cjskillingstad: I've seen examples of both. What do you think?

greg-a-smith: I've seen both as well.  We have gone through the work of adding display names for the components so it _seems_ like that's the preferred route now.  I'm open to thoughts from others though.

src/TreeView/TreeView.js
cjskillingstad: Added more specific prop descriptions for `children` in 268f179

src/TreeView/TreeView.js
cjskillingstad: Refactored to use an internal `_Tree` component in 74cff8f. 
I couldn't mark the `level` prop as required because im passing that prop through the parent component by iterating through `React.Children` and the required props are validated at creation time of the component so I was getting warnings. I read about it here: https://github.com/facebook/react/issues/4494

greg-a-smith: Yeah, that makes sense.

src/TreeView/TreeView.js
cjskillingstad: Using `displayName` in 2f74f50

src/TreeView/TreeView.js
cjskillingstad: Exposed `expandData` props and callbacks in 98d4612.

Ex API:
```
<TreeView expandData={{myRowId: true}}>
    <Tree>
         <TreeItem rowId={'myRowId'}>
                 <TreeRow>
                      <TreeCol>Row 1</TreeCol>
                   </TreeRow>
           </TreeItem>
    </Tree>
</TreeView>
```

greg-a-smith:
@cjskillingstad Per @jbadan's instructions, I took the liberty of merging master into this branch so I could test it out properly.

I added just a few minor cleanup items:

1. Fixed the calls to `Import` and `Properties` in `TreeView.Component.js`.
1. Added spaces between words (to match the established pattern) for "TimePicker" and "TreeView" in the left nav.
1. Cleaned up a couple minor prop descriptions.

Issue #314, opened 2019-01-29
Add examples of importing components from the lib/ directory to docs


mobot11:
### Description
After doing some research on tree shaking, I think the best way forward is to provide examples of how to import individual components from the `dist` folder if consumers want to reduce their library size by importing only the component they are using. [Material](https://material-ui.com/api/avatar/) and [React Bootstrap](https://react-bootstrap.github.io/getting-started/introduction/#es6) are good examples.

ex: 
```javascript
import {Alert} from 'react-fundamentals/lib/Alert/Alert'
```

### Versions
**fundamental-react:**

---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

PR #387, opened 2019-02-08
chore: build index files programmatically 

mobot11:
### Description
This PR will programmatically build index files in each directory so components can be imported from the top level
```javascript
import { Alert } from 'fundamental-react/lib/Alert/Alert'
```
becomes
```javascript
import { Alert } from 'fundamental-react/lib/Alert'
```

fixes https://github.com/SAP/fundamental-react/issues/331

PR #395, opened 2019-02-11
chore: add import path from lib to docs

mobot11:
### Description
Adds import path to docs that points users to cherry-pick components from `lib`

fixes https://github.com/SAP/fundamental-react/issues/314

### DIFF
--- a/src/_playground/documentation/Import/Import.js
+++ b/src/_playground/documentation/Import/Import.js
@@ -1,24 +1,24 @@
-import getSourceModules from '../utils/getSourceModules';
+import getSourceModule from '../utils/getSourceModule';
 import { googlecode } from 'react-syntax-highlighter/styles/hljs';
 import packageJson from '../../../../package.json';
 import PropTypes from 'prop-types';
 import React from 'react';
 import SyntaxHighlighter from 'react-syntax-highlighter';
 
-export const Import = ({ sourceModule }) => {
+export const Import = ({ sourceModulePath }) => {
     // remove everything up until the last forward slash and the file extension.
-    const importPath = sourceModule.replace(/^(.*[\\\/])/, '').replace('.js', '');
+    const importPath = sourceModulePath.replace(/^(.*[\\\/])/, '').replace('.js', '');
 
     return (
         <SyntaxHighlighter
             customStyle={{ padding: 0 }}
             language='javascript'
             style={googlecode}>
-            {`import { ${Object.keys(getSourceModules(sourceModule)).sort().join(', ')} } from '${packageJson.name}/lib/${importPath}';`}
+            {`import { ${Object.keys(getSourceModule(sourceModulePath)).sort().join(', ')} } from '${packageJson.name}/lib/${importPath}';`}
         </SyntaxHighlighter>
     );
 };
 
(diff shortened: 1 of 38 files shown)


bcullman:
should there be an accompanied change in readme.md for this?

mobot11:
Readme was updated as part of https://github.com/SAP/fundamental-react/pull/387

src/_playground/documentation/Import/Import.js
```diff
@@ -1,21 +1,24 @@
...
             style={googlecode}>
-            {`import { ${Object.keys(sourceModule).sort().join(', ')} } from '${packageJson.name}';`}
+            {`import { ${Object.keys(getSourceModules(sourceModule)).sort().join(', ')} } from '${packageJson.name}/lib/${importPath}';`}
         </SyntaxHighlighter>
     );
 };
 
 Import.propTypes = {
-    sourceModule: PropTypes.object.isRequired
+    sourceModule: PropTypes.string.isRequired
```
greg-a-smith: Since this is being passed in as a string now, the prop name should probably change as well.  Similar to the new `getSourceModules` util function, maybe use `sourceModulePath`?

src/_playground/documentation/Properties/Properties.js
```diff
@@ -28,7 +31,7 @@ export const Properties = ({ sourceModule }) => {
 };
 
 Properties.propTypes = {
-    sourceModule: PropTypes.object.isRequired
+    sourceModule: PropTypes.string.isRequired
```
greg-a-smith: Since this is being passed in as a string now, the prop name should probably change as well.  Similar to the new `getSourceModules` util function, maybe use `sourceModulePath`?

src/_playground/documentation/utils/getSourceModules.js
```diff
@@ -0,0 +1,9 @@
+const getSourceModules = (sourceModulePath) => {
```
greg-a-smith: Minor nit, but this should be `getSourceModule` (singular).  The singular module may have multiple exported components and/or functions, but it's still just one module.

src/_playground/documentation/utils/getSourceModules.js
```diff
@@ -0,0 +1,9 @@
...
+    const cleanSource = sourceModulePath.replace('./src', '');
+
+    const sourceModules = require('../../../../src' + cleanSource);
+
+    return sourceModules;
+};
+
+export default getSourceModules;
```
greg-a-smith: Other plural references should probably be renamed too (in this file and in `Import` and `Properties`).

src/_playground/documentation/utils/getSourceModules.js
```diff
@@ -0,0 +1,9 @@
+const getSourceModules = (sourceModulePath) => {
+    const cleanSource = sourceModulePath.replace('./src', '');
+
+    const sourceModules = require('../../../../src' + cleanSource);
```
greg-a-smith: I know we trudged through getting this to work, but it might be worth dropping a comment in here as to _why_ we are stripping out `src` and then adding it back in and _why_ we're doing the string concatenation the way we are.

PR #396, opened 2019-02-11
fix: Remove offset for start of week date in Calendar

### LINKED ISSUE #358: DatePicker Component: disableWeekends prop sometimes malfunctions
### Description

I think some digging will be required to figure this out, but I can only currently recreate it with June 2025. 

See gif:

### Versions
**fundamental-react:** 0.3.1
**fiori-fundamentals:**

---

_**DISCLAIMER:**

mjaycub:
### Description
The original issue (#358) describes the issue within `DatePicker` but it actually lies in the `Calendar` component calculation. The offset used for the calculation for the starting day of the month was causing issues with any month that started on Sunday. The offset could be increased to 7 (rather than 6) or removed completely (which is what I decided to do in this commit). The only difference would be that the previous week from previous month would be shown also.

For example, June 1st 2025 is a Sunday.

**Before fix** - June 1st 2025 is shown as a Saturday (incorrect) and Friday & Saturday are disabled rather than Saturday & Sunday.

screenshot 2019-02-11 at 22 04 50

**After Fix** - the day is corrected (June 1st 2025 is a Sunday) and Saturday & Sunday are disabled as expected.

screenshot 2019-02-11 at 22 05 05

fixes #358

### DIFF
--- a/src/Calendar/Calendar.js
+++ b/src/Calendar/Calendar.js
@@ -290,7 +290,7 @@ export class Calendar extends Component {
 
     retrieveStartOfWeek = (date) => {
         let day = date.getDay();
-        let difference = date.getDate() - day + (day === 0 ? -6 : 0);
+        let difference = date.getDate() - day;
         return new Date(date.setDate(difference));
     }
 


bcullman:
I think we can back this by a snapshot test. Consider opening prerequisite PR that does a calendar snapshot of `sep 2019`. 

Once that merges, *this* PR should rebase, and then we should see it fail on that snapshot (which is good because this PR will fix it)

mjaycub:
> I think we can back this by a snapshot test. Consider opening prerequisite PR that does a calendar snapshot of `sep 2019`.

Ok I'll create a separate PR to do this and will follow up.

greg-a-smith:
To avoid the calendar (or DatePicker) "jumping around", I think it would be good to always have it show 6 rows of dates.  The question (as you noted) is for months that start on a Sunday, should the first row be days 1-7 of the current month or should they be the last 7 days of the previous month?  My preference would be the last week of the previous month so regardless of which day the month starts on, there will always be previous month's days shown at the top and next month's days shown at the bottom.

Looking at another library that we have used, [react-datepicker](https://reactdatepicker.com/#example-38) offers a `fixedHeight` prop to force 6 rows of dates, but for months starting on Sunday, it shows 1-7 of the current month on the first row.  This leads to almost 2 weeks of the next month being shown at the bottom.  I guess this is okay, but as I said, I tend to lean toward always showing previous _and_ next month's days surrounding the current month.

This is probably something that should go to the Fiori Fundamentals team so that whatever route is chosen can be propagated through all the libraries.

PR #398, opened 2019-02-12
fix: Playground: Add a 404.html identical to index.html

skvale:
### Description
In the playground, hosting with Github pages doesn't support routing with SPAs very well because there is no server redirecting all requests to the index.html file.  

This is causing page refreshes or direct links to non-root routes (like https://sap.github.io/fundamental-react/datepicker) to return the Github 404 page.

This PR allows page refreshes and direct links by creating a 404.html file identical to the index.html file as part of the deploy script.  Github will direct 404s to this 404.html. 

It also provides a simple component for rendering a 404 within the docs

screen shot 2019-02-13 at 9 52 41 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "build-css": "node-sass-chokidar --include-path ./src --include-path ./node_modules src/ -o src/",
-    "build-doc": "node scripts/build.js",
+    "build-doc": "node scripts/build.js && cp build/index.html build/404.html",
     "build-js": "node scripts/build.js",
     "build:lint:fix": "eslint --fix 'src/**' --ext .js,.jsx --env browser,node",
     "build:lint": "eslint 'src/**' --ext .js,.jsx --env browser,node",
@@ -20,7 +20,7 @@
     "build:index": "babel-node devtools/buildIndexFiles.js",
     "config:lint": "eslint 'config/**' --ext .js,.jsx --env browser,node",
     "config:lint:fix": "eslint --fix 'config/**' --ext .js,.jsx --env browser,node",
-    "deploy": "cp build/index.html build/404.html && gh-pages -d build",
+    "deploy": "gh-pages -d build",
     "devtools:lint": "eslint 'devtools/**' --ext .js,.jsx --env browser,node",
     "lint:fix": "npm run build:lint:fix  && npm run scripts:lint:fix",
     "lint:pre-commit": "printf \"running pre-commit lint...\"  && npm run lint && printf \"done!\n\"",
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "build:index": "babel-node devtools/buildIndexFiles.js",
     "config:lint": "eslint 'config/**' --ext .js,.jsx --env browser,node",
     "config:lint:fix": "eslint --fix 'config/**' --ext .js,.jsx --env browser,node",
-    "deploy": "gh-pages -d build",
+    "deploy": "cp build/index.html build/404.html && gh-pages -d build",
     "devtools:lint": "eslint 'devtools/**' --ext .js,.jsx --env browser,node",
     "lint:fix": "npm run build:lint:fix  && npm run scripts:lint:fix",
     "lint:pre-commit": "printf \"running pre-commit lint...\"  && npm run lint && printf \"done!\n\"",
--- a/src/_playground/Routes.js
+++ b/src/_playground/Routes.js
@@ -273,7 +273,7 @@ const routes = [
 
 const RouteNotFound = () => (
     <div>
-        <Header>404: Not found</Header>
+        <Header>Sorry, page not found.</Header>
         <Link to=''>Home</Link>
     </div>
 );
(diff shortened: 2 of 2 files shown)


src/_playground/Routes.js
```diff
@@ -270,6 +271,12 @@ const routes = [
     }
 ];
 
+const NoMatch = () => (
+    <div>
```
bcullman: `<header>` has close tag. Is outer `<div>` needed?

src/_playground/Routes.js
```diff
@@ -270,6 +271,10 @@ const routes = [
     }
 ];
 
+const NoMatch = () => (
+    <Header>No match</Header>
+);
```
bcullman: this is not *quite* a 404. as, the left nav is still visible. But...

I would *consider* altering the title to say `404: not found`
I would *definitely* add a link back to home as part of a page description text.

bcullman:
i not sure I like this. It will result in urls like:
`https://sap.github.io/fundamental-react/#/actionBar.`

The fundamentals-ngx and fundamentals-vue team are actively working to updating their docs so that *all* libraries can live at the same relative location, ie:

`https://sap.github.io/fundamental-[react/vue/ngx]/actionBar`

if you are trying to make it so that component don't cause navigation away, i suggest we set all links `javascript:void(0)` or some other non-navigable value

(I do think we need some sort of 404-like handler, though)

skvale:
That makes sense.  

I think the biggest issue is the use case of somebody sharing a link. 
 
For example, sharing the datepicker: https://sap.github.io/fundamental-react/datepicker <-- 404

Everything except https://sap.github.io/fundamental-react is a 404.

An alternative could be something that creates a 404.html identical to index.html
```json
"build-doc": "node scripts/build.js && cp build/index.html build/404.html",
```

greg-a-smith:
I agree with @bcullman's thoughts.  I like the idea of a 404 handler, but not sure I like injecting `#` characters into the URLs.  If it just ran locally in a dev mode, it would be no problem, but since we are trying to align all the fundamental libraries to follow the same URL pattern, this might not be the right solution.

bcullman:
We actually were not seeing the 404 issue until we investigated this a bit more, and tried it in a few alternate browsers (was working fine in my primary browser) 

So, we now understand you are trying to accomplish. For others, the issue is laid out in this stack overflow 

https://stackoverflow.com/questions/46056414/getting-404-for-links-with-create-react-app-deployed-to-github-pages

2 Solutions are proposed, the `#` and the 404 handler redirect to index. I *think* we proposing going down the 404 handler route, but I think we would like to see what the performance of that solution would look/feel like

bcullman:
it appears that fundamentals-ngx (https://sap.github.io/fundamental-ngx/docs/home) uses the 404 approach.

skvale:
I could update this pr to these changes
```json
// package.json
"deploy": "npm run build-doc && cp build/index.html build/404.html && gh-pages -d build",
```

```jsx
// Routes.js
const RouteNotFound = () => (
    <div>
        <Header>404: Not found</Header>
        <Link to=''>Home</Link>
    </div>
);
```
Or move this conversation to an issue and close the PR.

greg-a-smith:
@skvale We chatted about this today during our stand-up meeting and we'd prefer the 404 solution to keep the URLs clean (especially since we are trying to align URL patterns between `fundamental`, `fundamental-react`, `fundamental-ngx` and `fundamental-vue`).  If you wanted to make that change in this PR, we will review it.  Thanks for the contribution!

bcullman:
i think you still need to supply an actual 404 page. something like:

https://github.com/rafrex/spa-github-pages/blob/gh-pages/404.html

or even tighter:

```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <meta http-equiv="X-UA-Compatible" content="ie=edge">
  <title>404 Not Found | My App</title>
</head>
<body>
  <script>
    (function redirect() {
      if (document && document.location) {
        document.location.replace(`/?redirect=${encodeURIComponent(document.location.pathname)}`);
      }
    }());
  </script>
</body>
</html> 
```

skvale:
That 404 route isn't the change I am suggesting.  I think copying index.html to 404.html works better.

Updating the package.json script to
```json
"deploy": "cp build/index.html build/404.html && gh-pages -d build",
```

I published an example of this PR here: https://skvale.github.io/fundamental-react

package.json
```diff
@@ -20,7 +20,7 @@
     "build:index": "babel-node devtools/buildIndexFiles.js",
     "config:lint": "eslint 'config/**' --ext .js,.jsx --env browser,node",
     "config:lint:fix": "eslint --fix 'config/**' --ext .js,.jsx --env browser,node",
-    "deploy": "gh-pages -d build",
+    "deploy": "cp build/index.html build/404.html && gh-pages -d build",
```
greg-a-smith: Could we add the `cp` statement to the end of the `build-doc` script instead?  We reduced the `deploy` script to just the `gh-pages` command in order to pass additional arguments to it (from `publish.sh`) and I don't believe that will work if this is two separate commands.

PR #400, opened 2019-02-12
chore: add prop-types as a dependency

mobot11:
### Description
We reference `prop-types` in our library but do not list it as a dependency. I _believe_ this works because `annotated-prop-types` lists `prop-types` as a dependency. We should list it as a dependency to avoid future bugs. Thanks @jbadan for pointing this out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
   "dependencies": {
     "classnames": "^2.2.6",
     "fiori-fundamentals": "^1.4.0",
-    "focus-trap-react": "^6.0.0"
+    "focus-trap-react": "^6.0.0",
+    "prop-types": "^15.7.1"
   },
   "devDependencies": {
     "@babel/cli": "^7.1.5",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -45,7 +45,8 @@
   "dependencies": {
     "classnames": "^2.2.6",
     "fiori-fundamentals": "^1.4.0",
-    "focus-trap-react": "^6.0.0"
+    "focus-trap-react": "^6.0.0",
+    "prop-types": "^15.7.1"
```
bcullman: hrm. this can probably be a devDep.

mobot11: I don't think I agree with that. It needs to be exported in our built code so consumers receive prop-type warnings.

mobot11: https://github.com/facebook/prop-types#how-to-depend-on-this-package

bcullman: im sold. 🚢

PR #404, opened 2019-02-14
chore: add search to documentation website

### LINKED ISSUE #296: Improve Docs: Add Search
### Description
Follow design seen here:

image

CORUI-6169

leannerivera:
### Description
Add search to documentation website.

fixes #296


src/_playground/App.scss
```diff
@@ -3,6 +3,7 @@ $fd-fonts-path: '../../node_modules/fiori-fundamentals/dist/fonts/';
 @import '../../node_modules/fiori-fundamentals/scss/all.scss';
 
 // Fundamental React Docs Styles
+
```
bcullman: unnecessary whitespace diff.

src/_playground/App.scss
```diff
@@ -3,6 +3,7 @@ $fd-fonts-path: '../../node_modules/fiori-fundamentals/dist/fonts/';
 @import '../../node_modules/fiori-fundamentals/scss/all.scss';
```
bcullman: file has unnecessary whitespace diffs

src/_playground/Routes.js
```diff
@@ -270,59 +271,99 @@ const routes = [
...
-                            key={route.url}
-                            to={{ pathname: route.url }}>
-                            {route.name}
-                        </NavLink>
-                    </li>
-                ))}
+    render() {
+        let sectionRoutes;
+        let routedItems;
+        this.state.query === '' ? routedItems = routes : routedItems = this.state.filteredItems;
```
greg-a-smith: I don't _think_ you need to do this.  You are storing the routes in the `filteredItems` in state so you should just always be able to use that.

src/_playground/Routes.js
```diff
@@ -270,59 +271,99 @@ const routes = [
...
+                </ul>
+            );
+        });
+
+        let navHomeSegment = navItems[0];
+        navItems = navItems.splice(1);
+        const noItemsFound = (
+            <ul className='frDocs-Nav__list'>
+                <li className='frDocs-Nav__headers'>
+                    No results found.
```
greg-a-smith: The "No results found." message shouldn't be a header.  All applicable section headers should still display and then the "No results found." message should be smaller text below those.  Again, the exact implementation depends on if this is a global nav search or just a component search.

src/_playground/Routes.js
```diff
@@ -270,59 +271,99 @@ const routes = [
...
+        };
+    }
 
-    const navItems = sections.sort(sortBy('sortOrder', 'name')).map(section => {
-        if (!groupedRoutes[section.name]) {
-            return;
-        }
+    onChangeHandler = (event) => {
+        let searchResults = routes.filter((navItem) => {
+            return navItem.sortOrder === 1 ? navItem : navItem.name.toLowerCase().includes(event.target.value.toLowerCase());
```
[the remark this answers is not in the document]
leannerivera: Thanks @greg-a-smith . Actually was trying to implement the spacing with sass, but it seems to be overwritten with that component's (using the inputgroup component) styling.

Re: the search, I believe this is purely meant to be a component search (like in the fundamental-ngx documentation). Thanks for the comment, will work on filtering be section instead of order. 👍

src/_playground/Routes.js
leannerivera: You're right. This is an artifact from a different implementation, so I've removed it now. It should now be fixed in latest push.

src/_playground/Routes.js
```diff
@@ -270,59 +271,97 @@ const routes = [
...
+        const groupedRoutes = groupArray(this.state.filteredItems, 'section');
+
+        let navItems = sections.sort(sortBy('sortOrder', 'name')).map(section => {
+            if (!groupedRoutes[section.name]) {
+                return;
+            }
+
+            sectionRoutes = groupedRoutes[section.name].sort(sortBy('sortOrder', 'name'));
+            return (
+                <ul className='frDocs-Nav__list' key={section.name}>
```
jeffredodd: It might be worth breaking this rendering logic out so it can be re-used for other menu items that might not be searchable.

src/_playground/Routes.js
```diff
@@ -270,59 +271,97 @@ const routes = [
...
+                                to={{ pathname: route.url }}>
+                                {route.name}
+                            </NavLink>
+                        </li>
+                    ))}
+                </ul>
+            );
+        });
+
+        let navHomeSegment = navItems[0];
```
jeffredodd: It might be worth filtering out the `Home` object when you create `navItems`. This would allow you to avoid hard coding `0` a location on an array and having to use splice. 

You could then call the home route and render it manually using the render function. I think `routes.find({name: 'home'})` would return just the home object as well.

src/_playground/Routes.js
leannerivera: Are the changes I pushed close to what you mean? It needs to be refactored but I want to make sure I'm going in the right direction.

src/_playground/Routes.js
leannerivera: Hi Greg: Am I going in the right direction with what you mean in all your comments with my latest push?

src/_playground/Routes.js
```diff
@@ -270,59 +273,124 @@ const routes = [
...
-                            {route.name}
-                        </NavLink>
-                    </li>
-                ))}
+    render() {
+        let sectionRoutes;
+        let freezeItems;
+        let navItems;
+        const groupedRoutes = groupArray(this.state.filteredItems, 'section');
+        //TODO: remove navItems declaration here
```
greg-a-smith: Is there still a "TODO" here?

src/_playground/Routes.js
```diff
@@ -270,59 +273,124 @@ const routes = [
...
+        sections.sort(sortBy('sortOrder', 'name')).map(section => {
+            if (!groupedRoutes[section.name]) {
+                return;
+            }
+            // eslint-disable-next-line no-console
+            console.log('line 304', groupedRoutes[section.omitSearch]);
+
+            sectionRoutes = groupedRoutes[section.name].sort(sortBy('sortOrder', 'name'));
+            // TODO: conditional return here, navItems and freezeItems for omitSearch:true items
```
greg-a-smith: Is there still a "TODO" here?

src/_playground/Routes.js
```diff
@@ -270,59 +273,124 @@ const routes = [
...
+                                </NavLink>
+                            </li>
+                        ))}
+                    </ul>
+                );
+            }
+        });
+        // eslint-disable-next-line no-console
+        console.log('line 326', navItems);
+        // TODO: remove hard-coding here
```
greg-a-smith: Is there still a "TODO" here?

src/_playground/Routes.js
```diff
@@ -270,59 +273,124 @@ const routes = [
...
-                        })}
-                        <Redirect exact from=''
-                            to='/home' />
-                    </Switch>
+        return (
+            <BrowserRouter basename={process.env.PUBLIC_URL}>
+                <div className='frDocs-Container'>
+                    <div className='frDocs-Sidebar'>
+                        <h1 className='frDocs-Logo'>FUNDAMENTAL REACT</h1>
+                        {/* TODO: change var for navHome and navItems */}
```
greg-a-smith: Is there still a "TODO" here?

src/_playground/Routes.js
```diff
@@ -270,59 +273,124 @@ const routes = [
...
+        let freezeItems;
+        let navItems;
+        const groupedRoutes = groupArray(this.state.filteredItems, 'section');
+        //TODO: remove navItems declaration here
+        sections.sort(sortBy('sortOrder', 'name')).map(section => {
+            if (!groupedRoutes[section.name]) {
+                return;
+            }
+            // eslint-disable-next-line no-console
+            console.log('line 304', groupedRoutes[section.omitSearch]);
```
greg-a-smith: This `console.log` should be removed.

src/_playground/Routes.js
```diff
@@ -270,59 +273,124 @@ const routes = [
...
+                                    {route.name}
+                                </NavLink>
+                            </li>
+                        ))}
+                    </ul>
+                );
+            }
+        });
+        // eslint-disable-next-line no-console
+        console.log('line 326', navItems);
```
greg-a-smith: This `console.log` should be removed.

src/_playground/Routes.js
greg-a-smith: Yes, it's certainly going in the right direction. 😄

src/_playground/Routes.js
leannerivera: Definitely will be removed for merge. :)

src/_playground/Routes.js
```diff
@@ -38,11 +38,13 @@ import { ToggleComponent } from '../Toggle/Toggle.Component';
...
 import { TreeComponent } from '../Tree/Tree.Component';
 import { BrowserRouter, NavLink, Redirect, Route, Switch } from 'react-router-dom';
+import React, { Component } from 'react';
 
 const sections = [
     {
         name: 'Getting Started',
-        sortOrder: 1
+        sortOrder: 1,
+        omitSearch: true //omit this category from search filter
```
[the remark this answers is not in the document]
leannerivera: Okay I have changed it so that now I believe the omitSearch in sections is getting used for the freezeItems var in my latest push.

src/_playground/Routes.js
leannerivera: All removed. :)

src/_playground/Routes.js
```diff
@@ -270,59 +273,117 @@ const routes = [
...
+                            {freezeItems}
+                            <div className='frDocs-Nav__inputGroup'>
+                                <InputGroup
+                                    inputId='frDocs-Nav__search'
+                                    inputPlaceholder='Search'
+                                    inputType='search'
+                                    inputValue={this.state.query}
+                                    onChange={this.onChangeHandler} />
+                            </div>
+                            {this.state.filteredItems.length === 1 ? noItemsFound : navItems}
```
jbadan: Should this be hardcoded to 1 or should it take into account how many items are `omitSearch: true`?

greg-a-smith:
@leannerivera, @SAP/fundamental-react-team  The more I look at this, the more I think this has gotten overly complicated simply due to the placement of the search field.  There is the need for some strange workarounds to render one section about the search field and all others below it.  I know we patterned after `fundamental-ngx`, but maybe it's time to set a new pattern.  It seems fairly likely that the "Getting Started" section will have additional links at some point and those should also be searchable.

What if we just moved the search input above all sections (just below the logo) and had it search everything?  We could still omit "Home" from the search so it would always appear.  Then there could just simply be a single array filter and a single loop to render nav items.  I have a hunch the search may eventually move into a top bar (similar to hig.concur.com) and this would make that transition much simpler.

Thoughts?

jbadan:
@greg-a-smith That seems like a super reasonable middle ground. I know we want all the docs sites to eventually match, so it might be worth getting @SAP/fundamental-ngx-team to weigh in as well.

leannerivera:
Yes I agree @greg-a-smith . It is just much cleaner and easier to de-bug with the search bar above all elements rather than sandwiched in-between. If we can get a consensus on this it would be great.

src/_playground/Routes.js
leannerivera: okay, will do this.

leannerivera: I just double-checked and added more omitSearch:true items and it still works 🤷‍♀️. I'm going to play around with this a little more and see how this is actually happening.

leannerivera: Okay, I believe this may be resolved now. :)

greg-a-smith:
@leannerivera I'm glad the `fundamental-ngx` group agreed to our proposal.  This is _MUCH_ cleaner!

I added just a few minor cleanup items:

1. Restored the `RouteNotFound` logic that had been merged to `master` earlier in the day (not sure what happened to it).
1. Removed `omitSearch` from sections as it's not used.
1. Removed the "No results found" logic since we have "Home" that will always appear.  It's also more of an "filter" rather than a "search" so when the left nav is empty, everything has been filtered out.

PR #405, opened 2019-02-14
Refactor FormRadio Component

### LINKED ISSUE #247: Consider using a component for the inputs on FormRadio
Currently, the `FormRadio` component is really more of a radio group component and the `inputs` prop accepts objects that are used to build the labels and inputs for each radio button.  Since the component internally iterates over a map of these inputs, it makes prop spreading more difficult.  And, it kind of doesn't follow the pattern of other components where there are React components for the inner elements (_e.g._ `ActionBar`).

Either `FormRadio` should become `FormRadioGroup` and then `FormRadio` can be a component that encapsulate the input and label elements for each radio button.

Or, `FormRadio` remains the "group" component (as it is today), but a new component is created to encapsulate the input and label for each radio button.

jeffredodd:
### Description
Initial pass at the refactor of `FormRadio` into `FormRadioGroup` and the initial creation of `FormRadioItem`.  This fixes #247 .

A few things to note...

- Snapshots have HTML changed significantly as Fundamental's only nests `input` inside of `label`. 
- State has been removed from `FormRadioGroup`. Consumers must pass `checked` or `defaultChecked` to a `FormRadioItem` and maintain state using `onChange` callbacks.
- `ShortId` has been introduced in both this PR as well as #394 .
- Both component and unit test examples both used `name` incorrectly and is now correct.


src/Forms/Forms.js
```diff
@@ -227,76 +228,95 @@ FormSelect.propTypes = {
...
-                        value={inputItem.value} />
-                    <label className='fd-form__label' htmlFor={inputItem.id}>
-                        {inputItem.label}
-                    </label>
-                </div>
-            ));
-        }
-        return <div>{result}</div>;
+        const { children, disabled, isInline, onChange, ...props } = this.props;
+
```
jbadan: Should this also have classnames so we can pass a custom one if needed?

jeffredodd: This should be handled by the `...props` spread. 😄

src/Forms/Forms.js
```diff
@@ -227,76 +228,95 @@ FormSelect.propTypes = {
...
+        return (
+            <div
+                {...props}>
+                {React.Children.map(children, child => {
+                    if (React.isValidElement(child)) {
+                        return React.cloneElement(child, {
+                            disabled: child.props.disabled ? child.props.disabled : disabled,
+                            inline: child.props.inline ? child.props.inline : isInline,
+                            name: child.props.name ? child.props.name : this.groupId,
```
greg-a-smith: I think this can be simplified to `name: child.props.name || this.groupId`

src/Forms/Forms.js
```diff
@@ -227,76 +228,95 @@ FormSelect.propTypes = {
...
+                            onChange: child.props.onChange ? child.props.onChange : onChange
```
greg-a-smith: I think this can be simplified to `name: child.props.onChange || onChange`

src/Forms/Forms.js
```diff
@@ -227,76 +228,95 @@ FormSelect.propTypes = {
...
-FormRadio.displayName = 'FormRadio';
 
-FormRadio.propTypes = {
-    inputs: PropTypes.array.isRequired,
-    defaultChecked: PropTypes.string,
+FormRadioGroup.propTypes = {
+    children: PropTypes.node,
     disabled: PropTypes.bool,
-    isInline: PropTypes.bool
+    isInline: PropTypes.bool,
```
greg-a-smith: Should this prop just be named `inline` to match the `FormRadioItem` component.

src/Forms/Forms.js
```diff
@@ -227,76 +228,95 @@ FormSelect.propTypes = {
...
+    id: PropTypes.string,
+    inline: PropTypes.bool,
+    name: PropTypes.string,
+    value: PropTypes.string
+};
+
+FormRadioItem.propDescriptions = {
+    checked: 'Set to **true** when radio input is checked and a controlled component.',
+    defaultChecked: 'Set to **true** when the radio input is checked and an uncontrolled component.',
+    inline: 'Set to **true** when radio input is inline.',
```
greg-a-smith: Is this actually an internal prop (or should it be)?  Would you ever have a radio item that is inline and the radio group isn't (or vice versa)?

src/Forms/Forms.js
```diff
@@ -227,76 +228,95 @@ FormSelect.propTypes = {
...
+    disabled: PropTypes.bool,
+    id: PropTypes.string,
+    inline: PropTypes.bool,
+    name: PropTypes.string,
+    value: PropTypes.string
+};
+
+FormRadioItem.propDescriptions = {
+    checked: 'Set to **true** when radio input is checked and a controlled component.',
+    defaultChecked: 'Set to **true** when the radio input is checked and an uncontrolled component.',
```
greg-a-smith: Should `defaultChecked` be on `FormRadioGroup`?

src/Forms/Forms.js
```diff
@@ -227,76 +228,95 @@ FormSelect.propTypes = {
...
+    defaultChecked: PropTypes.bool,
+    disabled: PropTypes.bool,
+    id: PropTypes.string,
+    inline: PropTypes.bool,
+    name: PropTypes.string,
+    value: PropTypes.string
+};
+
+FormRadioItem.propDescriptions = {
+    checked: 'Set to **true** when radio input is checked and a controlled component.',
```
greg-a-smith: How does it handle the `checked` prop if more than one item is set to true?

src/Forms/Forms.js
jeffredodd: This used to be on `FormRadioGroup` to set the default checked item. However this is pretty poor naming, `defaultChecked` and `checked` are both propTypes react supports for `input[type"radio"]` and belong on the `FormRadioItem`. 

We could omit defaultChecked from propTypes and it would still spread to the radio correctly. Should we do that?

See documentation here. https://reactjs.org/docs/uncontrolled-components.html

src/Forms/Forms.js
jeffredodd: I would guess React would be the one in charge of that behavior when they receive both? It can't be both a controlled and uncontrolled input at the same time that is for sure.

src/Forms/Forms.js
jeffredodd: > This used to be on `FormRadioGroup` to set the default checked item. However this is pretty poor naming, `defaultChecked` and `checked` are both propTypes react supports for `input[type"radio"]` and belong on the `FormRadioItem`.
> 
> We could omit defaultChecked from propTypes and it would still spread to the radio correctly. Should we do that?
> 
> See documentation here. https://reactjs.org/docs/uncontrolled-components.html

Issue #406, opened 2019-02-14
Cannot access component documentation pages


cellio365:
### Description
I cannot access documentation pages except their home.
e.g., I get 404 to access actionBar component page:
https://sap.github.io/fundamental-react/actionBar

### Versions
**fundamental-react:**
**fiori-fundamentals:**

---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

bcullman:
Looks like we hit this on 0.3.3 release. We are investigating...

bcullman:
this will be fixed by #407. 

unfortunately, we are blocked because coveralls is not returning coverage status. We are now looking in to that.

PR #407, opened 2019-02-14
fix: change require.resolve to path.join to fix webpack prod issues

mobot11:
### Description
Webpack handles require.resolve differently in production mode then node does. Switching to path.join fixes the issue.

fixes https://github.com/SAP/fundamental-react/issues/406

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "scripts:lint:fix": "eslint --fix 'scripts/**' --ext .js,.jsx --env browser,node",
     "size": "npm run build && size-limit",
     "size:debug": "npm run build && size-limit --why",
-    "start": "node scripts/start.js",
+    "start": "npm run build && node scripts/start.js",
     "start:playground": "FUNDAMENTAL_REACT_PLAYGROUND=true npm start",
     "std-version": "standard-version -m \"chore(release): version %s build ${TRAVIS_BUILD_NUMBER} [ci skip]\"",
     "test:coverage": "jest --coverage",
--- a/src/ActionBar/ActionBar.Component.js
+++ b/src/ActionBar/ActionBar.Component.js
@@ -1,3 +1,4 @@
+import path from 'path';
 import React from 'react';
 import { ActionBar, ActionBarActions, ActionBarBack, ActionBarHeader, Button, Menu, MenuItem, MenuList, Popover } from '../';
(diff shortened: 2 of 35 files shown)


src/ActionBar/ActionBar.Component.js
```diff
@@ -94,11 +95,11 @@ const clickBackBtn = () => {
                 The **Action Bar** is located at the top of the page and is used for page title and main actions for the
                 page.
             </Description>
-            <Import sourceModulePath={require.resolve('./ActionBar')} />
+            <Import sourceModulePath={path.join(__dirname, './ActionBar')} />
```
greg-a-smith: I am getting the following error when I attempt to run it in dev mode:
screen shot 2019-02-14 at 8 42 17 am

I'm not sure if this fixes it in the webpack mode, but it definitely doesn't work in local dev mode.  Is it possible that we need two different solutions?

jbadan: @greg-a-smith  I got it working in dev mode by running `npm run build` then `npm start`, I think that's all it needs.

jbadan: We could add a node environment check before this and switch out `lib` for `src` and avoid having to run build first: https://github.com/SAP/fundamental-react/blob/master/src/_playground/documentation/Import/Import.js#L17

greg-a-smith: I don't think a node environment check will work because we are running the docs in development mode to make sure `propTypes` don't get removed.

greg-a-smith: But if doing a `npm run build` solves it, then maybe we just wire that into the `npm start` script.

mobot11: I was thinking along the same line as @greg-a-smith I'll add build to the start script.

package.json
```diff
@@ -31,7 +31,7 @@
     "scripts:lint:fix": "eslint --fix 'scripts/**' --ext .js,.jsx --env browser,node",
     "size": "npm run build && size-limit",
     "size:debug": "npm run build && size-limit --why",
-    "start": "node scripts/start.js",
+    "start": "npm run build && node scripts/start.js",
```
bcullman: hrm... How did this (or could this) work before, then?

mobot11: I'm not entirely sure. I was going to look into it more after this got merged.

PR #413, opened 2019-02-15
fix: update index files generation to handle multiple component files in single directory

mobot11:
### Description
@chrismanciero found a bug that index file generation was failing on Travis when more than one component file existed in a directory. This fixes it.

### DIFF
--- a/devtools/buildIndexFiles.js
+++ b/devtools/buildIndexFiles.js
@@ -17,16 +17,22 @@ const isComponentFile = (source) => {
 const componentDirs = readdirSync(srcPath).map(name => path.join(srcPath, name)).filter(isComponentDirectory).map(directory => {
     return {
         path: directory,
-        fileName: readdirSync(directory).filter(isComponentFile)
+        fileNames: readdirSync(directory).filter(isComponentFile)
     };
 });
 
+// For every component directory.
 componentDirs.map((directory) => {
-    let component = require(`${directory.path}/${directory.fileName}`);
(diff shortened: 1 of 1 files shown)


devtools/buildIndexFiles.js
```diff
@@ -17,16 +17,22 @@ const isComponentFile = (source) => {
...
-    let component = require(`${directory.path}/${directory.fileName}`);
     let fileContents = '';
-    Object.keys(component).map((single) => {
-        fileContents += `export { ${single} } from './${directory.fileName}';\n`;
-        let indexPath = `${directory.path}/index.js`;
-        writeFileSync(indexPath, fileContents);
+    // Loop through it's files.
+    directory.fileNames.map((name) => {
+        // Grab the file's exports.
+        let component = require(`${directory.path}/${name}`);
```
greg-a-smith: Should `${directory.path}` and `${name}` be joined with `path.join` instead of a hard-coded slash (`'/'`)?

devtools/buildIndexFiles.js
```diff
@@ -17,16 +17,22 @@ const isComponentFile = (source) => {
...
+        Object.keys(component).map((single) => {
+            fileContents += `export { ${single} } from './${name}';\n`;
+        });
     });
+    // write the index file into the directory.
+    let indexPath = `${directory.path}/index.js`;
```
greg-a-smith: Should these be joined with `path.join`?

PR #414, opened 2019-02-15
fix: tabs component refactor

### LINKED ISSUE #344: Tabs shift in size when content changes
### Description
When changing through tabs, they shift in sidesize. 
screen shot 2019-02-01 at 9 49 03 am
screen shot 2019-02-01 at 9 48 58 am

While on Fundamentals they do not. https://sap.github.io/fundamental/components/tabs.html
screen shot 2019-02-01 at 9 54 06 am
screen shot 2019-02-01 at 9 54 03 am

Most likely this can be fixed by locating the tab content outside of the `Tab`'s `<ul>` component.

chrismanciero:
### Description
This is a complete refactor of the Tabs component which replaces the single file that contained both the **TabGroup** and **Tab** components with separate files for each component. This PR also adds a new component, **TabContent**, which is used to display the content of a tab. Each component resides in the /Tabs directory

- /Tabs/Tab.js - a component used to display the clickable link of the tab strip
- /Tabs/TabContent.js - a component used to display the contents of a selected tab
- /Tabs/TabGroup.js - a component that wraps the Tab and TabContent components. The TabGroup handles the placement of the Tab strip as well as the content.

As seen on the Fiori Fundamentals Tab page (https://sap.github.io/fundamental/components/tabs.html). The **Tabs with list element** is the implementation used. The **Tabs with nav element** is not implemented as there does not seem to be a valid use-case to have 2 implementations.

fixes #344


src/Tabs/TabContent.js
```diff
@@ -0,0 +1,36 @@
+import classnames from 'classnames';
```
bcullman: @chrismanciero - the SAP Concur team discussed the pattern to be used for subcomponents, and wanted to use a "private" component structure.

to do this we typically prefix private components with underscores, and import/export them as part of the parent component. Reach to @greg-a-smith or others for more guidance.

src/Tabs/TabGroup.js
```diff
@@ -0,0 +1,97 @@
...
+
+    // create content to show below tab list
+    renderContent = () => {
+        return React.Children.map(this.props.children, (child) => {
+            return (
+                <TabContent
+                    {...this.props.tabContentProps}
+                    selected={this.state.selectedId === child.props.id}>
+                    {child.props.children}
+                </TabContent>);
```
greg-a-smith: This is an interesting approach (using an internal component) to help encapsulate the details for tab content.  Per @bcullman's comment, we should just take a couple other steps to make sure this stays "private".

src/index.js
```diff
@@ -52,8 +52,10 @@ export { ProductTileContent } from './Tile/Tile';
 export { ProductTileMedia } from './Tile/Tile';
 export { SearchInput } from './SearchInput/SearchInput';
 export { SideNav, SideNavList, SideNavListItem } from './SideNavigation/SideNavigation';
+export { Tab } from './Tabs/Tab';
+export { TabContent } from './Tabs/TabContent';
```
greg-a-smith: I think the `TabContent` component should be internal-only and should not be exported.

src/Tabs/Tabs.test.js
```diff
@@ -1,101 +1,56 @@
 import { mount } from 'enzyme';
```
greg-a-smith: Now that these are separate components, we should have unit tests for each individually.

src/Tabs/Tab.js
```diff
@@ -0,0 +1,50 @@
...
+            aria-disabled={disabled}
+            aria-selected={selected}
+            className={linkClasses}
+            href={!disabled ? `#${id}` : null}
+            onClick={!disabled ? (event) => {
+                props.onClick(event, id);
+            } : null}
+            role='tab'>
+            {title}
+        </a>
```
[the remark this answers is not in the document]
chrismanciero: Yes, this needs to be brought up to Fiori Fundamentals because of styling issues. If i change from `<a>` tags to `<button>` tags the design looks bad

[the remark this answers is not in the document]
jbadan: @greg-a-smith This change is part of the a11y story we have in the backlog for Tabs.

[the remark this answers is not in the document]
greg-a-smith: Awesome! 👍

src/Tabs/Tab.js
```diff
@@ -0,0 +1,53 @@
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+export const Tab = (props) => {
+    const { title, className, disabled, glyph, id, selected, ...rest } = props;
```
greg-a-smith: Should probably add `onClick` to the list of props being pruned out.

src/Tabs/TabGroup.js
```diff
@@ -0,0 +1,111 @@
...
+TabGroup.defaultProps = {
+    selectedId: '1'
+};
+
+TabGroup.propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    selectedId: PropTypes.string,
+    tabContentProps: PropTypes.object,
+    tabLinkProps: PropTypes.object,
```
greg-a-smith: The props `tabContentProps` and `tabLinkProps` should be moved to the `Tab` component.  Otherwise, a consumer will not be able to provide different props for each tab and its content.  They can still be referenced in the render methods in this file, but would just come from `child.props` within those loops.

Taking that a step further, once moved to the `Tab` component, I think `tabLinkProps` could just become `linkProps` and those would be spread to the `<a>` element.  Additionally, the `Tab` component could render its own `<li>` and spread `...rest` to that element.

src/Tabs/TabGroup.js
```diff
@@ -0,0 +1,111 @@
...
+    // create tab list
+    renderTabs = () => {
+        return React.Children.map(this.props.children, (child) => {
+            return (
+                <li
+                    {...this.props.tabLinkProps}
+                    className='fd-tabs__item'
+                    key={child.props.id}>
+                    {this.cloneElement(child)}
+                </li>);
```
greg-a-smith: I _think_ it might be better to have `Tab` render both the `<li>` and the `<a>` elements for each tab.  I just think keeps "like" elements together.  Otherwise, `Tab` looks like a misnamed link renderer.  The map can stay here and then `className` and `key` could be moved to `Tab` as well.

src/Tabs/TabGroup.js
```diff
@@ -0,0 +1,111 @@
...
+        });
+    };
+
+    // create content to show below tab list
+    renderContent = () => {
+        return React.Children.map(this.props.children, (child) => {
+            return (
+                <TabContent
+                    {...this.props.tabContentProps}
```
greg-a-smith: See comment below, but I think `tabContentProps` should be coming from `child.props`.

src/Tabs/TabGroup.js
```diff
@@ -0,0 +1,111 @@
...
+                </ul>
+                {this.renderContent(children)}
+            </React.Fragment>
+        );
+    }
+}
+TabGroup.displayName = 'TabGroup';
+
+TabGroup.defaultProps = {
+    selectedId: '1'
```
greg-a-smith: I think this may be problematic.  This relies on the consumer setting the `id` prop on the first tab to `1`.  Maybe this should be changed to a number prop and just simply be the index of the selected tab rather than a specific `id` value.

src/Tabs/TabGroup.js
```diff
@@ -0,0 +1,111 @@
...
+
+    // set selected tab
+    handleTabSelection = (event, id) => {
+        event.preventDefault();
+        this.setState({
+            selectedId: id
+        });
+
+        if (this.props.onTabClick) {
+            this.props.onTabClick(event, id);
```
greg-a-smith: TIP: If you give `onTabClick` a default value of an empty function, you won't need to do the "if exists" check here.

src/Tabs/_TabContent.js
```diff
@@ -0,0 +1,36 @@
...
+            {children}
+        </div >
+    );
+};
+TabContent.displayName = 'TabContent';
+
+TabContent.propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    id: PropTypes.string,
```
greg-a-smith: The `id` prop has no way of getting set since this is a non-exported component.  The way to do it would be to include `id` in the `tabContentProps` object, but then that would end up on the `<div>` by virtue of the `...rest` spread.  The `className` prop is in a similar situation, however, that can stay since you need to have a way to get the `fd-tabs__panel` class on the `<div>`.

src/Tabs/Tab.js
```diff
@@ -0,0 +1,71 @@
...
+        id,
+        selected,
+        onClick,
+        tabContentProps,
+        linkProps,
+        ...rest } = props;
+
+    // css classes used for tabs
+    const linkClasses = classnames(
+        className,
```
greg-a-smith: Now that this renders both the `<li>` and the `<a>`, I think `className` should map to the `<li>` and be combined with `'fd-tabs__item'`.  I don't think we need a `linkClasses` prop simply to add classes to the `<a>` element since they could write a selector using any attributes from the `<li>`.

src/Tabs/TabGroup.js
```diff
@@ -0,0 +1,108 @@
...
+                </ul>
+                {this.renderContent(children)}
+            </React.Fragment>
+        );
+    }
+}
+TabGroup.displayName = 'TabGroup';
+
+TabGroup.defaultProps = {
+    selectedId: '1',
```
greg-a-smith: I think this prop's type needs to change to a number and represent the index of the selected tab.  The reason I say that is because using `id` attribute values requires the use of the `id` prop in the `Tab` component and the `selectedId` prop in `TabGroup`, both of which are currently optional.  It would be difficult to set a default `selectedId` since the tab's `id` value is not easily known.

The index of each tab could be bound to the `onClick` function being passed (in `cloneElement `) so there would be no other work needed.

src/Tabs/_TabContent.js
```diff
@@ -0,0 +1,34 @@
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+export const TabContent = (props) => {
+    const { children, id, selected, className, ...rest } = props;
```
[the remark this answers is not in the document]
chrismanciero: done.

src/Tabs/Tab.js
chrismanciero: done.

src/Tabs/TabGroup.js
chrismanciero: done.

Issue #408, opened 2019-02-14
Bug: InputGroup does not accept custom classNames


jbadan:
### Description

InputGroups do not accept custom classnames, see:
https://github.com/SAP/fundamental-react/blob/master/src/InputGroup/InputGroup.js#L83
https://github.com/SAP/fundamental-react/blob/master/src/InputGroup/InputGroup.js#L123
https://github.com/SAP/fundamental-react/blob/master/src/InputGroup/InputGroup.js#L160
https://github.com/SAP/fundamental-react/blob/master/src/InputGroup/InputGroup.js#L204

Should have additional className like as seen here in Badge: https://github.com/SAP/fundamental-react/blob/master/src/Badge/Badge.js#L14

### Versions
**fundamental-react:** current
**fiori-fundamentals:**

---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

PR #415, opened 2019-02-18
fix: InputGroup now can accept custom classNames

leannerivera:
### Description

InputGroups now accept custom classNames

fixes:  #408


leannerivera:
Done. :)

src/InputGroup/InputGroup.js
```diff
@@ -87,6 +89,7 @@ export class InputGroup extends Component {
                 );
 
                 const inputNumberClasses = classnames(
+                    className,
```
jbadan: It looks like this should be a different prop for the `input` className versus the `div` className. Maybe we should have `className` that get applied to the parent `div` and an `inputClassName` that gets applied to the `input` element.

jbadan:
The new props should also be added to the PropTypes list at the bottom of the file :)

leannerivera:
> The new props should also be added to the PropTypes list at the bottom of the file :)

Do you mean like the changes made in current push?

src/InputGroup/InputGroup.js
```diff
@@ -59,6 +59,8 @@ export class InputGroup extends Component {
 
     render() {
         const {
+            inputClassName,
+            inputType,
```
jbadan: I'm getting duplicate declaration errors on this - I think `inputType` is also on line 74

src/InputGroup/InputGroup.js
```diff
@@ -280,6 +299,7 @@ InputGroup.propDescriptions = {
     addon: 'The value of the add-on.',
     addonPos: 'Location of the add-on relative to the input.',
     inputId: 'Value for the `id` attribute on the `<input>` element.',
+    inputClassName: 'Value for the `classname` attribute on the `<input>` element.',
```
jbadan: Let's copy the current default for classNames - "CSS class(es) to add to the `<input>` element."

jbadan:
This is looking great! I think once we address those two outstanding issues this will be ready to ship! 🎉

src/InputGroup/InputGroup.js
leannerivera: oops looks like I merged this wrong. should be fixed with latest push.

greg-a-smith:
@cjskillingstad is correct.  We are using npm 6.4.1 which is packaged with node 10.15.0.  I also agree that we shouldn't be committing any changes to `package-lock.json` if there are no changes to `package.json`.

@leannerivera Have you setup `nvm` (Node Version Manager) yet?  It's a really nice way of switching Node and NPM versions quickly (since we use different versions on different repos).

leannerivera:
@greg-a-smith @cjskillingstad my mistake. haven't done nvm but will do.

Issue #416, opened 2019-02-18
Consider switching from coveralls to https://codecov.io/ for tracking code coverage


mobot11:
### Description
Coveralls is slow and error prone. We already had to remove the GitHub integration due to performance issues and now are running into issues with the build failing on Travis. https://codecov.io/ is a similar tool and also free for open source repos.

### Versions
**fundamental-react:**

---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

jbadan:
coveralls got better! 🎉

PR #417, opened 2019-02-18
fix: temporarily disable coveralls

jbadan:
### Description
Coveralls is currently causing all builds to fail in Travis. This is a known recurring issue on their end: 
https://github.com/lemurheavy/coveralls-public/issues/1264

This PR allows coveralls to fail without stopping the build.

related to #416

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ script:
 jobs:
   include:
   - stage: Lint and Test
-    script: npm run lint && npm run test && npm run size && npm run test:coveralls
+    script: npm run lint && npm run test:coverage && npm run size && npm run test:coveralls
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,7 +9,7 @@ script:
 jobs:
   include:
   - stage: Lint and Test
-    script: npm run lint && npm run test:coveralls && npm run size
+    script: npm run lint && npm run test && npm run size && npm run test:coveralls
```
mobot11: Can we run `npm run lint && npm run test:coverage && npm run size && npm run test:coveralls` and then edit the `npm run test:coveralls` command to just pipe the coverage info to coveralls instead of running the tests again?

PR #420, opened 2019-02-19
fix: CORUI-6264: One component per file ActionBar

cjskillingstad:
### Description
Separated all components from `ActionBar.js` into their own files. These new files all maintained the git history of the original `ActionBar.js` file.


src/ActionBar/ActionBar.test.js
```diff
@@ -6,7 +6,7 @@ import {
     ActionBarActions,
     ActionBarBack,
     ActionBarHeader
-} from './ActionBar';
+} from './';
```
cjskillingstad: These are being imported from the auto generated `index` file in the component directory. Let me know if I should import directly from the component files.

src/index.js
```diff
@@ -1,7 +1,9 @@
-export { ActionBar } from './ActionBar/ActionBar';
-export { ActionBarBack } from './ActionBar/ActionBar';
-export { ActionBarHeader } from './ActionBar/ActionBar';
-export { ActionBarActions } from './ActionBar/ActionBar';
+export {
+    ActionBar,
+    ActionBarBack,
+    ActionBarHeader,
+    ActionBarActions
+} from './ActionBar';
```
cjskillingstad: These are being imported from the auto generated `index` file in the component directory. Let me know if I should import directly from the component files.

src/ActionBar/ActionBarActions.js
```diff
@@ -0,0 +1,19 @@
...
+
+    return <div {...props} className={actionBarActionsClasses}>{children}</div>;
+};
+
+ActionBarActions.propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string
+};
+
+export default ActionBarActions;
```
greg-a-smith: When it makes sense, I thought we talked about naming the subcomponent files with a leading underscore (_e.g._ `_ActionBarActions`), not exporting them as top-level standalone components, but then adding static references to them in the main component (_e.g._ `ActionBar.Actions = ActionBarActions`).  This way, the consumer just has to import `ActionBar` and they automatically get it and all the subcomponents together.

I don't think this applies unconditionally across the board as we probably wouldn't want to have to code tabs as `TabGroup.Tab` when `Tab` could just be exported.

That said, I think this component is the perfect candidate for making the subcomponents private static references.

cjskillingstad: Forgot about that approach. That cleaned things up quite a bit as well. Fixed in 7f9264d

src/ActionBar/_ActionBarActions.js
```diff
@@ -0,0 +1,19 @@
...
+
+const ActionBarActions = ({ children, className, ...props }) => {
+    const actionBarActionsClasses = classnames(
+        'fd-action-bar__actions',
+        className
+    );
+
+    return <div {...props} className={actionBarActionsClasses}>{children}</div>;
+};
+
```
greg-a-smith: Should non-exported components have a `displayName` set?  Just gathering opinions as I'm not quite sure myself if they should or not.

greg-a-smith: Update:  I believe they should and I'll explain why.  In order to scan the component for sub-components, there needs to be some sort of property we can check against to validate this _is_ a component.  I think `displayName` should be that property.  I originally was thinking `propTypes`, but there's no guarantee every component has `propTypes` (especially sub-components).

devtools/buildIndexFiles.js
```diff
@@ -13,11 +13,15 @@ const isComponentFile = (source) => {
     return !ignoredFiles.some(ignored => source.includes(ignored));
 };
 
+const isComponentExported = (source) => {
+    return !source.match(/^\_/);
+};
+
```
greg-a-smith: Internal components are named with a leading underscore (`_`) so we should skip over those when building the index files.

src/_playground/documentation/Properties/Properties.js
```diff
@@ -16,13 +16,27 @@ export const Properties = ({ sourceModulePath }) => {
             <h2>Properties</h2>
             {componentNames.map((name, index) => {
                 const component = sourceModule[name];
+                const subcomponentNames = Object.keys(component)
+                    .sort()
+                    .filter(subName => component[subName] && component[subName].displayName);
```
greg-a-smith: Use `displayName` as our identifier that this key is, in fact, a sub-component.

src/_playground/documentation/Properties/Properties.js
```diff
@@ -16,13 +16,27 @@ export const Properties = ({ sourceModulePath }) => {
...
+                            const subcomponent = component[subName];
+                            return (
+                                <PropertyTable
+                                    defaultProps={subcomponent.defaultProps}
+                                    key={subIndex}
+                                    propDescriptions={subcomponent.propDescriptions}
+                                    propTypes={subcomponent.propTypes}
+                                    title={`${name}.${subName}`} />
+                            );
+                        })}
```
greg-a-smith: Added this additional loop to include all sub-components as well.  This still works if the component doesn't have any sub-components.

greg-a-smith:
@cjskillingstad I have added contributions to fix two issues:
1. The script that builds the `index.js` files was including internal components
1. The `Properties` component was no longer displaying props for sub-components

PR #426, opened 2019-02-21
fix: CORUI-6264: One component per file InputGroup

cjskillingstad:
### Description
* Separated all components in the `InputGroup` directory into their own files.
* Separate test files. Added snapshot.
* Maintained all git history.


greg-a-smith:
Not really part of this story, but why is `FormGroup` in the `InputGroup` directory?  This should be moved to `Forms`.  I created issue #427 to track this.

Issue #58, opened 2018-11-13
Add keyboard support


InnaAtanasova:
Users should be able to navigate and interact with the components using the keyboard.

droshev:
@InnaAtanasova @CodesOfRa shall we create a ticket per component and write the keys we have to support?

CodesOfRa:
Yes,  think it's better that way.

PR #431, opened 2019-02-22
feat: Add keyboard shortcut support

reenan:
### Description

Initial commit in order to fulfill #58

### DIFF
--- a/src/ListGroup/ListGroup.js
+++ b/src/ListGroup/ListGroup.js
@@ -1,24 +1,114 @@
 import classnames from 'classnames';
+import { findDOMNode } from 'react-dom';
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { Component } from 'react';
 
-export const ListGroup = ({ children, className, ...props }) => {
-    const listGroupClasses = classnames(
-        'fd-list-group',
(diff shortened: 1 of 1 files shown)


reenan:
Issues:
- Inner nested action handlers (buttons, flags, etc.) / how to manage this tab-chain?
- Custom handlers for keyboard events provided by outer component / should be possible to add those?

Questions:
- Should use a decorator to add handlers and manage focus state?
- Should implement full tab-chain handler? (how?)

bcullman:
@reenan - are you trying to address issues seen here:

https://sap.github.io/fundamental-react/listGroup

It's pretty clear that as they are currently written, the `simple list` and `list with checkboxes` are not tab-able.

bcullman:
The keyboard interaction model you are trying to implement here is based on Fiori 2/ui5.

Spoke with a member of the  SAP Design Decision Board about this. Essentially, for Fiori 3 has yet to make a decision on what keyboard interaction model will look like. We would prefer to wait until that is finalized before building custom handling. 

In the meantime, as I mentioned, there are some basic issues that are causing list items to not be tab-able. Opened #458 to address that.

PR #432, opened 2019-02-22
fix: CORUI-6264: One component per file ListGroup

cjskillingstad:
### Description

* Separated all components in the `ListGroup` directory into their own files.
* Maintained all git history.

`ListGroupItem`, `ListGroupItemActions` and `ListGroupItemCheckbox` have been moved to subcomponents of `ListGroup`.


src/index.js
```diff
@@ -28,7 +28,8 @@ export { Identifier } from './Identifier/Identifier';
 export { Image } from './Image/Image';
 export { InlineHelp } from './InlineHelp/InlineHelp';
 export { InputGroup, FormGroup } from './InputGroup/InputGroup';
-export { ListGroup, ListGroupItem, ListGroupItemActions, ListGroupItemCheckbox } from './ListGroup/ListGroup';
+export { default as ListGroup } from './ListGroup/ListGroup';
+export { default as ListGroupItem } from './ListGroup/ListGroupItem';
```
greg-a-smith: I guess I would have expected `ListGroupItem` to be a static on `ListGroup` because you'd never use it outside of the context of a list group.  I know it maybe makes it more complicated since `ListGroupItem` also has two statics of its own, but maybe they all move up to the `ListGroup` level.  Maybe something like:
```
ListGroup.Item = ListGroupItem
ListGroup.ItemActions = ListGroupItemActions;
ListGroup.ItemCheckbox = ListGroupItemCheckbox;
```

cjskillingstad: Moved to subcomponents of `ListGroup` in 8fbbb07

PR #434, opened 2019-02-22
fix: CORUI-6264: One component per file Menu

cjskillingstad:
### Description
* Separated all components in the `Menu` directory into their own files.
* Maintained all git history.


src/index.js
```diff
@@ -31,7 +31,10 @@ export { default as InputGroup } from './InputGroup/InputGroup';
 export { default as FormGroup } from './InputGroup/FormGroup';
 export { ListGroup, ListGroupItem, ListGroupItemActions, ListGroupItemCheckbox } from './ListGroup/ListGroup';
 export { LocalizationEditor } from './LocalizationEditor/LocalizationEditor';
-export { Menu, MenuList, MenuItem, MenuGroup } from './Menu/Menu';
+export { default as Menu } from './Menu/Menu';
+export { default as MenuList } from './Menu/MenuList';
+export { default as MenuItem } from './Menu/MenuItem';
+export { default as MenuGroup } from './Menu/MenuGroup';
```
greg-a-smith: I would have expected to see all these as static references on `Menu` and then only `Menu` would be exported.  Something like:
```js
Menu.Item = MenuItem;
Menu.Group = MenuGroup;
Menu.List = MenuList;
```

cjskillingstad: I agree, should've caught that. Moved to subcomponents in 8127fcc

Issue #436, opened 2019-02-24
update README.md with the example of importing fiori fundamentals


droshev:
### Description
fundamental-react needs to support IE11. `README.md` advises the user to include the file `"./node_modules/fiori-fundamentals/dist/fiori-fundamentals.min.css"`, which doesn't support IE11

### Versions
**fundamental-react:** latest
**fiori-fundamentals:** latest

---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

PR #439, opened 2019-02-25
chore: bump fundamentals version and reference ie 11 compat css

mobot11:
### Description
Bump fundamentals dependency to version with IE 11 css and reference IE 11 css distribution.

fixes https://github.com/SAP/fundamental-react/issues/436

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ For an existing react application, follow the steps below:
 1. Load the fiori-fundamentals styles. If using create-react-app, this will be in `App.css`.
 
     ```scss
-    @import '~fiori-fundamentals/dist/fiori-fundamentals-ie11.min.css';
+    @import '~node_modules/fiori-fundamentals/dist/fiori-fundamentals-ie11.min.css';
     ```
 
 1. Import components as needed. See [Component Documentation](https://sap.github.io/fundamental-react/) for examples and API details.
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ For an existing react application, follow the steps below:
 1. Load the fiori-fundamentals styles. If using create-react-app, this will be in `App.css`.
 
     ```scss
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -34,7 +34,7 @@ For an existing react application, follow the steps below:
 1. Load the fiori-fundamentals styles. If using create-react-app, this will be in `App.css`.
 
     ```scss
-    @import '~fiori-fundamentals/dist/fiori-fundamentals.min.css';
+    @import '~node_modules/fiori-fundamentals/dist/fiori-fundamentals-ie11.min.css';
```
greg-a-smith: I thought the tilde (`~`) character already resolved to the project's `node_modules` directory.

mobot11: https://stackoverflow.com/questions/39535760/what-does-a-tilde-in-a-css-url-do Looks like you are right. Updating.

README.md
```diff
@@ -34,7 +34,7 @@ For an existing react application, follow the steps below:
 1. Load the fiori-fundamentals styles. If using create-react-app, this will be in `App.css`.
 
     ```scss
-    @import '~fiori-fundamentals/dist/fiori-fundamentals.min.css';
+    @import '~fiori-fundamentals/dist/fiori-fundamentals-ie11.min.css';
```
greg-a-smith: This is probably a Fiori Fundamentals question, but why does the 1.4.3 distribution package have both the `-ie11` CSS file and the non-`-ie11` CSS file?  If it needs to support IE 11, why isn't _that_ just the only CSS file distributed?

mobot11: I wondered the same thing, see https://github.com/SAP/fundamental/issues/1244#issuecomment-459779074

bcullman:
this is not my preferred way to support IE, but I am having trouble locating automated tooling to generate an IE-only CSS override file.

PR #442, opened 2019-02-25
fix: One component per file for SideNav

greg-a-smith:
### Description
Splitting all the `SideNav` subcomponents into their own files.


src/SideNavigation/SideNavigation.js
```diff
@@ -1,8 +1,10 @@
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
+import SideNavList from './_SideNavList';
+import SideNavListItem from './_SideNavListItem';
 import React, { Component } from 'react';
 
-export class SideNav extends Component {
+class SideNav extends Component {
```
cjskillingstad: I wonder if we should have the naming be consistent between the file name `SideNavigation` and the component name `SideNav`

greg-a-smith: I wondered about it too.  I actually like the `SideNav` name better (probably mostly because it's shorter).  I could see leaving the directory name `SideNavigation`, but alter the file names to be just `SideNav` to match the component.

greg-a-smith:
@cjskillingstad Did some renaming of the source files to match the component names.

Issue #429, opened 2019-02-22
Split unit tests fo components in separate tests for each component


greg-a-smith:
### Description
With the refactoring of components into separate files, there should be a `*.test.js` file for each as well.  Created this as a separate story since it will likely involve writing or rewriting a number of tests to actually have them test individual components.

Components:
- ActionBar
- Button
- Forms
- ListGroup
- Panel
- SideNav
- Tile
- Time
- TimePicker
- TreeView

### Versions
**fundamental-react:** Any
**fiori-fundamentals:** Any

jbadan:
When addressing this issue, be sure to save git history: https://github.com/SAP/fundamental-react/wiki/One-Component-Per-File#how-to-preserve-git-history

jbadan:
In order for history to be preserved, we will need to turn off squash and merge. We should test this in a test repo, like deployment-lab, to make sure this doesn't increment the rc version for every commit as it's not squashed. I _think_ travis will take care of only incrementing once based on cancelling the other builds.

PR #443, opened 2019-02-25
fix: one component per file - Badge edition

jbadan:
### Description

Splitting all the status indicators into own files, a future story should rename the Badge directory Status Indicators


jbadan:
I didn't realize that #429 existed, so this pr also separated the unit tests. Let me know if it needs to be in separate prs :)

greg-a-smith:
@jbadan No problem to leave the tests split out if they have been.  That follow-up issue/story was to catch the ones that were a bit more complicated (and to not bog down the main purpose of this story).

PR #446, opened 2019-02-27
chore: Move FormGroup to forms directory

### LINKED ISSUE #427: Move FormGroup to Forms directory
### Description
The `FormGroup` component is currently in the `InputGroup` directory and thus shows up in the docs under the "Input Group" nav link, which makes no sense.  Sure, input groups can be used inside form groups, but they are not specifically related.  The `FormGroup` component should be in the `Forms` directory.

### Versions
**fundamental-react:** Any
**fiori-fundamentals:** Any

skvale:
### Description
Move `FormGroup` and `FormGroup.test` from the `InputGroup` directory to the `Forms` directory

fixes #427


skvale:
It was only two files.  I used this `gitcp` script https://github.com/cjskillingstad/scripts/blob/master/gitcp

jbadan:
We might want to consider moving over one or so of the examples of FormGroup from the InputGroup component page for the docs. The examples are intertwined, so it might be more straightforward to create a new example on the Forms page. Thoughts?

greg-a-smith:
@jbadan Agreed.  Although I'm sure we could just construct a new example of `FormGroup` using other `Form*` components.

greg-a-smith:
Although this is a good change, there is a major refactoring effort in #452 that should get merged first.  That PR will make this change simpler when it comes to examples on the docs pages.  I'm also not quite sure about using the `gitcp` script for retaining git history.  I think we should use https://gist.github.com/emiller/6769886 instead since we're not extracting multiple components from a single file.  Closing this PR for now.

skvale:
Sounds good, I was curious about two things:
- What makes a `FormSet` different from a `FormGroup` and how to call that out in the examples?
- Why not move the `InputGroup` to forms folder as well?

greg-a-smith:
@skvale I think your question about the difference between `FormSet` and `FormGroup` is an excellent question and one that I don't believe I can answer.  I will add this to our list of topics to discuss with the Fiori Fundamentals team (https://github.com/SAP/fundamental-react/wiki/Topics-for-Fiori-Fundamentals).  Even if they are different, it seems like they could be collapsed into a single component with props to handle the differences, but maybe I'm missing something.

As for moving `InputGroup` as well ... hmmm.  Interesting thought.  It sort of _seems_ like it can stand on its own, but then again, it does seem related to gathering form input.

In the future, I can see separate component pages for each `Form*` component with a "Patterns" page explaining the relationship between them and best practices for use.  I guess at that point, it really becomes a balance between the directory structure in `src` vs. the design and hierarchy of the left nav.  For that reason, I'd suggest we just leave it for now and revisit it going forward.

PR #447, opened 2019-02-27
docs: Add toggle right to left control to playground

### LINKED ISSUE #444: Docs: add RTL toggle for examples
### Description

See fundamental-ngx: https://sap.github.io/fundamental-ngx/action-bar

also see: https://github.com/SAP/fundamental-ngx/pull/511

### Versions
**fiori-fundamentals:**

---

_**DISCLAIMER:**

skvale:
### Description
screen shot 2019-02-27 at 5 28 01 am

screen shot 2019-02-27 at 5 28 07 am

fixes #444

### DIFF
--- a/src/_playground/documentation/DocsTile/DocsTile.js
+++ b/src/_playground/documentation/DocsTile/DocsTile.js
@@ -44,10 +44,10 @@ export class DocsTile extends React.Component {
                 <div className='frDocs-tile__features'>
                     <Toggle
                         className='frDocs-tile__feature'
-                        inputProps={{ 'aria-label': 'Show right to Left' }}
+                        inputProps={{ 'aria-label': 'Show right to left' }}
                         onChange={this.toggleRTL}
                         size='xs'>
-                        Show right to Left
+                        Show right to left
                     </Toggle>
                     <Toggle
                         className='frDocs-tile__feature'
--- a/src/_playground/documentation/DocsTile/DocsTile.js
+++ b/src/_playground/documentation/DocsTile/DocsTile.js
@@ -2,51 +2,62 @@ import { Button } from '../../../';
 import classnames from 'classnames';
 import { googlecode } from 'react-syntax-highlighter/styles/hljs';
 import PropTypes from 'prop-types';
+import React from 'react';
 import SyntaxHighlighter from 'react-syntax-highlighter';
 import { Toggle } from '../../../Toggle/Toggle';
-import React, { Component } from 'react';
 
-export class DocsTile extends Component {
+export class DocsTile extends React.Component {
     constructor(props) {
         super(props);
         this.state = {
-            backgroundToggle: false
+            backgroundToggle: false,
+            rightToLeft: false
         };
     }
 
     handleToggle = () => {
-        this.setState({
-            backgroundToggle: !this.state.backgroundToggle
-        });
+        this.setState(prevState => ({
+            backgroundToggle: !prevState.backgroundToggle
+        }));
+    };
+
+    handleToggleRightToLeft = () => {
+        this.setState(prevState => ({
+            rightToLeft: !prevState.rightToLeft
+        }));
     };
 
     render() {
         const { centered, children } = this.props;
+        const { backgroundToggle, rightToLeft } = this.state;
 
-        const outerDivClasses = classnames(
-            'frDocs-Content__tile',
-            {
-                'frDocs-Content__tile-background': !this.state.backgroundToggle
-            }
-        );
+        const outerDivClasses = classnames('frDocs-Content__tile', {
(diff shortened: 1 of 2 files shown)


src/_playground/documentation/DocsTile/DocsTile.js
```diff
@@ -2,51 +2,62 @@ import { Button } from '../../../';
...
     constructor(props) {
         super(props);
         this.state = {
-            backgroundToggle: false
+            backgroundToggle: false,
+            rightToLeft: false
         };
     }
 
     handleToggle = () => {
```
bcullman: should probably be renamed.

skvale: good call e0cb5c4

src/_playground/documentation/DocsTile/DocsTile.js
```diff
@@ -2,51 +2,62 @@ import { Button } from '../../../';
...
+                        size='xs'>
+                        Toggle right to left
+                    </Toggle>
+                    <Toggle
+                        className='frDocs-tile__feature'
+                        inputProps={{ 'aria-label': 'Toggle background color' }}
+                        onChange={this.handleBackgroundToggle}
+                        size='xs'>
+                        Toggle background
+                    </Toggle>
```
greg-a-smith: This is probably unfair to put this on you, but we should adjust the verbiage on both of these toggles.  A toggle is an on/off switch so the label should reflect that.  The label "Toggle ..." doesn't really state what the "on" or the "off" state will do.  In fact, the background one is actually probably backwards.  The "off" state shows a background and "on" does not.  I think the default state of both toggles should be "off".  That said, the default for "right to left" is LTR and the default for "background" is to have a background color.

How about these for the labels:
* Show RTL
* Hide background

That way, both can be "off" by default and then it's clear _what_ should happen when the toggle is "on".

**NOTE:** The `aria-label` values should be updated to match.

src/_playground/documentation/DocsTile/DocsTile.js
```diff
@@ -2,51 +2,62 @@ import { Button } from '../../../';
...
-import React, { Component } from 'react';
 
-export class DocsTile extends Component {
+export class DocsTile extends React.Component {
     constructor(props) {
         super(props);
         this.state = {
-            backgroundToggle: false
+            backgroundOn: true,
+            rightToLeft: false
```
greg-a-smith: Per comment below, maybe change these state variables to `hideBackground` and `showRTL` so both can be defaulted to `false` (matching the visual "off" state of the toggle).

src/_playground/documentation/DocsTile/DocsTile.js
skvale: I agree.  
I think the label `Show right to left` would be an improvement to `Show RTL` though

src/_playground/documentation/DocsTile/DocsTile.js
```diff
@@ -2,51 +2,62 @@ import { Button } from '../../../';
...
-                    size='xs'>Toggle background</Toggle>
-                <div className={innerDivClasses}>
-                    {children}
+                <div className='frDocs-tile__features'>
+                    <Toggle
+                        className='frDocs-tile__feature'
+                        inputProps={{ 'aria-label': 'Show right to Left' }}
+                        onChange={this.toggleRTL}
+                        size='xs'>
+                        Show right to Left
```
jbadan: Can we fix the capitalization on "Left" then I think this is good to merge!

src/_playground/documentation/DocsTile/DocsTile.js
```diff
@@ -2,51 +2,62 @@ import { Button } from '../../../';
...
-                    className='frDocs-tile__background-toggle'
-                    inputProps={{ 'aria-label': 'Toggle background color' }}
-                    onChange={this.handleToggle}
-                    size='xs'>Toggle background</Toggle>
-                <div className={innerDivClasses}>
-                    {children}
+                <div className='frDocs-tile__features'>
+                    <Toggle
+                        className='frDocs-tile__feature'
+                        inputProps={{ 'aria-label': 'Show right to Left' }}
```
greg-a-smith: One minor nit and then I'm on board.  The "L" in "Left" should be lowercase.

src/_playground/documentation/DocsTile/DocsTile.js
greg-a-smith: One minor nit and then I'm on board.  The "L" in "Left" should be lowercase.

greg-a-smith:
@skvale I changed this to be prefixed with `docs` rather than `feat`.  Although it is a feature for the documentation pages, nothing functional changed for the distribution package so it doesn't seem worthy of a minor version bump.

PR #448, opened 2019-02-27
docs: responsiveness

### LINKED ISSUE #302: Improve Docs: Make site fully responsive
### Description

Follow design seen here:

CORUI-6165

leannerivera:
### Description

site now changes dynamically with screen size, including sidebar toggle for smaller screens

fixes #302


src/_playground/documentation/Playground/Playground.js
```diff
@@ -223,7 +223,7 @@ export class Playground extends Component {
               case 'boolean':
                   return (
                       <div className='form-group-sublevel' key={item.attribute}>
-                          <div>Properties</div>
+                          <div className='test'>Properties</div>
```
bcullman: accidental checkin?

src/_playground/App.scss
```diff
@@ -13,8 +14,8 @@ $library-color: #61dafb;
         justify-content: space-between;
         height: 100vh;
         overflow: hidden;
-        margin-left: auto;
-        margin-right: auto;
+        // margin-left: auto;
+        // margin-right: auto;
         width: 100%;
```
bcullman: delete, don't comment out.

src/Button/Button.Component.js
```diff
@@ -142,7 +142,7 @@ content like appearing in a table or list. There may be more than one on the pag
 * **Negative Button**: Used for negative actions such as decline, cancel, no.
 `;
     return (
-        <div>
+        <div className='playground-wrap'>
```
bcullman: there probably is a better BEM-like naming convention for this.

src/_playground/documentation/Playground/Playground.js
leannerivera: sorry, yes, fixing now!

.vscode/settings.json
```diff
@@ -0,0 +1,2 @@
+{
+}
```
greg-a-smith: This file probably shouldn't be committed.

src/_playground/App.scss
```diff
@@ -1,3 +1,4 @@
+$mediaMaxWidth: 700px;
```
greg-a-smith: I think this value should be altered a bit.  The SAP Concur HIG uses` 1280px` as the breakpoint for collapsing the left navigation.  I think if you go all the way down to `700px`, the content area is going to get really compressed.  Besides, we also have a story coming down the pipeline to add a table of contents which will likely end up as a right column (once again similar to the SAP Concur HIG).

src/_playground/App.scss
```diff
@@ -96,12 +95,28 @@ $library-color: #61dafb;
             }
         }
 
+        @media (max-width: $mediaMaxWidth) {
+            .frDocs-Sidebar.sidebar-hidden{
```
greg-a-smith: We should probably get stylelint setup on this repo, but there should be a space before the `{`.

src/_playground/App.scss
```diff
@@ -96,12 +95,28 @@ $library-color: #61dafb;
...
+            }
+        }
+
         .frDocs-Content {
             flex-grow: 1;
             overflow-y: scroll;
             background-color: white;
             padding: 2rem;
 
+            .sidebar-toggle{
```
greg-a-smith: We should probably get stylelint setup on this repo, but there should be a space before the `{`.

src/_playground/App.scss
```diff
@@ -96,12 +95,28 @@ $library-color: #61dafb;
             }
         }
 
+        @media (max-width: $mediaMaxWidth) {
+            .frDocs-Sidebar.sidebar-hidden{
+                        margin-left: -20rem;
```
greg-a-smith: Are we sure that's wide enough?  The sidebar has its width specified in pixels so this should probably also be in pixels.

.vscode/settings.json
jbadan: did you mean to check this file in?

src/_playground/App.scss
```diff
@@ -96,12 +95,28 @@ $library-color: #61dafb;
...
+        }
+
         .frDocs-Content {
             flex-grow: 1;
             overflow-y: scroll;
             background-color: white;
             padding: 2rem;
 
+            .sidebar-toggle{
+                visibility: hidden;
```
greg-a-smith: I think this should toggle between `display: none` and `display: block` (not `visibility`).

jbadan:
This is looking really great! 

I was thinking in this sort of interaction, should the side nav auto hide once you've navigated to a new page without having to click the hamburger menu again? That's how we do it on the HIG, but am totally open to other opinions.

src/_playground/Routes.js
```diff
@@ -285,7 +285,8 @@ export class Routes extends Component {
         super(props);
         this.state = {
             query: '',
-            filteredItems: routes
+            filteredItems: routes,
+            showSideNav: false //for css media queries, default is invisible-'false'
```
greg-a-smith: How is this state variable staying in sync with the CSS media queries?  I only see one place where this value gets set and it's in the `onClick` handler of the toggle button.

leannerivera: it adds a class to line 537, frDocsContent depending on state

leannerivera:
@jbadan yeah that thought totally slipped my mind will work on that today.

src/_playground/Routes.js
```diff
@@ -345,7 +354,8 @@ export class Routes extends Component {
                             {navItems}
                         </nav>
                     </div>
-                    <div className='frDocs-Content'>
+                    <div className={`frDocs-Content ${showNav}`}>
+                        <button className='fd-button sap-icon--menu2 fd-button--standard fd-button--light sidebar-toggle' onClick={this.toggleNavVis.bind(this)} />
```
jbadan: I think this button will need an `aria-label` and `aria-expanded` true/false toggle on it as well for a11y reasons.

src/_playground/Routes.js
```diff
@@ -299,6 +300,12 @@ export class Routes extends Component {
         });
     };
 
+    toggleNavVis() {
```
jbadan: If you convert this to an arrow function, you won't need to bind `this` inside your `onClick`.

src/_playground/App.scss
leannerivera: Okay I will change it. I believe angular had a max of 680px or something like that and I was modeling them.

.vscode/settings.json
leannerivera: i dont' even know how that showed up 😖

src/_playground/Routes.js
```diff
@@ -327,10 +341,12 @@ export class Routes extends Component {
             );
         });
 
+        const showNav = this.state.showSideNav ? '' : 'sidebar-hidden';
+
         return (
             <BrowserRouter basename={process.env.PUBLIC_URL}>
                 <div className='frDocs-Container'>
-                    <div className='frDocs-Sidebar'>
+                    <div className={`frDocs-Sidebar ${showNav}`}>
```
greg-a-smith: A cleaner way of implementing conditional CSS classes is to use the `classnames` package.  You can check a number of other components for usage examples.

src/_playground/Routes.js
```diff
@@ -345,7 +361,8 @@ export class Routes extends Component {
                             {navItems}
                         </nav>
                     </div>
-                    <div className='frDocs-Content'>
+                    <div className={`frDocs-Content ${showNav}`}>
```
greg-a-smith: A cleaner way of implementing conditional CSS classes is to use the `classnames` package.  You can check a number of other components for usage examples.

src/_playground/Routes.js
```diff
@@ -327,10 +341,12 @@ export class Routes extends Component {
             );
         });
 
+        const showNav = this.state.showSideNav ? '' : 'sidebar-hidden';
```
greg-a-smith: This `sidebar-hidden` class name is potentially misleading.  Since the CSS media queries and the state in this component aren't always in sync, the `sidebar-hidden` class actually appears when the side bar _isn't_ hidden.  I get what you're going for here since you need a class for the purposes of a selector, but maybe a different name.  Maybe something like `sidebar-hidable`?  This would indicate it can be hidden if the CSS media query dictates.

src/_playground/Routes.js
leannerivera: great, i'll take a look 👍

src/_playground/Routes.js
leannerivera: i've changed it to `sidebar-visibility`, does that work?

src/_playground/Routes.js
```diff
@@ -318,6 +332,7 @@ export class Routes extends Component {
                                 activeClassName='frDocs-Nav__item--active'
                                 className='frDocs-Nav__item'
                                 key={route.url}
+                                onClick={this.resetNavState}
```
jbadan: Do you think it would work to just reuse `toggleNavVis` here?

src/_playground/Routes.js
```diff
@@ -327,10 +342,24 @@ export class Routes extends Component {
             );
         });
 
+        const sideBarClasses = classnames(
+            'frDocs-Sidebar',
+            {
+                'sidebar-visibility': !this.state.showSideNav
```
jbadan: Should we follow the same naming convention the docs have in general? Something like `frDocs-Sidebar__visibility`?

jbadan:
I'm running into some horizontal scrolling issues on the content side of things -

src/_playground/Routes.js
```diff
@@ -327,10 +342,20 @@ export class Routes extends Component {
             );
         });
 
+        const sideBarClasses = classnames({
+            'frDocs-Sidebar': this.state.showNav,
```
greg-a-smith: The base class `frDocs-Sidebar` shouldn't be conditional otherwise you can lose all styling on that section.

src/_playground/Routes.js
```diff
@@ -327,10 +342,20 @@ export class Routes extends Component {
...
+            'frDocs-Sidebar__visibility': !this.state.showSideNav
+        });
+
+        const docsContentClasses = classnames({
+            'frDocs-Content': this.state.showNav,
```
greg-a-smith: The base class `frDocs-Content` shouldn't be conditional otherwise you can lose all styling on that section.

src/_playground/App.scss
```diff
@@ -96,12 +95,28 @@ $library-color: #61dafb;
             }
         }
 
+        @media (max-width: $mediaMaxWidth) {
+            .frDocs-Sidebar__visibility {
```
greg-a-smith: This class name isn't technically correct for BEM.  This would mean `visibility` is the element, whereas it should be modifier.  That said, I think the name itself is misleading.  The class name should be representative of _what_ it's being used for.  This class is something that gets added to two elements so you can make use of them as a CSS selector for the media queries.  Maybe `frDocs-Sidebar--isHidable`?

src/_playground/Routes.js
```diff
@@ -345,7 +370,12 @@ export class Routes extends Component {
                             {navItems}
                         </nav>
                     </div>
-                    <div className='frDocs-Content'>
+                    <div className={docsContentClasses}>
+                        <button
```
greg-a-smith: Should we be using the fundamental-react `Button` component here rather than a plain `<button>`.  I think you'd get a lot of the className assistance for free (or at least with props).

src/_playground/Routes.js
```diff
@@ -327,10 +343,20 @@ export class Routes extends Component {
             );
         });
 
+        const sideBarClasses = classnames({
+            'frDocs-Sidebar': true,
+            'frDocs-Sidebar--isHidable': !this.state.showSideNav
+        });
```
greg-a-smith: So although this works, the better practice for non-conditional classes is to just include them as comma-separated args outside of the object.  Example:
```js
const sideBarClasses = classnames('frDocs-Sidebar', {
    'frDocs-Sidebar--isHidable': !this.state.showSideNav
});
```
So this syntax would always include `frDocs-Sidebar` and only include `frDocs-Sidebar--isHidable` if the condition was true.

**NOTE:** This would apply to the `docsContentClasses ` block below too.

src/_playground/Routes.js
```diff
@@ -327,10 +343,20 @@ export class Routes extends Component {
             );
         });
 
+        const sideBarClasses = classnames({
+            'frDocs-Sidebar': true,
```
jbadan: If these are always true, you can just list them like this: https://github.com/SAP/fundamental-react/blob/master/src/Alert/Alert.js#L34

src/_playground/Routes.js
```diff
@@ -345,7 +371,14 @@ export class Routes extends Component {
                             {navItems}
                         </nav>
                     </div>
-                    <div className='frDocs-Content'>
+                    <div className={docsContentClasses}>
+                        <Button
+                            aria-expanded={this.state.showSideNav}
+                            aria-label='Toggle Navigation'
+                            className='fd-button sap-icon--menu2 fd-button--standard fd-button--light sidebar-toggle'
```
jbadan: We can remove all the classNames that aren't unique to this Button (ie keep sidebar-toggle). The component should handle adding all the other classes for you.

greg-a-smith: You shouldn't need all these.  This should only be for custom classes that `Button` doesn't already handle.  I _think_ the only one you need here is `sidebar-toggle`.

.vscode/settings.json
bcullman: add `./vscode` to 

https://github.com/SAP/fundamental-react/blob/master/.gitignore

PR #451, opened 2019-02-28
fix: Add react/no-multi-comp lint rule

cjskillingstad:
### Description
- Separated Documentation related components into their own files. (Maintained git history)
- Made all components default exports
- Added `react/no-multi-comp` lint rule


greg-a-smith:
@mobot11 The `Import` component is still valid (for now, anyway).  That path to import from just points to the `index.js` file in the directory and that file still produces named exports.
```js
export { default as ActionBar } from './ActionBar.js';
```
Although we _could_ alter the `index.js` files for directories with a single component (such as `ActionBar`) so it would have a default export, that same pattern won't work for a directory with multiple exported components (such as `Forms`).  We will have to find a way to flatten the `lib` directory in order to accomplish this, but that's part of another story.

jbadan:
@greg-a-smith @mobot11 When I linked the latest iteration to nui-widgets-lab and imported a Button, it only worked with: `import Button from 'fundamental-react/lib/Button`. Button's index file has Button and ButtonGroup so I think it's safe to take the brackets off the import statement with this pr.

greg-a-smith:
@jbadan Not that I disagree with you, but I don't understand _how_.  That `Button/index.js` file doesn't have a default export so how does a default import work?  Are you able to import `ButtonGroup` the same way?

jbadan:
I have styles disabled since this is from my work on CSS Modules, but there's this: 

screen shot 2019-03-01 at 9 51 47 am

from: 

```
import Button from 'fundamental-react/lib/Button/Button';
import ButtonGroup from 'fundamental-react/lib/Button/ButtonGroup';
import React from 'react';

const FundamentalReactPage = () => (
    <div>
        <h2>Fundamentals-React</h2>

        <h3>Test Button</h3>
        <Button>Action Button</Button>
        <Button type='standard'>Standard Button</Button>
        <Button type='positive'>Positive Button</Button>
        <Button type='medium'>Medium Button</Button>
        <Button type='negative'>Negative Button</Button>

        <h3>Button Group</h3>
        <ButtonGroup>
            <Button type='standard'>Standard Button</Button>
            <Button type='positive'>Positive Button</Button>
            <Button type='medium'>Medium Button</Button>
        </ButtonGroup>

    </div>
);

export default FundamentalReactPage;
```

jbadan:
Also, just to make sure I haven't done any weird changes to the one I have linked, but my `Button/index.js` looks like: 
```
export { default as Button } from './Button.js';
export { default as ButtonGroup } from './ButtonGroup.js';
```

jbadan:
OH DUH! Ignore me - I'm going one level too deep!!! It does still need to be `import { ButtonGroup } from 'fundamental-react/lib/Button';`

Issue #409, opened 2019-02-14
Docs are broken in IE 11


mobot11:
### Description
https://sap.github.io/fundamental-react/home doesn't load in IE 11. Looks like Object.Assign isn't being polyfilled correctly.

![ie-11-docs](https://user-images.githubusercontent.com/9958488/52810017-74fb9700-3046-11e9-8c11-8b793bea5670.png)

### Versions
**fundamental-react:**

---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

jbadan:
fixed by https://github.com/SAP/fundamental-react/pull/742

PR #452, opened 2019-03-01
docs: Site layout, Page templates, TOC

### LINKED ISSUE #292: Improve Docs: Create Table of Contents (TOC) component
### Description
The documentation pages (`*.Component.js` files) should generate a "(Table of) Contents" so users can easily navigate the page.

As an aside to this work, the "Properties" should move to the bottom of the page so visual examples show up first.

CORUI-6170

greg-a-smith:
### Description
Created documentation page templates `ComponentPage` and `MarkdownPage` to control the layout of pages.  Updated all the `*.Component.js` files to use the new page templates.  Modified the overall page layout to include a top menu bar.

Added a TOC generator (using `tocbot`) and made it responsive.

Updated responsiveness for side nav (moved the toggle button up to the top menu).

**NOTE:** There are a few follow-up items that will need to be addressed:
1. The `react-element-to-jsx-string` package that builds the JSX strings for the code samples apparently uses `displayName` so our subcomponents may need to have their display names changed to a `Parent.Sub` format.
1. The `App.scss` file is using a number of Fiori-defined colors, but not using CSS variables yet.  It should use the variables, but that is also going to require making it IE11 compatible.  That work should be done as part of #409.
1. The TOC on a `MarkdownPage` template (_e.g._ "Home" page) shows up above the `<h1>` since the `<h1>` is in the markdown file.  It would be nice to find a cleaner way to implement that.  It _looks_ okay on larger screens because it shows up in its fixed right position.

fixes #292


src/_playground/App.scss
```diff
@@ -1,243 +1,303 @@
-$mediaMaxWidth: 1280px;
+$breakpoint-lg: 1280px;
```
greg-a-smith: Sorry for the massive diff in here.  I flattened the class structure which is why there are so many diffs.

bcullman:
Seeing some weirdness. When selecting a TOC item at the bottom of the list, and then navigating to another component page, the page does not return to the top.

bcullman:
also, when selecting an item from the toc list, the title that the page jumps is getting a focus ring..see:

image

greg-a-smith:
@bcullman Pushed changes for both issues raised.

bcullman:
Looks good. We could arguably clean up the TOC a *bit* more on the markdown page(s) with a bit of CSS, but I would not restructure them just so the TOC falls below the `<H1>` the fact it is not under the H1 comes into play on small screens:

image

with a bit of CSS, it could look cleaner in this resolution:

image

greg-a-smith:
@bcullman I actually found a way to leverage the heading renderer that we were already using to skip over H1 elements when the `title` prop is passed.  I think this might be a better solution yet.

bcullman:
@greg-a-smith that is absolutely better than what I proposed. 

looks fantastic. 🚢

PR #453, opened 2019-03-01
fix: forminput state proptype is now oneOf, not string

### LINKED ISSUE #428: Change FormInput state prop to oneOf validator
### Description
The `state` prop for `FormInput` has a list of valid values, but the prop is defined as a string.  This should be converted to a `PropTypes.oneOf` validator.

### Versions
**fundamental-react:** Any
**fiori-fundamentals:** Any

leannerivera:
### Description

The state prop for FormInput has a list of valid values, but the prop now uses PropTypes.oneOf validator rather than .string

fixes #428

### DIFF
--- a/src/Forms/FormInput.js
+++ b/src/Forms/FormInput.js
@@ -1,4 +1,5 @@
 import classnames from 'classnames';
+import { INPUT_TYPES } from '../utils/constants';
 import PropTypes from 'prop-types';
 import React from 'react';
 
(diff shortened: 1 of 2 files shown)


greg-a-smith:
@leannerivera, @mobot11 This PR should have also updated the `FormInput.propDescriptions` for `state`.  The description no longer needs all the valid values listed -- those are now called out in the "TYPE" column automatically because of the `oneOf` prop type validator.

PR #454, opened 2019-03-04
docs: ignoring snap files for ghpages build

jbadan:
### Description

Add snap files to ignore glob in webpack prod config for github pages. This will prevent new files from being uploaded, but still need to be manually deleted from gh-pages branch.

New build folder output:
screen shot 2019-03-04 at 3 44 04 pm

fixes #issueid

### DIFF
--- a/config/webpack.config.prod.js
+++ b/config/webpack.config.prod.js
@@ -415,7 +415,7 @@ module.exports = {
                         // it's runtime that would otherwise be processed through "file" loader.
                         // Also exclude `html` and `json` extensions so they get processed
                         // by webpacks internal loaders.
-                        exclude: [/\.(js|mjs|jsx|ts|tsx)$/, /\.html$/, /\.json$/, /\.snap$/],
+                        exclude: [/\.(js|mjs|jsx|ts|tsx|snap)$/, /\.html$/, /\.json$/],
                         options: {
                             name: 'static/media/[name].[hash:8].[ext]'
                         }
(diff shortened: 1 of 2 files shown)


config/webpack.config.prod.js
```diff
@@ -415,7 +415,7 @@ module.exports = {
                         // it's runtime that would otherwise be processed through "file" loader.
                         // Also exclude `html` and `json` extensions so they get processed
                         // by webpacks internal loaders.
-                        exclude: [/\.(js|mjs|jsx|ts|tsx)$/, /\.html$/, /\.json$/],
+                        exclude: [/\.(js|mjs|jsx|ts|tsx|snap)$/, /\.html$/, /\.json$/],
```
greg-a-smith: I know you didn't make this change, but is there any reason why `.html` and `.json` extensions are listed as separate regex strings?  Couldn't the regex just add `|snap|html|json` to the end?

bcullman: I think we just inherited this line from here:

https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/config/webpack.config.js#L507. It's strangely broken out separately there as well.

greg-a-smith: Yeah, was just more curious than anything. 😄

greg-a-smith: Should we make the same change in `webpack.config.dev.js`?  Probably not as big of a deal, but just asking.

jbadan: Only prod is used for making the docs site, so I think it should be fine, but we might as well?

jbadan: I've moved them around and gotten the same results, but honestly I wasn't sure the reasoning for them being separated so didn't want to change it. I think we do need a follow up story to remove all the unnecessary stuff from the webpack configs. They're really unwieldy and have slowed down my progress on the CSS modules story a lot.

PR #455, opened 2019-03-05
fix: Move FormGroup to Forms directory

### LINKED ISSUE #427: Move FormGroup to Forms directory

greg-a-smith:
### Description
The `FormGroup` component is not specific to input groups and thus should just be in the `src/Forms` directory with the other `Form*` components.

fixes #427


greg-a-smith:
Closing PR as the file diffs were not correct.

PR #463, opened 2019-03-07
fix: update FormInput state prop description

### LINKED ISSUE #428: Change FormInput state prop to oneOf validator

jbadan:
### Description

State prop description update was missed in https://github.com/SAP/fundamental-react/pull/453

These are now pulled from the Proptypes OneOf list.

Removed disabled, readOnly and help from state prop constant.

`readOnly` and `disabled` did not work as described (`state='disabled' or state='readonly'`) and instead are standalone attributes that work as expected when added like: `<FormInput disabled />` or `<FormInput readOnly />`. The examples have been updated as well as the prop list. 

`state='help'` also did not exist, although they have an example using a help input here: https://sap.github.io/fundamental/components/form.html#input-states, these are the same thing: https://sap.github.io/fundamental/components/form.html#input-help-elements, which we have also here: https://sap.github.io/fundamental-react/forms#input-help-elements

One snapshot had to be updated because it used the now defunct state of help.

screen shot 2019-03-06 at 6 57 46 pm
screen shot 2019-03-06 at 6 58 08 pm
screen shot 2019-03-06 at 6 58 17 pm

fixes #428

### DIFF
--- a/src/Forms/FormInput.js
+++ b/src/Forms/FormInput.js
@@ -29,7 +29,7 @@ FormInput.propTypes = {
 };
 
 FormInput.propDescriptions = {
-    state: 'Sets the state of the input.'
+    state: 'Sets the state of the input. Can be left empty for default styles.'
 };
 
 export default FormInput;
--- a/src/Forms/FormInput.js
+++ b/src/Forms/FormInput.js
@@ -23,6 +23,8 @@ FormInput.displayName = 'FormInput';
 
 FormInput.propTypes = {
     className: PropTypes.string,
+    disabled: PropTypes.bool,
+    readOnly: PropTypes.bool,
     state: PropTypes.oneOf(INPUT_TYPES)
 };
 
--- a/src/Forms/FormInput.js
+++ b/src/Forms/FormInput.js
@@ -27,7 +27,7 @@ FormInput.propTypes = {
 };
 
 FormInput.propDescriptions = {
-    state: 'Sets the state of the input. Options include \'normal\', \'valid\', \'invalid\', \'warning\', \'help\', \'disabled\', and \'readonly\'. Leave empty for normal.'
+    state: 'Sets the state of the input.'
 };
 
 export default FormInput;
--- a/src/Forms/Forms.Component.js
+++ b/src/Forms/Forms.Component.js
@@ -92,7 +92,7 @@ export const FormsComponent = () => {
 
             <Example
                 description={`The state of the input field can reflect validity of the data entered, 
-                    whether the input data is editable or disabled.\n\n* **Default**: The field is 
+                    whether the input data is editable or disabled.\n\n* **Normal**: The field is 
                     editable but no validation has occurred. \n\n* **Valid**: The data format entered 
                     has been validated and it’s correct, such as an email address.\n\n* **Invalid**: The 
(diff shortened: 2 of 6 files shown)


src/Forms/__snapshots__/Forms.test.js.snap
```diff
@@ -33,7 +33,7 @@ exports[`<Forms /> create form item 1`] = `
       *
     </label>
     <input
-      className="fd-form__control is-help"
+      className="fd-form__control is-warning"
```
jbadan: just to clarify - `is-help` is not a fundamentals style if you check the markup here: https://sap.github.io/fundamental/components/form.html#input-help-elements

bcullman: nice catch, @jbadan

src/Forms/Forms.Component.js
```diff
@@ -92,21 +92,18 @@ export const FormsComponent = () => {
 
             <Example
                 description={`The state of the input field can reflect validity of the data entered, 
-                    whether the input data is editable or disabled.\n\n* **Default**: The field is 
+                    whether the input data is editable or disabled.\n\n* **Normal**: The field is
```
bcullman: This is still "default", isn't it?

src/Forms/Forms.Component.js
```diff
@@ -92,21 +92,18 @@ export const FormsComponent = () => {
...
-                    selection within the form.\n\n* **Read Only**: Used to display static information 
-                    in the context of a form.\n\nAlong with Invalid and Warning, error messages should 
+                    the user from moving forward.\n\nAlong with Invalid and Warning, error messages should 
                     be displayed below the field so the user can correct the error and move forward.`}
                 title='Input States'>
                 <div>
                     <FormSet>
                         <FormItem>
-                            <FormLabel htmlFor='OatmD552'>Default Input</FormLabel>
+                            <FormLabel htmlFor='OatmD552'>Normal Input</FormLabel>
```
bcullman: I think i see what you are trying to do here - "default" is simply an unexpressed type, How about "Input (no state)"?

jbadan: Yeah I really don't like calling it "normal", but we do have a "normal" state and fundamentals refers to defaults as "normal". We have this in a couple other places to, so I'd almost rather have a cleanup story to make it consistent throughout the components.

src/Forms/Forms.Component.js
jbadan: See comment below and https://sap.github.io/fundamental/components/form.html#input-states, although I'm totally on board with changing this, we should do it consistently.

src/Forms/FormInput.js
```diff
@@ -23,11 +23,13 @@ FormInput.displayName = 'FormInput';
 
 FormInput.propTypes = {
     className: PropTypes.string,
+    disabled: PropTypes.bool,
+    readOnly: PropTypes.bool,
```
greg-a-smith: The `readOnly` prop is missing a prop description.  The `disabled` prop is okay because there is a default set for it already.  The `readOnly` might actually be a good candidate for another default.

PR #467, opened 2019-03-11
feat: Use popper for popovers

### LINKED ISSUE #313: Edge detection for popover-style elements
### Description
Edge detection for popover-style elements

### Versions
**fundamental-react: 0.2.X**
**fiori-fundamentals: 1.4.1**

---
popovers (and other popup style components. should edge-detect and flip where there is no space.

[popperjs](https://popper.js.org/) / [popperjs github](https://github.com/FezVrasta/popper.js) provides this functionality, and has libraries to react, angular and vue. 

https://www.npmjs.com/package/react-popper
https://www.npmjs.com/package/ngx-popper
https://www.npmjs.com/package/vue-popperjs

SAP Concur has previously had good success with popper.js, and implemented within our pre-fundamentals components library. 

image

see also:
https://github.com/SAP/fundamental-vue/issues/134
https://github.com/SAP/fundamental-ngx/issues/410

greg-a-smith:
### Description
The `Popover` component has been refactored to use `react-popper`, which in turn uses `popper.js`, to handle positioning of the overlays and the arrows on them (if shown).  Additionally, the `Popover` component is used throughout a number of other components and examples such as:

* `ActionBar`
* `ComboboxInput`
* `ContextualMenu`
* `Dropdown`
* `LocalizationEditor`
* `Panel`
* `Shellbar`
* `Table`
* `Tile`
* `TimePicker`

In addition to positioning, it handles edge detection and out of boundaries scenarios.

fixes #313


__mocks__/popper.js.js
```diff
@@ -0,0 +1,12 @@
...
+export default class Popper {
+    static placements = PopperJs.placements;
+
+    constructor() {
+        return {
+            destroy: () => { },
+            scheduleUpdate: () => { }
+        };
+    }
+}
```
greg-a-smith: Used to suppress console warnings in the unit tests since popper.js is basically only written to work in a real DOM.

src/Icon/Icon.js
```diff
@@ -25,12 +24,10 @@ Icon.displayName = 'Icon';
 Icon.propTypes = {
     glyph: PropTypes.string.isRequired,
     className: PropTypes.string,
-    clickHandler: PropTypes.func,
     size: PropTypes.oneOf(ICON_SIZES)
 };
 
 Icon.propDescriptions = {
-    clickHandler: 'Callback function when user clicks on the component.',
```
greg-a-smith: Needed to standardize on using `onClick` so it gets wired up properly if an `Icon` is used as the `control` for a `Popover`.

src/Token/Token.js
```diff
@@ -23,12 +24,7 @@ Token.displayName = 'Token';
 
 Token.propTypes = {
     children: PropTypes.node,
-    className: PropTypes.string,
-    clickHandler: PropTypes.func
-};
-
-Token.propDescriptions = {
-    clickHandler: 'Callback function when user clicks on the component.'
```
greg-a-smith: Needed to standardize on using `onClick` so it gets wired up properly if a `Token` is used as the `control` for a `Popover`.

src/_playground/App.scss
```diff
@@ -571,3 +571,134 @@ a.toc-link {
         }
     }
 }
+
```
greg-a-smith: This is the block of styles that will eventually need to migrate to Fiori Fundamentals so all libraries can take advantage of them.

src/utils/modalManager.js
```diff
@@ -0,0 +1,3 @@
+import ModalManager from 'react-overlays/ModalManager';
+
+export let modalManager = new ModalManager({ hideSiblingNodes: false, handleContainerOverflow: false });
```
greg-a-smith: If there's a better way to manage this, I'm all ears.

src/_playground/App.scss
```diff
@@ -571,3 +571,134 @@ a.toc-link {
...
+    position: absolute;
+    white-space: nowrap;
+    z-index: 2;
+    border-radius: 4px;
+    background-color: var(--fd-color-background-2);
+    box-shadow: 0 5px 20px 0 #d9d9d9, 0 2px 8px 0 #eeeeef;
+    box-shadow: 0 5px 20px 0 var(--fd-color-neutral-3), 0 2px 8px 0 var(--fd-color-neutral-2);
+    opacity: 1;
+
+    &--no-arrow {
```
bcullman: In the past I have disliked this method of dynamically generating classnames, because it can be missed in future refactors. I see fiori fundamentals uses it quite a bit. 

Should we allow or disallow this practice?

greg-a-smith: Yeah, I could go either way.  I know it makes searching for exact matches harder, but if we don't do any of these, then it almost seems like we've lost one of the big advantages of sass.

greg-a-smith: I added a "Use of sass to build full class names" item to the [Topics for Fiori Fundamentals](https://github.com/SAP/fundamental-react/wiki/Topics-for-Fiori-Fundamentals}.

src/_playground/App.scss
```diff
@@ -571,3 +571,134 @@ a.toc-link {
         }
     }
 }
+
+
```
bcullman: we can use separate file here rather than using app.scss as a dumping ground.

seems like we should break this up. it also might allow use to better segment out the CSS that is intended to be temporary

greg-a-smith: Good point.  I pushed a change to do this.

src/_playground/_popper.scss
```diff
@@ -0,0 +1,128 @@
+$arrow-height: 8px;
```
bcullman: Thanks. 

we should be able to point fundamental-vue and furnamental-ngx to this same sass.

bcullman:
looks good. 🚢

src/_playground/_popper.scss
ChristianKienle: Found ya. :)

PR #470, opened 2019-03-13
docs: skip nav link to header

### LINKED ISSUE #459: Improve Docs: a11y: skip nav link
### Description

We should have a skip link like on HIG for keyboard users.

screen shot 2019-03-05 at 4 27 40 pm

### Versions
**fundamental-react:** 0.4.0
**fiori-fundamentals:** N/A

---

_**DISCLAIMER:**

hertweckhr1:
### Description

Fixes #459 Added hidden skip navigation link in header that allows the user to skip to the main content.

### DIFF
--- a/src/_playground/App.scss
+++ b/src/_playground/App.scss
@@ -84,27 +84,6 @@ $menu-spacing: 20px;
             height: 2rem;
         }
     }
-
-    &__screen-reader-only-focusable {
-        background-color: transparent;
-        color: #fff;
-        position: absolute;
-        width: 1px;
-        height: 1px;
-        margin: -1px;
-        padding: 0;
-        overflow: hidden;
-        border: 0;
-
-        &:focus {
-        position: static;
-        width: auto;
-        height: auto;
-        margin: 0;
-        overflow: visible;
-        clip: auto;
-        }
-    }
 }
 
 .frDocs-Sidebar {
@@ -519,6 +498,28 @@ a.toc-link {
 
 .node-name--H2 {
     font-size: 1.2rem;
+}
+
+.screen-reader-only--focusable {
+    background-color: transparent;
+    color: #fff;
+    position: absolute;
+    width: 1px;
+    height: 1px;
+    margin: -1px;
+    padding: 0;
+    overflow: hidden;
+    border: 0;
+
+    &:focus {
+        top: 0;
+        left: 0;
+        width: auto;
+        height: auto;
+        margin: 0;
+        overflow: visible;
+        clip: auto;
+    }
 }
 
 @media (min-width: $breakpoint-md) {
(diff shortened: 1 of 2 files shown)


src/_playground/App.scss
```diff
@@ -84,6 +84,27 @@ $menu-spacing: 20px;
...
+        border: 0;
+
+        &:focus {
+        position: static;
+        width: auto;
+        height: auto;
+        margin: 0;
+        overflow: visible;
+        clip: auto;
+        }
```
greg-a-smith: Minor linting nit: these properties (inside `&:focus`) should be indented one more level.

src/_playground/App.scss
```diff
@@ -84,6 +84,27 @@ $menu-spacing: 20px;
             height: 2rem;
         }
     }
+
+    &__screen-reader-only-focusable {
```
greg-a-smith: I don't think it's necessary to make this a child class name concatenation.  I think this class could simply stand on its own.  Also, to follow BEM, I would make this class name `screen-reader-only--focusable`.

src/_playground/Routes.js
```diff
@@ -356,6 +356,11 @@ export class Routes extends Component {
                                 onClick={this.toggleNavVis}
                                 option='light'
                                 type='standard' />
+                            <a
+                                className='frDocs-Menu__screen-reader-only-focusable'
+                                href='#frDocs-Content'>
+                                Skip Navigation Links
+                            </a>
```
greg-a-smith: I think this block should come before the toggle `Button` in case you're on a small screen.

src/_playground/App.scss
```diff
@@ -84,6 +84,27 @@ $menu-spacing: 20px;
...
+        position: absolute;
+        width: 1px;
+        height: 1px;
+        margin: -1px;
+        padding: 0;
+        overflow: hidden;
+        border: 0;
+
+        &:focus {
+        position: static;
```
greg-a-smith: I wouldn't have this flow with the other content if it takes focus.  I would set `position: absolute` and then set `top: 0` and `left: 0`.  Although it's already absolutely positioned so maybe remove `position` from this block and just set `top: 0` and `left: 0` at the class level.

src/_playground/App.scss
bcullman: good opportunity to init stylelint here... (but in a separate PR)

src/_playground/App.scss
hertweckhr1: @greg-a-smith just for clarification - you do not think screen-reader-only--focusable should be under frDocs-Menu?  just className: screen-reader-only--focusable?

greg-a-smith: I actually meant that `screen-reader-only--focusable` could live at the root level (without any parent classes) since there's nothing specific about its styles related to the menu.

PR #471, opened 2019-03-18
docs: Call out FormInput's props explicitly

### LINKED ISSUE #464: FormInput should have more props explicitly called out
### Description

FormInput should have more props explicitly listed for PropType validation like FormRadioItem

See: https://github.com/SAP/fundamental-react/blob/master/src/Forms/FormRadioItem.js#L35

### Versions
**fundamental-react:** 0.4.1
**fiori-fundamentals:** N/A

---

_**DISCLAIMER:**

truclyle:
### Description

Call out more of FormInput's props explicitly for PropType validation using FormRadioItem as a model.

fixes #464

### DIFF
--- a/src/Forms/FormInput.js
+++ b/src/Forms/FormInput.js
@@ -41,7 +41,10 @@ FormInput.propTypes = {
 };
 
 FormInput.propDescriptions = {
-    state: 'Sets the state of the input. Can be left empty for default styles.'
+    name: 'Value for the `name` attribute on the input.',
+    state: 'Sets the state of the input. Can be left empty for default styles.',
+    type: 'Value for the `type` attribute on the input.',
+    value: 'Value for the `value` attribute on the input.'
 };
 
 export default FormInput;
--- a/src/Forms/FormInput.js
+++ b/src/Forms/FormInput.js
@@ -3,7 +3,7 @@ import { INPUT_TYPES } from '../utils/constants';
 import PropTypes from 'prop-types';
 import React from 'react';
 
-const FormInput = ({ state, className, ...props }) => {
+const FormInput = ({ state, className, disabled, id, name, placeholder, readOnly, type, value, ...props }) => {
     const formInputClasses = classnames(
         'fd-form__control',
         {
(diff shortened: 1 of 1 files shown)


src/Forms/FormInput.js
```diff
@@ -24,8 +31,13 @@ FormInput.displayName = 'FormInput';
...
     disabled: PropTypes.bool,
+    id: PropTypes.string,
+    name: PropTypes.string,
+    placeholder: PropTypes.string,
     readOnly: PropTypes.bool,
-    state: PropTypes.oneOf(INPUT_TYPES)
+    state: PropTypes.oneOf(INPUT_TYPES),
+    type: PropTypes.string,
+    value: PropTypes.string
 };
```
bcullman: prop description should be provided for all non default props seen here:

https://github.com/SAP/fundamental-react/blob/master/src/_playground/documentation/Properties/defaults.js

src/Forms/FormInput.js
```diff
@@ -24,11 +31,18 @@ FormInput.displayName = 'FormInput';
...
 
 FormInput.propDescriptions = {
+    disabled: 'Set to **true** to mark component as disabled and make it non-interactive.',
+    readOnly: 'Set to **true** to mark component as readonly.',
```
bcullman: those are default props - their descriptions will be assigned automatically by https://github.com/SAP/fundamental-react/blob/master/src/_playground/documentation/Properties/defaults.js

You will need add descriptions here for `name`, `state`, and arguably `type` as the definition differs here.

truclyle: Ok, I want to make sure I understand you correctly.

The descriptions at [https://github.com/SAP/fundamental-react/blob/master/src/_playground/documentation/Properties/defaults.js](https://github.com/SAP/fundamental-react/blob/master/src/_playground/documentation/Properties/defaults.js) are automatically added.

What I need to provide are the descriptions for the props (`name`, `value`) that are not described in the above file as well as modify `type`'s description to reflect the FormInput component.

Issue #468, opened 2019-03-11
Docs: bug: logo does not route correctly


jbadan:
### Description

When clicking on the fundamental-react logo, you are routed to https://sap.github.io/?sort=newest&filter=all

![Screen Shot 2019-03-11 at 8 51 00 AM](https://user-images.githubusercontent.com/29607818/54137419-deb75880-43da-11e9-91c3-649fba07688e.png)

### Versions
**fundamental-react:** 0.4.1
**fiori-fundamentals:**

---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

PR #474, opened 2019-03-19
docs: update title and meta information on page change

### LINKED ISSUE #456: Improve Docs: a11y: title does not update on page change
### Description

The title and meta description do not change on each page. This means someone using a screen reader would not know when a page reloaded or be alerted on changing pages within our site.

Title should match h1 on page, more or less.

See point 2 here for a potential solution: https://medium.com/scribd-data-science-engineering/common-accessibility-pitfalls-for-single-page-apps-and-how-to-avoid-them-16ecb19dcb71

### Versions
**fiori-fundamentals:** n/a

---

_**DISCLAIMER:**

jbadan:
### Description
Using react-helmet-async, we can update the title and meta description in the head to be accurate on page change. I put the SetMeta component inside utils for now, but am open to moving it to it's own folder.

Also fixes the logo routing issue.

fixes #456 #468

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,6 +120,7 @@
     "react-dev-utils": "^6.1.1",
     "react-dom": "^16.6.3",
     "react-element-to-jsx-string": "^14.0.2",
+    "react-helmet-async": "^0.2.0",
     "react-markdown": "^4.0.6",
     "react-router": "^4.3.1",
     "react-router-dom": "^4.3.1",
--- a/src/_playground/Routes.js
+++ b/src/_playground/Routes.js
@@ -364,7 +364,7 @@ export class Routes extends Component {
                             <a
                                 aria-label='Home'
                                 className='frDocs-Menu__logo'
-                                href='/'>
+                                href='/home'>
                                 Fundamental <span className='frDocs-Menu__logo--library'>React</span>
                             </a>
                             <span className='frDocs-Menu__spacer' />
--- a/src/_playground/documentation/Home/Home.js
+++ b/src/_playground/documentation/Home/Home.js
@@ -5,6 +5,7 @@ import Readme from './README.md';
 export const Home = () => {
     return (
         <MarkdownPage
+            description='Fundamental React Homepage'
             sourceFile={Readme}
             title='Home' />
     );
--- a/src/_playground/documentation/Home/Home.js
+++ b/src/_playground/documentation/Home/Home.js
@@ -5,7 +5,6 @@ import Readme from './README.md';
 export const Home = () => {
     return (
         <MarkdownPage
-            description='Fundamental React Homepage'
             sourceFile={Readme}
             title='Home' />
     );
--- a/src/_playground/documentation/utils/SetMeta.js
+++ b/src/_playground/documentation/utils/SetMeta.js
@@ -4,7 +4,9 @@ import React from 'react';
 
 const SetMeta = ({ title, description }) => {
     return (
-        <Helmet>
+        <Helmet
+            defaultTitle='Fundamental React'
+            titleTemplate='Fundamental React - %s'>
             <title>{title}</title>
             <meta content={description} name='description' />
         </Helmet>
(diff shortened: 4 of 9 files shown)


package.json
```diff
@@ -120,6 +120,7 @@
     "react-dev-utils": "^6.1.1",
     "react-dom": "^16.6.3",
     "react-element-to-jsx-string": "^14.0.2",
+    "react-helmet-async": "^0.2.0",
```
jbadan: @greg-a-smith does this need to be in regular dependencies or is dev deps ok?

src/_playground/documentation/utils/SetMeta.js
```diff
@@ -0,0 +1,19 @@
...
+    return (
+        <Helmet>
+            <title>{title}</title>
+            <meta content={description} name='description' />
+        </Helmet>
+    );
+};
+
+SetMeta.propTypes = {
+    description: PropTypes.string.isRequired,
```
greg-a-smith: Should `description` be required?  Seems like it could be optional.  The `description` prop in `ComponentPage` and `MarkdownPage` is optional so there might be work to do in those components if this needs to remain required.

package.json
greg-a-smith: Seems like it could just live in dev dependencies.

src/_playground/documentation/Home/Home.js
```diff
@@ -5,6 +5,7 @@ import Readme from './README.md';
 export const Home = () => {
     return (
         <MarkdownPage
+            description='Fundamental React Homepage'
```
greg-a-smith: If `description` was marked as optional, this could be removed.

src/_playground/Routes.js
```diff
@@ -364,7 +364,7 @@ export class Routes extends Component {
                             <a
                                 aria-label='Home'
                                 className='frDocs-Menu__logo'
-                                href='/'>
+                                href='/home'>
```
greg-a-smith: Not the end of the world, but this would make two places that would need to change if the default landing page ever changed.  This is currently handled by the `Redirect` component below.

src/_playground/documentation/utils/SetMeta.js
```diff
@@ -0,0 +1,19 @@
+import Helmet from 'react-helmet-async';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const SetMeta = ({ title, description }) => {
+    return (
+        <Helmet>
+            <title>{title}</title>
```
greg-a-smith: Is it worth prefixing the title prop with "Fundamental React"?  So, if it's loading the Button page, it would say "Fundamental React - Button" instead of just "Button".

src/_playground/documentation/utils/SetMeta.js
```diff
@@ -0,0 +1,21 @@
...
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const SetMeta = ({ title, description }) => {
+    return (
+        <Helmet
+            defaultTitle='Fundamental React'
+            titleTemplate='Fundamental React - %s'>
+            <title>{title}</title>
+            <meta content={description} name='description' />
```
greg-a-smith: For pages without a description (_e.g._ Home), it is emitting:
```
<meta content="" name="description" data-rh="true">
```
Should we only output this if a description exists?

greg-a-smith: Also, this is outputting markdown source (if it was supplied), which might be okay and might be problematic.  See:
```
<meta content="The **Calendar** is commonly used as the contents of a **Popover** when composing a **Date Picker**.
                It is rarely used on its own as a standalone component." name="description" data-rh="true">
```
Is there a way we can quickly cleanse this output?

jbadan: For documentation, @greg-a-smith and I talked offline and decided to have a followup story to add a description meta tag to the head as it's not strictly necessary at the moment.

PR #476, opened 2019-03-25
fix:added CustomPropTypes file

hertweckhr1:
### Description
Added CustomPropTypes file in order to begin shaping localizedText PropTypes

This is prework needed for adding a localizedText prop, tracked here: https://github.com/SAP/fundamental-react/wiki/localizedText-shape-Prop-for-All-Components

fixes #issueid

### DIFF
--- a/src/_playground/documentation/utils/CustomPropTypes.js
+++ b/src/_playground/documentation/utils/CustomPropTypes.js
@@ -1,0 +1,68 @@
+const ANONYMOUS = '<<anonymous>>';
+
+/* eslint-disable no-console */
+
+const wrapValidator = (validator, typeName, typeChecker = null) => {
+    return Object.assign(validator.bind(), {
+        typeName,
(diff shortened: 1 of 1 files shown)


greg-a-smith:
@hertweckhr1 @jbadan This should have actually been merged into the existing `CustomPropTypes` in `src/utils/CustomPropTypes/CustomPropTypes.js`.  The functions `wrapValidator` and `createChainableTypeChecker` are identical so they don't need to be included again, but the `i18n` function should be added to this existing file.  It will then be exported along with the `range` function that's already there.  We should create a separate PR to remove the file from the `_playground` directory and make the changes to the existing file.

greg-a-smith:
@hertweckhr1 @jbadan Never mind.  I see these changes are included in #477.

jbadan:
@greg-a-smith sorry that was my bad! will be fixed in the next pr :)

PR #477, opened 2019-03-25
fix: Added propertyTable formatting for localizedText PropType

hertweckhr1:
### Description
Added property table formatting for localized text

This is prework needed for adding a localizedText prop, tracked here: https://github.com/SAP/fundamental-react/wiki/localizedText-shape-Prop-for-All-Components


src/Badge/Counter.js
```diff
@@ -22,9 +23,18 @@ Counter.displayName = 'Counter';
...
     className: PropTypes.string,
+    localizedText: CustomPropTypes.i18n({
+        ariaLabel: PropTypes.string
+    }),
     notification: PropTypes.bool
 };
 
+Counter.defaultProps = {
+    localizedText: {
+        ariaLabel: 'Unread count'
```
jeffredodd: I think for this label name `ariaLabel` is too generic and could be used for any `aria-label` on the page. How about something like `counterLabel`? Maybe thats not the right one either but it's a start int he right direction.

src/_playground/documentation/Properties/_PropertyShape.js
```diff
@@ -0,0 +1,59 @@
...
+                <PropertyDefault
+                    defaultValue={defaultProps && defaultProps[propName]}
+                    prop={propTypes[propName]} />
+            ]
+        });
+    });
+
+    return (
+        <div className='frDocs-Properties__shape'>
+            {title && <Heading level={4}>{title}</Heading>}
```
jeffredodd: Is there a way for us to change the title so that it says `Counter - Localized Text` instead of `Localized Text - Counter`. This is super nit picky so if anyone disagrees I am fine with this way. 😄 . I just like Counter first as its the parent item and localization is the loc text being a subset of that item.

src/_playground/App.scss
```diff
@@ -285,6 +285,10 @@ $menu-spacing: 20px;
     }
 }
 
+.frDocs-Properties__shape {
+    margin: 0 100px 50px 100px;
```
jeffredodd: This is a pretty dramatic and doesn't seem to work well at 320px (mobile view) for the docs site. Can we drop some of these down while still keeping the indentation. I think the idea of the indentation is great however because it keeps the content looking associated with the above property table but different enough that as these props grow it will stand out in the correct way.

src/Badge/Counter.js
hertweckhr1: Makes sense to me 👍

src/_playground/documentation/Properties/_PropertyShape.js
hertweckhr1: Makes sense to me 👍

src/_playground/App.scss
```diff
@@ -285,6 +285,10 @@ $menu-spacing: 20px;
     }
 }
 
+.frDocs-Properties__shape {
+    margin: 0 50px 30px 50px;
```
greg-a-smith: Minor nit:  this can be simplified slightly as `margin: 0 50px 30px;`

src/_playground/documentation/Properties/_PropertyTable.js
```diff
@@ -2,12 +2,15 @@ import { defaultPropDescriptions } from './defaults';
...
 import PropertyDescription from './_PropertyDescription';
+import PropertyShape from './_PropertyShape';
 import PropertyType from './_PropertyType';
 import PropTypes from 'prop-types';
 import React from 'react';
 import sortBy from 'sort-by';
 import Table from '../../../Table/Table';
 
+/*eslint-disable no-console*/
+
```
greg-a-smith: This was just for debugging purposes -- this should be taken back out.

PR #478, opened 2019-03-26
fix: Renamed popper CSS classes

greg-a-smith:
### Description
In preparation for the move of the popper.js CSS to the Fiori Fundamentals library, renamed the classes to retain "popover" as the block-level name and made "popper" an element.

_**NOTE:** A number of snapshot files changes because of the classname changes._

### DIFF
--- a/src/ComboboxInput/__snapshots__/ComboboxInput.test.js.snap
+++ b/src/ComboboxInput/__snapshots__/ComboboxInput.test.js.snap
@@ -7,7 +7,7 @@ exports[`<ComboboxInput /> create combobox input 1`] = `
   <div
     className="fd-popover"
   >
-    <span
+    <div
       onFocus={[Function]}
       onMouseDown={[Function]}
       onTouchStart={[Function]}
@@ -37,7 +37,7 @@ exports[`<ComboboxInput /> create combobox input 1`] = `
           </div>
         </div>
       </div>
-    </span>
+    </div>
   </div>
 </div>
 `;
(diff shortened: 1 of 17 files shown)


src/ComboboxInput/__snapshots__/ComboboxInput.test.js.snap
```diff
@@ -4,74 +4,82 @@ exports[`<ComboboxInput /> create combobox input 1`] = `
...
-  <span
-    onFocus={[Function]}
-    onMouseDown={[Function]}
-    onTouchStart={[Function]}
+  <div
+    className="fd-popover"
   >
-    <div
-      className="fd-popper__control fd-popover"
+    <span
```
bcullman: bad practice to put div's inside spans. can you change?

bcullman: i know this is in the snap shot, but it is generally not good practice to put `<div>`s inside of `<span>`s. 

This is coming from the clickoutside library. Is emiting `<div>` vs `<span>` configurable? If not, should it be?

greg-a-smith: It is currently not configurable, but I just created an (enhancement) issue in that repo to allow for that customization.

bcullman: that was... fast:

https://github.com/nanot1m/react-foco/pull/3

greg-a-smith:
@bcullman I have updated this PR with the updated version of `react-foco`, which gets rids of the `<span>` element containing `<div>` elements.

bcullman:
🚢

PR #484, opened 2019-03-28
fix: Added localized text shape to alert

hertweckhr1:
### Description
Current Component: Alert


src/Alert/Alert.js
```diff
@@ -80,9 +82,18 @@ Alert.propTypes = {
...
 
+Alert.defaultProps = {
+    localizedText: {
+        closeButton: 'Close'
+    }
+};
+
 Alert.propDescriptions = {
     dismissible: 'Set to **true** to show a dismiss button.',
     link: 'Value to be applied to the anchor\'s `href` attribute.',
```
bcullman: are localizedText magic props that don't need descriptions because they go to their own table?

PR #486, opened 2019-03-28
fix: added localized text shape to input group

hertweckhr1:
### Description
Current Component: Input Group

### DIFF
--- a/src/InputGroup/InputGroup.js
+++ b/src/InputGroup/InputGroup.js
@@ -104,12 +104,12 @@ class InputGroup extends Component {
                         <span className='fd-input-group__addon fd-input-group__addon--button fd-input-group__addon--after'>
                             <button
                                 {...numberUpButtonProps}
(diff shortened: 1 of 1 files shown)


src/InputGroup/InputGroup.js
```diff
@@ -276,7 +283,12 @@ InputGroup.propTypes = {
 InputGroup.defaultProps = {
     addonPos: 'after',
     inputType: 'text',
-    inputValue: ''
+    inputValue: '',
+    localizedText: {
+        clearButton: 'Clear',
+        stepDownButton: 'Step down',
+        stepUpButton: 'Step up'
```
bcullman: I wonder if this should simply be `up` and `down`

hertweckhr1: @bcullman are you thinking the actual aria-label that was there before should just be 'up'/'down' or stepUpButton/StepDownButton should be called up/down?

PR #490, opened 2019-03-28
fix: Added description to localized text property table

hertweckhr1:
### Description
Added Description column to localized text table

This is continued work needed for adding a localizedText prop, tracked here: https://github.com/SAP/fundamental-react/wiki/localizedText-shape-Prop-for-All-Components

### DIFF
--- a/src/_playground/documentation/Properties/_PropertyShape.js
+++ b/src/_playground/documentation/Properties/_PropertyShape.js
@@ -7,7 +7,7 @@ import React from 'react';
 import sortBy from 'sort-by';
 import Table from '../../../Table/Table';
 
-const PropertyShape = ({ title, propTypes, defaultProps, description }) => {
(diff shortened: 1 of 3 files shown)


src/_playground/documentation/Properties/_PropertyShape.js
```diff
@@ -27,6 +28,10 @@ const PropertyShape = ({ title, propTypes, defaultProps }) => {
                 <PropertyType prop={propTypes[propName]} />,
                 <PropertyDefault
                     defaultValue={defaultProps && defaultProps[propName]}
+                    prop={propTypes[propName]} />,
+                <PropertyDescription
+                    defaultValue={defaultProps && defaultProps[propName]}
+                    description={description[propName]}
```
greg-a-smith: Does `PropertyDescription` take all these props?

Issue #342, opened 2019-02-01
DatePicker Invalid Date Detection fails as of 02/01/2018


jeffredodd:
### Description
The travis environment rolled over to `02/01/18` today and one our our datepicker tests no longer is passing. 

See this build for more information. https://travis-ci.org/SAP/fundamental-react/builds/487238965#L510

Upon doing some digging it appears that `2/3/2019-1/31/2019` is detected as an invalid range but `2/4/2019-2/1/2019` is not. I'd guess that only months and maybe years are being compared but dates are not? 

### Versions
**fundamental-react:** 0.3.0-rc.11

jeffredodd:
This will be resolved by #337

PR #492, opened 2019-03-29
fix: update failing calendar tests to work on all dates

jbadan:
### Description
Like #329, and #342 ,  Travis builds are failing due to our unit tests for dates not being hardcoded.
It will fail throughout tonight (Travis is on German time and therefore already it is already 3/29) and tomorrow because 2/29 does not exist.

fixes #issueid

### DIFF
--- a/src/Calendar/Calendar.test.js
+++ b/src/Calendar/Calendar.test.js
@@ -226,7 +226,9 @@ describe('<Calendar />', () => {
 
     test('click previous button', () => {
         let wrapper = mount(defaultCalendar);
-        const currentDateDisplayed = Object.assign(new Date(), wrapper.state('currentDateDisplayed'));
+
+        let initialDate = new Date('3/28/2019');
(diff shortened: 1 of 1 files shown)


src/Calendar/Calendar.test.js
```diff
@@ -226,7 +226,9 @@ describe('<Calendar />', () => {
 
     test('click previous button', () => {
         let wrapper = mount(defaultCalendar);
-        const currentDateDisplayed = Object.assign(new Date(), wrapper.state('currentDateDisplayed'));
+
+        let initialDate = new Date('3/28/2019');
+        wrapper.setState({ currentDateDisplayed: initialDate });
```
bcullman: There is a datetime testing lesson to be learned from this here...

jbadan: If only we learned that in #329 and #342 😸

bcullman: and apparently, #493 as well.

PR #494, opened 2019-03-30
fix: added localized text shape to calendar

hertweckhr1:
### Description
Current Component: Calendar

* Needed to update propertydefault.js to include arrays
* Needed to update propertyTable.js to fix bug in propDescriptions

### DIFF
--- a/src/Calendar/Calendar.js
+++ b/src/Calendar/Calendar.js
@@ -111,7 +111,10 @@ class Calendar extends Component {
 
     changeMonth = (month) => {
         let date = this.state.currentDateDisplayed;
-        let months = this.props.localizedText.monthsOfYear;
+        const { monthJan, monthFeb, monthMar, monthApr, monthMay, monthJun, monthJul,
+            monthAug, monthSep, monthOct, monthNov, monthDec } = this.props.localizedText;
+        let months = [monthJan, monthFeb, monthMar, monthApr, monthMay, monthJun, monthJul,
+            monthAug, monthSep, monthOct, monthNov, monthDec];
         date.setMonth(months.indexOf(month));
         // reset date to first of month
         date.setDate(1);
@@ -157,7 +160,10 @@ class Calendar extends Component {
     }
 
     generateMonths = (monthProps) => {
-        let months = this.props.localizedText.monthsOfYear;
+        const { monthJan, monthFeb, monthMar, monthApr, monthMay, monthJun, monthJul,
+            monthAug, monthSep, monthOct, monthNov, monthDec } = this.props.localizedText;
+        let months = [monthJan, monthFeb, monthMar, monthApr, monthMay, monthJun, monthJul,
+            monthAug, monthSep, monthOct, monthNov, monthDec];
         let listOfMonths = months.map(element => {
             let shortenedNameMonth = '';
             if (element.length > 3) {
@@ -338,7 +344,8 @@ class Calendar extends Component {
     }
 
     disableWeekday = (date, weekDays) => {
-        let daysName = this.props.localizedText.daysOfWeek;
+        const { daySun, dayMon, dayTues, dayWed, dayThurs, dayFri, daySat } = this.props.localizedText;
+        let daysName = [daySun, dayMon, dayTues, dayWed, dayThurs, dayFri, daySat];
 
         if (typeof weekDays === 'undefined') {
             return false;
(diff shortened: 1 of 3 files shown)


src/Calendar/Calendar.js
```diff
@@ -564,6 +570,11 @@ Calendar.basePropTypes = {
 Calendar.propTypes = {
     ...Calendar.basePropTypes,
 
+    localizedText: CustomPropTypes.i18n({
+        daysOfWeek: PropTypes.arrayOf(PropTypes.string),
+        monthsOfYear: PropTypes.arrayOf(PropTypes.string),
+        singleCharDaysOfWeek: PropTypes.arrayOf(PropTypes.string)
+    }),
```
greg-a-smith: It _seems_ wrong to pass localized text as an array of strings.  I know a calendar is just a table with rows and columns, but we have gotten bitten before by just sending localized text strings in as arrays for API simplicity.  If the code within the component assembles the strings into arrays for construction purposes, that's fine, but I think localized text strings should come in as individual strings.

src/_playground/documentation/Properties/_PropertyTable.js
```diff
@@ -29,21 +29,22 @@ const PropertyTable = ({ title, propTypes, defaultProps, propDescriptions }) =>
...
+                    <PropertyType componentName={title} prop={propTypes[propName]} />,
+                    <PropertyDefault
+                        defaultValue={defaultProps && defaultProps[propName]}
+                        prop={propTypes[propName]} />,
+                    <PropertyDescription
+                        defaultValue={defaultProps && defaultProps[propName]}
+                        description={mergedPropDescriptions[propName]}
+                        prop={propTypes[propName]} />
+                ]
+            });
```
greg-a-smith: I see where you're going with this and on the surface, it's not a bad idea.  The problem, however, is there is no entry in the main table of props called `localizedText` anymore.  And there _should_ be since that will be necessary for consumers to properly construct their props.  You and I know it's called `localizedText`, but not all consumers will.  Also, what's to stop us from changing that top-level prop name since all it's really tied to is the `i18n` custom prop type validator.

Additionally, I think this same type of pattern will be extended to `shape` prop types.

src/Calendar/Calendar.js
```diff
@@ -110,7 +111,10 @@ class Calendar extends Component {
 
     changeMonth = (month) => {
         let date = this.state.currentDateDisplayed;
-        let months = ['January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'];
+        const { monthJan, monthFeb, monthMar, monthApr, monthMay, monthJun, monthJul,
+            monthAug, monthSep, monthOct, monthNov, monthDec } = this.props.localizedText;
+        let months = [monthJan, monthFeb, monthMar, monthApr, monthMay, monthJun, monthJul,
+            monthAug, monthSep, monthOct, monthNov, monthDec];
```
greg-a-smith: Couple things here:

1. For readability, I'd put each value from `localizedText` on its own line such as:
   ```
   const {
      monthJan,
      monthFeb,
      monthMar,
      ...
   } = this.props.localizedText;
   ```
2. Since this work is done in multiple places, let's make a function that returns the array of localized month names.  That function can do the extraction of values from the `localizedText` object and the array construction.

src/Calendar/Calendar.js
```diff
@@ -337,7 +344,8 @@ class Calendar extends Component {
     }
 
     disableWeekday = (date, weekDays) => {
-        let daysName = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'];
+        const { daySun, dayMon, dayTues, dayWed, dayThurs, dayFri, daySat } = this.props.localizedText;
+        let daysName = [daySun, dayMon, dayTues, dayWed, dayThurs, dayFri, daySat];
```
greg-a-smith: Same comments as with month names.

src/Calendar/Calendar.js
```diff
@@ -416,7 +427,8 @@ class Calendar extends Component {
 
     generateWeekdays = () => {
         let weekDays = [];
-        let daysName = ['S', 'M', 'T', 'W', 'T', 'F', 'S'];
+        const { dayCharSun, dayCharMon, dayCharTues, dayCharWed, dayCharThurs, dayCharFri, dayCharSat } = this.props.localizedText;
+        let daysName = [dayCharSun, dayCharMon, dayCharTues, dayCharWed, dayCharThurs, dayCharFri, dayCharSat];
```
greg-a-smith: Same comments as with month names.

greg-a-smith:
@hertweckhr1 On a separate note, I noticed the automatic sorting of prop names in the Localized Text table actually makes these harder to comprehend.  This is definitely a time when a logical order (rather than alphabetical) would be useful.  Without rearchitecting the property table components, what about naming the localized text keys like such:
```
month01Jan
month02Feb
day01Sun
day02Mon
dayChar01Sun
dayChar02Mon
```
This way, when sorted alphabetically, they appear in a logical order.

PR #495, opened 2019-04-01
fix: added localizedText prop to time

hertweckhr1:
### Description
Current Component: Time

### DIFF
--- a/src/Time/Time.js
+++ b/src/Time/Time.js
@@ -238,7 +238,7 @@ Time.propDescriptions = {
     hoursDownButtonProps: 'Additional props to be spread to the hours down `<button>` element.',
     hoursInputProps: 'Additional props to be spread to the hours `<input>` element.',
     hoursUpButtonProps: 'Additional props to be spread to the hours up `<button>` element.',
-    localizedText: {
+    localizedTextShape: {
         meridiemAM: 'Ante meridiem for 12 hour clock. ',
         meridiemPM: 'Post meridiem for 12 hour clock.'
     },
--- a/src/Time/Time.js
+++ b/src/Time/Time.js
@@ -123,7 +123,7 @@ class Time extends Component {
                         downButtonProps={minutesDownButtonProps}
                         format12Hours={format12Hours}
                         inputProps={minutesInputProps}
-                        localizedText={this.props.localizedText}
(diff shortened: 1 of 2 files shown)


src/Time/_TimeItem.js
```diff
@@ -250,6 +251,10 @@ TimeItem.propTypes = {
     downButtonProps: PropTypes.object,
     format12Hours: PropTypes.bool,
     inputProps: PropTypes.object,
+    localizedText: CustomPropTypes.i18n({
```
jbadan: Could the `Time` component pass these props to `TimeItem` so that consumers only have to pass the localized text once?

hertweckhr1: @greg-a-smith for suggested comment associated with similar suggestion in timepicker

greg-a-smith: To @jbadan's point, the `TimeItem` component is not exported for consumers to use.  It only exists as an internal component to better architect the code.  To that end, the `TimeItem` component does not need the same level of prop type validation that `Time` does.  You could simply make this:
```
localizedText: PropTypes.object.isRequired
```
And then remove it from `defaultProps` below.

src/Time/Time.js
```diff
@@ -163,7 +167,7 @@ class Time extends Component {
                         type={'Period'}
                         upButtonProps={meridiemUpButtonProps}
                         updateTime={this.updateTime}
-                        value={CLOCK[this.state.time.meridiem]} />
+                        value={this.CLOCK[this.state.time.meridiem]} />
```
greg-a-smith: You will need to pass `localizedText` into each `TimeItem` component call since the consumer doesn't have access to that internal component.

jbadan:
Sorry about all these netlify status checks!  They should go away, just ignore them 😅

src/Time/Time.js
```diff
@@ -156,14 +163,15 @@ class Time extends Component {
                         disabled={disabled}
                         downButtonProps={meridiemDownButtonProps}
                         inputProps={meridiemInputProps}
+                        localizedText={this.props.localizedText}
```
greg-a-smith: In these four instances, you can just use the variable `localizedText` that you pruned from `this.props` up on line 59.

PR #496, opened 2019-04-01
fix: add time picker localized text shape

hertweckhr1:
### Description
Current Component: Time Picker

### DIFF
--- a/src/TimePicker/TimePicker.js
+++ b/src/TimePicker/TimePicker.js
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import Time from '../Time/Time';
 import TimePickerItem from './_TimePickerItem';
 import React, { Component } from 'react';
-/*eslint-disable*/
+
 class TimePicker extends Component {
     constructor(props) {
         super(props);
--- a/src/TimePicker/TimePicker.js
+++ b/src/TimePicker/TimePicker.js
@@ -135,6 +135,7 @@ class TimePicker extends Component {
                                 format12Hours={format12Hours}
                                 id={id}
                                 inputProps={inputProps}
+                                localizedText={this.props.localizedText}
                                 onChange={this.onChange}
                                 placeholder={this.state.placeholder}
                                 showHour={showHour}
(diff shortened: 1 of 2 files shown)


src/TimePicker/_TimePickerItem.js
```diff
@@ -253,6 +253,10 @@ TimePickerItem.propTypes = {
     inputProps: PropTypes.object,
     isValid: PropTypes.bool,
     length: PropTypes.number,
+    localizedText: CustomPropTypes.i18n({
+        meridiemAM: PropTypes.string,
+        meridiemPM: PropTypes.string
```
jbadan: Same question as other pr on Time, could `TimePicker` pass these props to `_TimePickerItem` so that consumers only have to provide the localized text once?

hertweckhr1: @jbadan totally agree! since they were originally coming in as global variables, this is the way @greg-a-smith and I had talked about solving it for the meantime before we end up changing time and timepicker altogether with moment.js(or similar). @greg-a-smith do you think its possible to pass props only once as it is right now?

greg-a-smith: Yes.  See my comments in #495.  The same pattern will occur here.  The `TimePicker` component will use the `i18n` custom prop type validator and then `TimePickerItem` can just use the `object` prop type validator.

greg-a-smith: Yes.  See my comments in #495.  The same pattern will occur here.  The `TimePicker` component will use the `i18n` custom prop type validator and then `TimePickerItem` can just use the `object` prop type validator.

jbadan:
Sorry about all these netlify status checks!  They should go away, just ignore them 😅

src/TimePicker/TimePicker.js
```diff
@@ -178,6 +190,10 @@ TimePicker.defaultProps = {
 
 TimePicker.propDescriptions = {
     ...Time.propDescriptions,
+    localizedTextShape: {
+        meridiemAM: 'Ante meridiem for 12 hour clock.',
+        meridiemPM: 'Post meridiem for 12 hour clock.'
+    },
```
greg-a-smith: I _think_ this is redundant.  This likely already comes from the line above where we spread the `Time.propDescriptions`.  Just try removing this and see if the values still show up in the docs.

PR #497, opened 2019-04-02
fix: make all heading elements customizable

### LINKED ISSUE #249: Allow for customizable heading levels in components
Currently, a number of components that use heading tags (_e.g._ `<h1>`, `<h2>`, etc.) have them hard-coded to use `<h1>`.  There should only be one `<h1>` element per page.  These components should take a prop to use the proper heading element (`<h1>`, `<h2>`, `<h3>`, `<h4>`, `<h5>`, `<h6>`).

Components include:
* `ActionBar`
* `MegaMenu`
* `Menu`
* `Modal`
* `Panel`
* `SideNavigation`

jbadan:
### Description
`<h1>` headings were used in following components: 
    * ActionBar 
    * MegaMenu  (Deprecated - does not exist in FR)
    * Menu 
    * Modal 
    * Panel 
    * SideNavigation 
   * Tile
   * ProductTile

* Default header is now `<h3>` with the option to customize using `<h2>`-`<h6>`
* `<h1>` is not allowed due to a11y reasons (except in ActionBar)
*  updated unit tests for each component - I did not separate these into separate files as we have a story in the backlog to separate the testing files for each subcomponent
* update snapshots for each component 
* chose not to add additional examples as it didn't seem necessary with the props table explaining the options. Open for other opinions!

fixes #249 
Accompanied by changes in fundamental repo: https://github.com/SAP/fundamental/pull/1367

No visual changes in examples

### DIFF
--- a/src/ActionBar/ActionBar.test.js
+++ b/src/ActionBar/ActionBar.test.js
@@ -74,6 +74,20 @@ describe('<ActionBar />', () => {
         expect(tree).toMatchSnapshot();
     });
 
+    describe('ActionBar Header', () => {
+        test('should allow customization of header level', () => {
+            const element = mount(
+                <ActionBar.Header
+                    description=''
+                    level={2}
+                    title='' />
+            );
+            expect(
+                element.find('.fd-action-bar__title').type()
+            ).toBe('h2');
+        });
+    });
+
     describe('Prop spreading', () => {
         test('should allow props to be spread to the ActionBar component', () => {
             const element = mount(<ActionBar data-sample='Sample' />);
@@ -121,7 +135,7 @@ describe('<ActionBar />', () => {
             ).toBe('Sample');
         });
 
-        test('should allow props to be spread to the ActionBarHeader component\'s h1 element', () => {
+        test('should allow props to be spread to the ActionBarHeader component\'s heading element', () => {
             const element = mount(
                 <ActionBar.Header
                     description=''
@@ -130,7 +144,7 @@ describe('<ActionBar />', () => {
             );
 
             expect(
-                element.find('h1').getDOMNode().attributes['data-sample'].value
+                element.find('h3').getDOMNode().attributes['data-sample'].value
             ).toBe('Sample');
         });
 
--- a/src/ActionBar/ActionBar.test.js
+++ b/src/ActionBar/ActionBar.test.js
@@ -79,7 +79,7 @@ describe('<ActionBar />', () => {
             const element = mount(
                 <ActionBar.Header
                     description=''
-                    level={2}
+                    headingLevel={2}
                     title='' />
             );
             expect(
--- a/src/ActionBar/_ActionBarHeader.js
+++ b/src/ActionBar/_ActionBarHeader.js
@@ -32,7 +32,7 @@ ActionBarHeader.propTypes = {
     className: PropTypes.string,
     description: PropTypes.string,
     descriptionProps: PropTypes.object,
-    headingLevel: CustomPropTypes.range(2, 6),
+    headingLevel: CustomPropTypes.range(1, 6),
     titleProps: PropTypes.object
 };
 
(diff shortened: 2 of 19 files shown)


src/ActionBar/ActionBar.test.js
```diff
@@ -74,6 +74,20 @@ describe('<ActionBar />', () => {
         expect(tree).toMatchSnapshot();
     });
 
+    describe('ActionBar Header', () => {
+        test('should allow customization of header level', () => {
+            const element = mount(
+                <ActionBar.Header
+                    description=''
+                    level={2}
```
bcullman: for this, and all of them, I think we should use `headingLevel` `headLevel`, or something similar

src/_playground/documentation/Properties/defaults.js
```diff
@@ -8,6 +8,7 @@ export const defaultPropDescriptions = {
     id: 'Value for the `id` attribute on the element.',
     inputProps: 'Additional props to be spread to the `<input>` element.',
     labelProps: 'Additional props to be spread to the `<label>` element.',
+    level: 'Heading level. `<h1>` is not allowed for accessibility reasons.',
```
bcullman: this might be better...

`<H1>` is reserved for the page title. It should not appear in components.

..that said, seems like the actionbar *could* be used as a page title. Thoughts?

src/ActionBar/ActionBar.test.js
greg-a-smith: I tend to agree.  I would vote for `headingLevel`.  I think if this was a `Heading` component that only emitted the `<h#>` tag, you could make a strong case for just `level`, but since this becomes a passthrough prop, I like being more explicit.

src/ActionBar/_ActionBarHeader.js
```diff
@@ -29,9 +32,14 @@ ActionBarHeader.propTypes = {
     className: PropTypes.string,
     description: PropTypes.string,
     descriptionProps: PropTypes.object,
+    level: CustomPropTypes.range(2, 6),
```
greg-a-smith: See comment about making this `headingLevel`.

src/Menu/_MenuGroup.js
```diff
@@ -21,7 +24,12 @@ MenuGroup.displayName = 'Menu.Group';
 MenuGroup.propTypes = {
     title: PropTypes.string.isRequired,
     className: PropTypes.string,
+    level: CustomPropTypes.range(2, 6),
```
greg-a-smith: See comment about making this `headingLevel`.

src/Modal/Modal.js
```diff
@@ -106,6 +108,7 @@ Modal.propTypes = {
     contentProps: PropTypes.object,
     footerProps: PropTypes.object,
     headerProps: PropTypes.object,
+    level: CustomPropTypes.range(2, 6),
```
greg-a-smith: See comment about making this `headingLevel`.

src/Panel/_PanelHead.js
```diff
@@ -23,9 +26,14 @@ PanelHead.displayName = 'Panel.Head';
 PanelHead.propTypes = {
     className: PropTypes.string,
     description: PropTypes.string,
+    level: CustomPropTypes.range(2, 6),
```
greg-a-smith: See comment about making this `headingLevel`.

src/SideNavigation/_SideNavList.js
```diff
@@ -62,13 +65,18 @@ SideNavList.propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
     hasParent: PropTypes.bool,
+    level: CustomPropTypes.range(2, 6),
```
greg-a-smith: See comment about making this `headingLevel`.

src/_playground/documentation/Properties/defaults.js
greg-a-smith: See comment about making this `headingLevel`.

greg-a-smith: I wondered the same thing about `ActionBar`.  Not quite sure what the design team was envisioning for that component, but it does seem like it _could_ be used as a page title.

jbadan:
@greg-a-smith @bcullman Updated the pr to use `headingLevel` and changed the default text. I agree that ActionBar should be able to use `<h1>` tags. I've left the default for it to be `<h3>`, but it allows `<h1>` tags, unlike the other components.

jbadan:
@greg-a-smith Updated to include Tile and Product Tile components

PR #500, opened 2019-04-02
chore: changes needed for Netlify PR previews

jbadan:
### Description

In order for netlify to publish correctly, we need to remove the `/fundamental-react/` from the public path for it's build script. 

This PR was opened before the netlify config was setup so the bot will not comment on it. It will look like this in the future:
https://github.com/SAP/fundamental-react/pull/513 
> Note: This build failed because the script doing the building isn't on that branch. This is just to show the comment.

See this pr's build here: https://fundamental-react.netlify.com/home
Once this branch is merged, the above url will become a preview site for the master branch and will publish each time we merge to master. Each pr comment will look something like:
https://app.netlify.com/sites/fundamental-react/deploys/[HASH HERE]

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "scripts": {
     "build-css": "node-sass-chokidar --include-path ./src --include-path ./node_modules src/ -o src/",
     "build-doc": "node scripts/build.js && cp build/index.html build/404.html",
-    "build-doc:netlify": "NETLIFY=true npm run build-doc",
+    "build-doc:netlify": "npm run build && NETLIFY=true npm run build-doc",
     "build-js": "node scripts/build.js",
     "build:lint:fix": "npm run build:lint -- --fix",
     "build:lint": "eslint 'src/**' --ext .js,.jsx --env browser,node",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
   "scripts": {
     "build-css": "node-sass-chokidar --include-path ./src --include-path ./node_modules src/ -o src/",
     "build-doc": "node scripts/build.js && cp build/index.html build/404.html",
+    "build-doc:netlify": "npm run build && NETLIFY=true npm run build-doc",
```
greg-a-smith: Am I missing something here?  This new script has been added, but it doesn't appear to ever get called.  Is there some background black magic happening?

jbadan: @greg-a-smith Yeah this pr is a little confusing without context, sorry about that. I also did it a little out of order, so the comments won't show until the next pr. 

The magic is happening on the netlify dashboard. We were using a `netlify.toml` file in the fundamental repo as that overrides any settings in the dashboard, which we didn't have access to at the time. 

Screen Shot 2019-04-12 at 7 34 07 AM

greg-a-smith: Got it, thanks.  Although we don't _have to_, I'm curious if it's worth just using a `netlify.toml` file in each repo to reduce the "magic".

PR #503, opened 2019-04-04
docs: Property shapes

greg-a-smith:
### Description
Did some refactoring to make the `PropertyShape` component more generic so it will work with both `i18n` custom prop types _AND_ `shape` prop types.  Below are screenshots of both prop types.

Did some other minor cleanup:
- Changed some example titles/headings to prevent id collisions with the property tables
- Fixed an issue with `localizedText` for the `Counter` component
- Updated some prop descriptions where appropriate

Screen Shot 2019-04-03 at 11 57 43 PM

Screen Shot 2019-04-03 at 11 58 00 PM

### DIFF
--- a/src/Table/Table.js
+++ b/src/Table/Table.js
@@ -69,7 +69,7 @@ Table.propTypes = {
 };
 
 Table.propDescriptions = {
-    tableData: 'Array of objects that contain two properties: `rowData` (an array of strings containing data for each column in the row), and `children` (an array of objects containing additional rows).',
+    tableData: 'Array of objects that contain one property: `rowData` (an array of strings containing data for each column in the row).',
     headers: 'Array of localized text strings for the column headers.',
     tableBodyProps: 'Additional props to be spread to the `<tbody>` element.',
     tableBodyRowProps: 'Additional props to be spread to the `<tr>` elements within `<tbody>`. If using a function, the parameters passed will be an object representing the row (from `tableData`) and the row index.',
--- a/src/_playground/documentation/Properties/_PropertyDescription.js
+++ b/src/_playground/documentation/Properties/_PropertyDescription.js
@@ -1,31 +1,19 @@
-import PropertyTable from './_PropertyTable';
 import PropTypes from 'prop-types';
 import React from 'react';
 import ReactMarkdown from 'react-markdown';
 
-const PropertyDescription = ({ defaultValue, description, prop }) => {
-    const typeChecker = prop.typeChecker;
-
-    return (
-        <React.Fragment>
-            <div>
-                <ReactMarkdown source={description} />
-            </div>
-            {prop.typeName === 'shape' &&
-                <PropertyTable
-                    defaultProps={defaultValue}
-                    propTypes={typeChecker} />
-            }
-        </React.Fragment>
-    );
-};
+const PropertyDescription = ({ description }) => (
+    <React.Fragment>
+        <div>
+            <ReactMarkdown source={description} />
+        </div>
+    </React.Fragment>
+);
 
 PropertyDescription.displayName = 'PropertyDescription';
 
 PropertyDescription.propTypes = {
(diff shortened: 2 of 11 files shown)


src/Table/Table.js
```diff
@@ -69,7 +69,7 @@ Table.propTypes = {
 };
 
 Table.propDescriptions = {
-    tableData: 'Array of objects that contain two properties: `rowData` (an array of strings containing data for each column in the row), and `children` (an array of objects containing additional rows).',
+    tableData: 'Array of objects that contain one property: `rowData` (an array of strings containing data for each column in the row).',
```
greg-a-smith: Not quite sure where this previous description came from, but that is _NOT_ how the component works.

src/_playground/documentation/Properties/_PropertyDescription.js
```diff
@@ -1,31 +1,19 @@
...
-        </React.Fragment>
-    );
-};
+const PropertyDescription = ({ description }) => (
+    <React.Fragment>
+        <div>
+            <ReactMarkdown source={description} />
+        </div>
+    </React.Fragment>
+);
```
greg-a-smith: Moved display of shape table out of description column.

src/_playground/documentation/Properties/_PropertyShape.js
```diff
@@ -7,31 +8,36 @@ import React from 'react';
...
+const PropertyShape = ({ title, propTypes = {}, defaultProps = {}, propDescriptions = {} }) => {
     if (!propTypes) {
         return (
             <em>This shape has no defined properties.</em>
         );
     }
 
     let propInfo = Object.keys(propTypes).map((propName) => {
-        return { propName: propName, required: propTypes[propName].typeRequired };
+        return { propName: propName, required: !!propTypes[propName].typeRequired };
```
greg-a-smith: Force a boolean to prevent sorting issues (if `typeRequired` was undefined rather than false, props were sorting at the bottom for what appeared to be no reason).

src/_playground/documentation/Properties/_PropertyTable.js
```diff
@@ -9,15 +10,15 @@ import React from 'react';
...
+const PropertyTable = ({ title, propTypes = {}, defaultProps = {}, propDescriptions = {} }) => {
     if (!propTypes) {
         return (
             <em>This component has no defined properties.</em>
         );
     }
 
     let propInfo = Object.keys(propTypes).map(propName => {
-        return { propName: propName, required: propTypes[propName].typeRequired };
+        return { propName: propName, required: !!propTypes[propName].typeRequired };
```
greg-a-smith: Force a boolean to prevent sorting issues (if `typeRequired` was undefined rather than false, props were sorting at the bottom for what appeared to be no reason).

src/_playground/documentation/Properties/_PropertyType.js
```diff
@@ -57,9 +57,14 @@ const PropertyType = ({ prop, componentName }) => {
...
                 </React.Fragment>
             );
+        case 'shape':
         case 'i18n':
-            const shapeName = `${componentName} - Localized Text`;
-            return (<a href={`#${makeSafeId(shapeName)}`}>Localized Text</a>);
+            return (
+                <React.Fragment>
+                    <div>{prop.typeName}</div>
+                    {topLevel && <div>(<a href={`#${makeSafeId(getShapeTitle(componentName, propName, prop.typeName))}`}>details</a>)</div>}
```
greg-a-smith: For now anyway, let's only show "details" link for shapes that are top-level prop types.  It could get messy in a hurry if shapes are embedded within out prop types.

src/Table/Table.js
jbadan: Array of objects _contains_ right?

greg-a-smith: Actually, I think it is just _contain_ because "objects" is plural.

PR #504, opened 2019-04-04
fix: Popover popper styles consumed from fiori-fundamentals

greg-a-smith:
### Description
Removed popover popper styles and consumed them from fiori-fundamentals.


jbadan:
I don't think this is caused by this pr, but have we noticed this before:

greg-a-smith:
@jbadan Good catch.  No, it's not specifically from this PR, but once the popper moved to different DOM nodes, the z-index needs to come into play.  I will create a new PR in `fundamental` to address this and then bump the dependency version in this PR.

greg-a-smith:
@jbadan New PR in fundamental:  https://github.com/SAP/fundamental/pull/1374

PR #507, opened 2019-04-05
fix: Implement our SideNav Component and replace previous custom SideNav in docs

hertweckhr1:
### Description
Implemented our own SideNav Component by replacing previous custom built SideNav 

Things to consider/feedback:
* Is there anything additional we want to change for style or do we want to stay with how the component given in is styled?
* The background for the side nav is now white - the scroll bar creates a slight distinction between sidenav and main content, is this enough, or do we want a bigger dinstinction?


src/_playground/Routes.js
```diff
@@ -387,9 +387,9 @@ export class Routes extends Component {
                                     inputValue={this.state.query}
                                     onChange={this.onChangeHandler} />
                             </div>
-                            <nav className='frDocs-Nav'>
+                            <SideNav className='frDocs-Nav'selectedId=''>
```
greg-a-smith: I think the `selectedId` should always be set here.  Initially, it should be for the "Home" page, but then should get reset each time the page changes.

src/_playground/App.scss
```diff
@@ -95,6 +95,7 @@ $menu-spacing: 20px;
     background-color: #EDEFF0;
     display: flex;
     flex-direction: column;
+    overflow-y: auto;
```
hertweckhr1: By getting rid of previous custom CSS for frDocs-Nav, overflow was no longer an attribute for side-nav.  Larger question for Fiori Fundamentals - do we want to add an overflow option to side-nav to scroll through list? Right now, the list is not scrollable unless you put custom CSS in.

greg-a-smith: I think this got added to the wrong container class.  See the before and after screen captures below.  The scrolling region should only be the side navigation component and should not extend into the search field or the top header.

#### Before (Correct)
fd-react-side-nav-scrolling-before

#### After (Incorrect)
fd-react-side-nav-scrolling-after

src/_playground/Routes.js
```diff
@@ -387,9 +387,9 @@ export class Routes extends Component {
                                     inputValue={this.state.query}
                                     onChange={this.onChangeHandler} />
                             </div>
-                            <nav className='frDocs-Nav'>
+                            <SideNav selectedId='Home'>
```
greg-a-smith: I don't know if this should be hard-coded.  It seems as though there should be a different state variable to track this.  The state variable starts as `'Home'`, but should change each time a new item is selected.

That said, just looking at the `SideNav` component, it looks like there is no way to get that information out of the component to track it as a consumer.  The component takes `selectedId` as a prop and tracks its own state, but it would be very easy for `SideNav` and its consumer to get out of sync.  I think there should be an `onItemSelect` callback function prop on the `SideNav` component that a consumer could subscribe to.

I guess this was one of our goals -- eating our own dog food. 🐶 

We should create a new story to fix `SideNav` and then once that is complete, we can make additional adjustments to this story.

PR #511, opened 2019-04-09
fix: added onSelectItem callback to SideNav Component

### LINKED ISSUE #508: Add callback prop to SideNav
### Description
The `SideNav` component exposes a prop called `selectedId` and respects changes to it within its own state.  However, there is no prop for a consumer to use to be notified when that internal state changes.  There should be a prop (maybe `onItemSelect`?) that gets called whenever the value changes.

### Versions
**fundamental-react:** 0.5.1-rc.18
**fiori-fundamentals:** Any

CORUI-6474

hertweckhr1:
### Description
Added onSelectItem callback to SideNav Component for a consumer to use prop when SelectedId is changed. 

Relates to issue #507 

fixes #508

### DIFF
--- a/src/SideNavigation/SideNav.js
+++ b/src/SideNavigation/SideNav.js
@@ -69,6 +69,7 @@ SideNav.defaultProps = {
 
 SideNav.propDescriptions = {
     icons: 'Set to **true** to only render icons for each `SideNavListItem`.',
+    onItemSelect: 'Prop callback to notify of internal state changes',
     selectedId: 'The `id` of the selected `SideNavListItem`.'
 };
 
--- a/src/SideNavigation/SideNav.js
+++ b/src/SideNavigation/SideNav.js
@@ -69,7 +69,7 @@ SideNav.defaultProps = {
 
 SideNav.propDescriptions = {
     icons: 'Set to **true** to only render icons for each `SideNavListItem`.',
-    onItemSelect: 'Prop callback to notify of internal state changes',
+    onItemSelect: 'Callback function when a navigation item is selected. Arguments passed are the event and the id of the selected item.',
     selectedId: 'The `id` of the selected `SideNavListItem`.'
 };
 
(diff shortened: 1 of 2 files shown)


src/SideNavigation/SideNav.js
```diff
@@ -57,7 +59,12 @@ SideNav.propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
     icons: PropTypes.bool,
-    selectedId: PropTypes.string
+    selectedId: PropTypes.string,
+    onItemSelect: PropTypes.func
+};
+
+SideNav.defaultProps = {
+    onItemSelect: () => { }
```
greg-a-smith: Nice job adding the default empty function. 👍 This prop needs to be added to the `SideNav.propDescriptions` too.

src/SideNavigation/SideNav.test.js
```diff
@@ -374,4 +374,23 @@ describe('<SideNav />', () => {
...
+                    <SideNav.ListItem
+                        id='item-1'
+                        name='Link Item 1'
+                        url='#' />
+                </SideNav.List>
+            </SideNav>);
+
+            element.find('#item-1 a').simulate('click');
+
+            expect(f).toHaveBeenCalledTimes(1);
```
greg-a-smith: Although this test is good, I think it might be worth also adding a check to make sure the second argument is the proper `id` value (you can use `toHaveBeenCalledWith`).

src/SideNavigation/SideNav.js
```diff
@@ -57,11 +59,17 @@ SideNav.propTypes = {
...
 };
 
 SideNav.propDescriptions = {
     icons: 'Set to **true** to only render icons for each `SideNavListItem`.',
+    onItemSelect: 'Prop callback to notify of internal state changes',
```
jbadan: Can we include something here that tells you what parameters are given (event and id)? - see HIG callbacks: https://hig.concur.com/pages/develop/web/components/nui-widgets/v11/popover.html#callbacks

greg-a-smith: Agreed.  I would also make this a little more English-sounding (rather than purely technical).  Maybe something like "Callback function when a navigation item is selected. Arguments passed are the `event` and the `id` of the selected item."

PR #517, opened 2019-04-15
fix: Use React.Children.toArray to prune null children in TabGroup 

### LINKED ISSUE #514: Tabs component does not accept null as child
### Description

image
image

`Tabs` component doesn't accept null as child, leading to crash as shown in screenshots above.

### Versions
**fundamental-react:** 0.5
**fiori-fundamentals:** 1.5.3

---

_**DISCLAIMER:**

skvale:
### Description
```js
React.Children.map(this.props.children, (child, index)
```
maintains null children in the map function while
```js
React.Children.toArray(this.props.children).map((child, index) => {
```
removes null values

fixes #514

### DIFF
--- a/src/Tabs/TabGroup.js
+++ b/src/Tabs/TabGroup.js
@@ -57,7 +57,7 @@ class TabGroup extends Component {
             return (
                 <TabContent
                     {...child.props.tabContentProps}
-                    key={child.props.id || index}
+                    key={index}
                     selected={this.state.selectedIndex === index}>
                     {child.props.children}
                 </TabContent>);
--- a/src/Tabs/TabGroup.js
+++ b/src/Tabs/TabGroup.js
@@ -46,17 +46,18 @@ class TabGroup extends Component {
 
     // create tab list
     renderTabs = () => {
-        return React.Children.map(this.props.children, (child, index) => {
+        return React.Children.toArray(this.props.children).map((child, index) => {
             return this.cloneElement(child, index);
         });
     };
 
     // create content to show below tab list
     renderContent = () => {
-        return React.Children.map(this.props.children, (child, index) => {
+        return React.Children.toArray(this.props.children).map((child, index) => {
             return (
                 <TabContent
                     {...child.props.tabContentProps}
+                    key={child.props.id || index}
                     selected={this.state.selectedIndex === index}>
                     {child.props.children}
                 </TabContent>);
--- a/src/Tabs/TabGroup.test.js
+++ b/src/Tabs/TabGroup.test.js
@@ -89,4 +89,27 @@ describe('<Tabs />', () => {
             ).toBe('Sample');
         });
     });
+
+    describe('null tab children', () => {
+        test('should render all tabs that are not null', () => {
+            const tabsWithNull = (
+                <TabGroup>
+                    <Tab
+                        id='1'
+                        title='Tab 1'>
+                    Lorem ipsum dolor sit amet consectetur adipisicing elit. Dolore et ducimus veritatis officiis amet? Vitae officia optio dolor exercitationem incidunt magnam non, suscipit, illo quisquam numquam fugiat? Debitis, delectus sequi?
+                    </Tab>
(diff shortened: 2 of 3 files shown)


greg-a-smith:
@skvale I'm curious of the use of _why_ this would be needed.  It seems pretty legitimate to have the consumer own filtering out the child objects being passed.  Generally speaking, these components are supposed to be pretty "dumb" and should leave most of the control in the hands of the consumers.

skvale:
Good point.

Sounds like a good thing to ask the issue reporter @aerfio as well.

greg-a-smith:
Sorry @skvale.  I didn't even notice that this was addressing a reported issue.  I will comment in that issue back to the original author.

src/Tabs/TabGroup.test.js
```diff
@@ -89,4 +89,27 @@ describe('<Tabs />', () => {
...
+    describe('null tab children', () => {
+        test('should render all tabs that are not null', () => {
+            const tabsWithNull = (
+                <TabGroup>
+                    <Tab
+                        id='1'
+                        title='Tab 1'>
+                    Lorem ipsum dolor sit amet consectetur adipisicing elit. Dolore et ducimus veritatis officiis amet? Vitae officia optio dolor exercitationem incidunt magnam non, suscipit, illo quisquam numquam fugiat? Debitis, delectus sequi?
+                    </Tab>
+                    {null}
```
greg-a-smith: Minor nit here: how about writing this more like:
```js
const condition = false;

...

    {condition && <Tab id='2' title='Tab 2'>Lorem ipsum</Tab>}
```
This will ultimately pass in `null`, but it makes the test look more "real".

src/Tabs/TabGroup.js
```diff
@@ -46,17 +46,18 @@ class TabGroup extends Component {
...
     };
 
     // create content to show below tab list
     renderContent = () => {
-        return React.Children.map(this.props.children, (child, index) => {
+        return React.Children.toArray(this.props.children).map((child, index) => {
             return (
                 <TabContent
                     {...child.props.tabContentProps}
+                    key={child.props.id || index}
```
greg-a-smith: Maybe just `key={index}` since the `id` prop is optional on each tab?

greg-a-smith:
I also added issue #523 to track similar changes in other components.

src/Tabs/TabGroup.js
skvale: I think using `key={child.props.id || index}` prevents an important bug.

Suppose someone is rendering tabs like this:
```js
const {tabs} = this.state;
return (
  <TabGroup>
    {tabs.map(tab => <Tab title={tab.title} />)}
  </TabGroup>
)
```
If tabs from state were to be updated to a different array, React may not render the change correctly if it is using `index` as a key.  Two different tabs would have the same key.

Writing it like `key={child.props.id || index}` catches the typical use case of static tabs, but allows for dynamic tabs if an id is passed through props.

```js
const {tabs} = this.state;
return (
  <TabGroup>
    {tabs.map(tab => <Tab id={tab.id} title={tab.title} />)}
  </TabGroup>
)
```

Though, yeah, that is a separate issue.  The propTypes of `TabContent` should be updated to include id as well, if this is approved.

skvale: I'll try to create a broken example in the docs page to illustrate it.

skvale: welp, I couldn't create any issues with using the index, so disregard.

greg-a-smith:
@skvale I've approved the PR, but the license/cla check never reported finishing so I can't merge it yet.  Can you push an empty commit to your branch to force a rebuild to see if that clears it up?

jbadan:
@skvale I don't think it's you for the CLA - it's failing to report on the other open pr too. As well as a new pr in the fundamental repo :(

jbadan:
@skvale Something went wrong with our webhook. You will need to resign the CLA (again) and then it should work this time!!

PR #519, opened 2019-04-15
Fix: Add Callbacks to Alert and Modal 

hertweckhr1:
### Description
Added Callback and Test to Alert 
Added propDescriptions and propDefault to existing callback for Modal 

fixes #issueid


src/Modal/Modal.js
```diff
@@ -133,6 +135,7 @@ Modal.propDescriptions = {
     localizedTextShape: {
         closeButton: 'Aria-label for <button> element.'
     },
+    onClose: 'Callback function passing event when close button is clicked.',
```
jbadan: I know this already existed and we're just documenting it now, but should we have them all match and make it `onCloseClicked`

src/Alert/Alert.js
```diff
@@ -12,14 +12,18 @@ class Alert extends Component {
...
             isActive: false
+        }, () => {
+            this.props.onCloseClicked(e);
         });
     };
 
     render() {
         const {
+            onCloseClicked,
+            onClick,
```
jbadan: does `onClick` need to be here?

src/Modal/Modal.js
greg-a-smith: This has actually made this a breaking change.  I know we're still on version 0, so that's good, but there are now docs pages that don't work correctly (_e.g._ Modal, Popover) because although `onClose` was never a documented callback, it _was_ being supported internally with `Modal`.

PR #535, opened 2019-05-03
fix(Forms;SideNav;TreeView): Make React.Children null safe

### LINKED ISSUE #523: Make React.Children maps null-safe
### Description
As was found in #517, the following syntax:
```jsx
    {condition && <Component />}
```
will pass `null` in the `children` array when the condition is false.  It's unfortunate that it just simply isn't omitted (as the syntax would seem to indicate), but that's how it works.  #517 is addressing `TabGroup`, but there are other components that should have similar protection added to be safe against `null` children.  At a minimum, it looks like:

* `FormRadioGroup`
* `SideNav`
* `SideNavList`
* `SideNavListItem`
* `BaseTree`
* `TreeHead`
* `TreeItem`
* `TreeRow`
* `TreeView`

Those components are using `React.Children`, but there may be others doing a map on children a different way.

### Versions
**fundamental-react:** 0.5.1
**fiori-fundamentals:** n/a

skvale:
This prevents cloneElement throwing an error when a child is null, such as with
```jsx
{condition && <TreeView.Col ...}
```

### Description
- FormRadioGroup was safe because it had `React.isValidElement` checks, but I changed it to be consistent.
- Similar for SideNav
- TreeView had issues with the introduced test, which now passes

fixes #523

### DIFF
--- a/src/Forms/FormRadioGroup.js
+++ b/src/Forms/FormRadioGroup.js
@@ -15,17 +15,13 @@ class FormRadioGroup extends Component {
         return (
             <div
                 {...props}>
-                {React.Children.map(children, child => {
-                    if (React.isValidElement(child)) {
-                        return React.cloneElement(child, {
-                            disabled: child.props.disabled || disabled,
-                            inline: child.props.inline || inline,
-                            name: child.props.name || this.groupId,
-                            onChange: child.props.onChange || onChange
-                        });
-                    } else {
-                        return child;
-                    }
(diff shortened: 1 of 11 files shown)


hertweckhr1:
@skvale can you update this PR? We split our tests up so there are some conflicts, but changes should be pretty simple. Thank you!

skvale:
@hertweckhr1 I just rebased onto master 👍

PR #536, opened 2019-05-03
Chore: Split Panel testing

hertweckhr1:
### Description
_Component_: Panel
* Added additional test files for PanelAction, PanelBody, PanelFilters, PanelHead, PanelHeader, and rearranged tests accordingly.  

Part of #429


scripts/testScripts.sh
```diff
@@ -0,0 +1,115 @@
+#!/bin/bash
```
bcullman: I believe this is a utility script. it should not be checked in.

scripts/testingScripts.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
```
bcullman: I believe this is a utility script. it should not be checked in.

scripts/testScripts.sh
hertweckhr1: Yup yup - I made a commit to delete it but forgot to push it up - it should be deleted now

greg-a-smith:

greg-a-smith:

src/Panel/_PanelHead.test.js
```diff
@@ -0,0 +1,46 @@
+import { mount } from 'enzyme';
+import Panel from './Panel';
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+describe('<Panel />', () => {
```
greg-a-smith: Should reference `<PanelHead />`.

src/Panel/_PanelHeader.test.js
```diff
@@ -0,0 +1,37 @@
+import { Button } from '../';
+import { mount } from 'enzyme';
+import Panel from './Panel';
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+describe('<Panel />', () => {
```
greg-a-smith: Should reference `<PanelHeader />`.

greg-a-smith:

scripts/start.js
```diff
@@ -1,122 +0,0 @@
-/* eslint-disable no-console */
```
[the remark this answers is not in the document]
hertweckhr1: Thank you for the catch - this was never intended to be deleted - it looks like it was accidentally deleted instead of the testing util script file - it's now added back.

PR #537, opened 2019-05-03
Chore: Split side nav tests

hertweckhr1:
### Description
_Component_: SideNav
* Added additional test files for SideNavList and rearranged tests accordingly.  

Part of #429


greg-a-smith:

greg-a-smith:

src/SideNavigation/_SideNavList.test.js
```diff
@@ -0,0 +1,74 @@
+import { mount } from 'enzyme';
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { SideNav } from '../';
+
+describe('<SideNavList />', () => {
+    const sideNavList = (
+        <SideNav.List>
+            <SideNav.ListItem
```
greg-a-smith: There should be a `SideNavListItem.test.js` file too.  If we don't have any tests for it, we should at least create the prop spreading test.

greg-a-smith:

PR #538, opened 2019-05-04
Chore: Split tile tests

hertweckhr1:
### Description
_Component_: Tile
* Added additional test files for ProductTileContent, ProductTileMedia, TileActions, TileCOntent, TileMedia, ProductTile, TileGrid, and rearranged tests accordingly.  

Part of #429


greg-a-smith:

greg-a-smith:

greg-a-smith:

PR #539, opened 2019-05-04
Chore: Split time tests

hertweckhr1:
### Description
_Component_: TimeItem
* Added additional test files for TimeItem and rearranged tests accordingly.  

Part of #429


greg-a-smith:

greg-a-smith:

src/Time/_TimeItem.test.js
```diff
@@ -0,0 +1,176 @@
+import { mount } from 'enzyme';
+import React from 'react';
+import Time from './Time';
+
+describe('Time Item Prop spreading', () => {
```
greg-a-smith: Should just reference `TimeItem` as the top-level `describe`.  Since there are multiple prop spreading tests, you could put another prop spreading `describe` and then move all those tests into that.  Just better test organization.

greg-a-smith:

PR #543, opened 2019-05-08
fix: manually updated rc version

hertweckhr1:
### Description
I merged my previous PRs too quickly from splitting tests.  I manually updated the rc version because of this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+<a name="0.5.2-rc.9"></a>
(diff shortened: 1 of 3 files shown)


greg-a-smith:
This seems like a bad idea.  Why does this need to be updated?

hertweckhr1:
@greg-a-smith because I had merged too fast not fully completing the process in travis and the rc version was not matching up from the release page with package.json, Jenna suggested updating it manually.  @jbadan might have a better explanation.

greg-a-smith:
@hertweckhr1 Looking at npmjs.com, rc 9 never actually made it to getting published (see screenshot).  I think the better approach might be removing the rc 9 tag from the repo.

Screen Shot 2019-05-08 at 11 55 37 AM

hertweckhr1:
At @greg-a-smith's insight, deleted release tag for 0.5.2 rc 9 so no need to change manually.

PR #546, opened 2019-05-17
fix: add disable edge detection to popover

### LINKED ISSUE #540: Need to create disable edge detection prop for Popover
### Description
When popover is longer than screen size it automatically changes the orientation from bottom to top.  Consumer wants to disable edge detection, but there is currently not a prop to do so. 

Need to create a prop that allows consumers to disable edge detection on popover.js. Will need to pass this prop from popper.js to our popover component. 

### Versions
**fundamental-react:** 0.5.1

hertweckhr1:
### Description
* added disableEdgeDetectionProp to Popover passed from popper.js
* verified spread works properly on all components using popover under the covers. 
* added example in popover 
* added links to all popover-derived components (localizationEditor, combo box, time picker, multi input) back to popover in propDescriptions

fixes #540

### DIFF
--- a/src/ComboboxInput/ComboboxInput.js
+++ b/src/ComboboxInput/ComboboxInput.js
@@ -62,8 +62,7 @@ ComboboxInput.propTypes = {
 };
 
 ComboboxInput.propDescriptions = {
-    menu: 'An object containing a `Menu` component.',
-    popoverProps: 'Additional props to be spread to the [Popover](/popover) component.'
+    menu: 'An object containing a `Menu` component.'
 };
 
 export default ComboboxInput;
--- a/src/ComboboxInput/ComboboxInput.js
+++ b/src/ComboboxInput/ComboboxInput.js
@@ -62,7 +62,8 @@ ComboboxInput.propTypes = {
 };
 
 ComboboxInput.propDescriptions = {
-    menu: 'An object containing a `Menu` component.'
+    menu: 'An object containing a `Menu` component.',
+    popoverProps: 'Additional props to be spread to the [Popover](/popover) component.'
 };
 
 export default ComboboxInput;
--- a/src/LocalizationEditor/LocalizationEditor.js
+++ b/src/LocalizationEditor/LocalizationEditor.js
@@ -145,7 +145,6 @@ LocalizationEditor.propDescriptions = {
         language: 'Text to display on the `<button>` element. Meant to be the language of the text in the `<input>`/`<textarea>` element.'
     },
     menu: 'An array of objects that represent the values of the elements in the dropdown menu. The shape of the objects in the array is `{ placeholder: string, language: string, inputProps: object }`.',
-    popoverProps: 'Additional props to be spread to the [Popover](/popover) component.',
     textarea: 'Set to **true** to enable a Localization Editor with a textarea.'
 };
 
--- a/src/LocalizationEditor/LocalizationEditor.js
+++ b/src/LocalizationEditor/LocalizationEditor.js
@@ -145,6 +145,7 @@ LocalizationEditor.propDescriptions = {
         language: 'Text to display on the `<button>` element. Meant to be the language of the text in the `<input>`/`<textarea>` element.'
     },
     menu: 'An array of objects that represent the values of the elements in the dropdown menu. The shape of the objects in the array is `{ placeholder: string, language: string, inputProps: object }`.',
+    popoverProps: 'Additional props to be spread to the [Popover](/popover) component.',
     textarea: 'Set to **true** to enable a Localization Editor with a textarea.'
 };
 
--- a/src/MultiInput/MultiInput.js
+++ b/src/MultiInput/MultiInput.js
@@ -108,6 +108,7 @@ class MultiInput extends Component {
 
     render() {
         const {
+            popoverProps,
             buttonProps,
             compact,
             className,
@@ -147,6 +148,7 @@ class MultiInput extends Component {
                 className={multiInputClasses}>
                 <div className='fd-multi-input-field'>
                     <Popover
+                        {...popoverProps}
                         body={
                             <nav className='fd-menu'>
                                 <ul {...listProps} className='fd-menu__list'>{this.createTagList(data)}</ul>
@@ -201,6 +203,7 @@ MultiInput.propTypes = {
         imageLabel: PropTypes.string
     }),
     placeHolder: PropTypes.string,
+    popoverProps: PropTypes.object,
     tagProps: PropTypes.object,
     onTagsUpdate: PropTypes.func
 };
--- a/src/MultiInput/MultiInput.js
+++ b/src/MultiInput/MultiInput.js
@@ -220,7 +220,6 @@ MultiInput.propDescriptions = {
     localizedTextShape: {
         imageLabel: 'Aria-label in <div> element for image.'
     },
-    popoverProps: 'Additional props to be spread to the [Popover](/popover) component.',
     onTagsUpdate: 'Callback function when a tag is added or removed. Returns array of tags selected.',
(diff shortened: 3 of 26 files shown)


src/Popover/Popover.Component.js
```diff
@@ -163,6 +163,24 @@ export class PopoverComponent extends React.Component {
...
                 </Example>
 
+                <Example
+                    centered
+                    title='Disable Edge Detection'>
+                    <div className='fd-doc__margin--popover'>
+                        <Popover
+                            body={bodyContent}
+                            control={<Button glyph='navigation-up-arrow' option='light' />}
+                            disableEdgeDetection='true'
```
jbadan: For the example we can just put `disableEdgeDetection` instead of `disableEdgeDetection='true'`, since it's boolean it's presence is the list makes it true

src/Popover/Popover.test.js
```diff
@@ -90,6 +106,11 @@ describe('<Popover />', () => {
         component = renderer.create(popOverNoArrow);
         tree = component.toJSON();
         expect(tree).toMatchSnapshot();
+
+        // popover with no disable edge detection
+        component = renderer.create(popOverDisableEdgeDetection);
+        tree = component.toJSON();
+        expect(tree).toMatchSnapshot();
```
jbadan: I'm not sure if this actually tests anything. I can't think of a useful unit test for this, perhaps it's time to bite the bullet and set up puppeteer? Thoughts @greg-a-smith ?

src/TimePicker/TimePicker.js
```diff
@@ -190,6 +193,7 @@ TimePicker.defaultProps = {
 
 TimePicker.propDescriptions = {
     ...Time.propDescriptions,
+    popoverProps: 'Additional props to be spread to the [Popover](/popover) component.',
```
jbadan: Do we have a default for this? If not, we should add one so we don't have to rewrite it in all the components.

src/MultiInput/MultiInput.js
```diff
@@ -217,6 +220,7 @@ MultiInput.propDescriptions = {
     localizedTextShape: {
         imageLabel: 'Aria-label in <div> element for image.'
     },
+    popoverProps: 'Additional props to be spread to the [Popover](/popover) component.',
```
jbadan: Same as other comment, use default here.

src/ComboboxInput/ComboboxInput.js
```diff
@@ -62,7 +62,8 @@ ComboboxInput.propTypes = {
 };
 
 ComboboxInput.propDescriptions = {
-    menu: 'An object containing a `Menu` component.'
+    menu: 'An object containing a `Menu` component.',
+    popoverProps: 'Additional props to be spread to the [Popover](/popover) component.'
```
greg-a-smith: This actually already exists in the `defaults.js` file albeit with "Popover" marked in backticks rather than a link.  I think making it a link is a good idea, but then it should be edited in `defaults.js` so all components get the same change.

src/LocalizationEditor/LocalizationEditor.js
```diff
@@ -145,6 +145,7 @@ LocalizationEditor.propDescriptions = {
         language: 'Text to display on the `<button>` element. Meant to be the language of the text in the `<input>`/`<textarea>` element.'
     },
     menu: 'An array of objects that represent the values of the elements in the dropdown menu. The shape of the objects in the array is `{ placeholder: string, language: string, inputProps: object }`.',
+    popoverProps: 'Additional props to be spread to the [Popover](/popover) component.',
```
greg-a-smith: See previous note about using `defaults.js`.

src/MultiInput/MultiInput.js
greg-a-smith: See previous note about using `defaults.js`.

src/Popover/Popover.Component.js
```diff
@@ -163,6 +163,24 @@ export class PopoverComponent extends React.Component {
...
+                            placement='top' />
+
+                        <Popover
+                            body={bodyContent}
+                            control={<Button glyph='navigation-down-arrow' option='light' />}
+                            disableEdgeDetection='true'
```
greg-a-smith: Same comment about boolean props.

src/Popover/Popover.test.js
greg-a-smith: I agree that this test doesn't really do anything.  The only way to visually test this prop is to open the popover and scroll to a point where it should flip and then verify it doesn't.

Since that functionality is really core to popper.js, we could probably get away with a test that makes sure our internal Popper component gets this value passed to it and the modifiers are set correctly.  Let me know if you need/want more explanation on that.

src/TimePicker/TimePicker.js
greg-a-smith: See previous note about using `defaults.js`.

src/Popover/Popover.Component.js
```diff
@@ -163,6 +163,24 @@ export class PopoverComponent extends React.Component {
...
+                        <Popover
+                            body={bodyContent}
+                            control={<Button glyph='navigation-up-arrow' option='light' />}
+                            disableEdgeDetection
+                            placement='top' />
+
+                        <Popover
+                            body={bodyContent}
+                            control={<Button glyph='navigation-down-arrow' option='light' />}
+                            disableEdgeDetection='true'
```
jbadan: I think you missed one on 179.

src/Popover/Popover.js
```diff
@@ -97,9 +100,10 @@ Popover.defaultProps = {
 Popover.propDescriptions = {
     body: 'Node(s) to render in the overlay.',
     control: 'Node to render as the reference element (that the `body` will be placed in relation to).',
+    disableEdgeDetection: 'Set to **true** to render popover without edge detection so popover will not flip from top to bottom with scroll.',
     noArrow: 'Set to **true** to render a popover without an arrow.',
     placement: 'Initial position of the `body` (overlay) related to the `control`.',
-    popperProps: 'Additional props to be spread to the overlay element.',
+    popperProps: 'Additional props to be spread to the overlay element, supported by [popper.js](https://popper.js.org/).',
```
jbadan: I'm not sure if this will work, but can we try adding `{:target="_blank"}` to this link.

hertweckhr1: {:target="_blank"} does not work - and currently can't find a solution to open on a separate page.

jbadan: I just played around with it and if you edit the `_PropertyDescription.js` to `import ReactMarkdown from 'react-markdown/with-html';` and `escapeHtml={false}` to the component, you can change that link to be an anchor tag in regular html with a target blank. See the react-markdown docs for more info!

jbadan: https://github.com/rexxars/react-markdown#parsing-html

src/Popover/Popover.js
```diff
@@ -97,9 +100,10 @@ Popover.defaultProps = {
 Popover.propDescriptions = {
     body: 'Node(s) to render in the overlay.',
     control: 'Node to render as the reference element (that the `body` will be placed in relation to).',
+    disableEdgeDetection: 'Set to **true** to render popover without edge detection so popover will not flip from top to bottom with scroll.',
     noArrow: 'Set to **true** to render a popover without an arrow.',
     placement: 'Initial position of the `body` (overlay) related to the `control`.',
-    popperProps: 'Additional props to be spread to the overlay element.',
+    popperProps: 'Additional props to be spread to the overlay element, supported by <a href="https://popper.js.org" target="blank">popper.js</a>.',
```
greg-a-smith: Although it appears to work, the correct target value is `_blank`.

PR #548, opened 2019-05-25
feat: refactor Tile removing ProductTile component

### LINKED ISSUE #440: Refactor Tile and ProductTile into a single component
### Description
The components `Tile` and `ProductTile` are very similar -- really just a difference in CSS classes.  Their respective subcomponents should be able to be refactored as well.  There shouldn't have to be both `TileContent` and `ProductTileContent` nor should there need to be both `TileMedia` and `ProductTileMedia`.

The specs don't show `ProductTile` being used in a `TileGrid`, but is there a reason why it _shouldn't_ work?

### Versions
**fundamental-react:** Any
**fiori-fundamentals:** Any

hertweckhr1:
### Description
Refactored Tile to reduce redundancy:
* Product Tile now consumer Tile
* ProductTileMedia now consumes TileMedia
* ProductTileContent now consumes TileContent

New prop productTile(boolean) was created and passed down. 

Questions to consider: 
* They are now refactored so product tile and its sub components are consuming Tile and its subcomponents.  Is this what we want to or are we trying to eliminate productTile altogether and just have the productTile prop?

fixes #440


src/Tile/_ProductTileContent.js
```diff
@@ -1,37 +1,13 @@
...
-    const HeadingTag = `h${headingLevel}`;
-
-    return (
-        <div {...rest} className={tileProductContentClasses}>
-            <HeadingTag {...titleProps} className='fd-product-tile__title'>{title}</HeadingTag>
-            {children}
-        </div>
-    );
+    return (<TileContent productTile title={title}
+        {...props} />);
```
jbadan: Can we just pass props without pruning title here?

src/Tile/_ProductTileContent.js
```diff
@@ -1,37 +1,10 @@
-import classnames from 'classnames';
```
greg-a-smith: I think this sub-component can be removed altogether as the only thing it does is pass a `productTile` prop to `TileContent`.  This can be done by cloning children with the `Tile` component and passing the top-level `productTile` prop.

src/Tile/ProductTile.js
```diff
@@ -1,37 +1,14 @@
...
 
 ProductTile.displayName = 'ProductTile';
 
-ProductTile.propTypes = {
-    children: PropTypes.node,
-    className: PropTypes.string,
-    disabled: PropTypes.bool
-};
-
 ProductTile.Content = ProductTileContent;
```
greg-a-smith: This can be changed to `ProductTile.Content = TileContent;` (see other related comments)

src/Tile/Tile.js
```diff
@@ -47,13 +49,19 @@ Tile.propTypes = {
     colorAccent: PropTypes.number,
     columnSpan: CustomPropTypes.range(1, 6),
     disabled: PropTypes.bool,
+    productTile: PropTypes.bool,
```
greg-a-smith: I think this component should clone children and pass the top-level `productTile` prop to each child so the prop only needs to be added at the top-level component. (see related comments)

src/Tile/_ProductTileMedia.js
```diff
@@ -1,28 +1,20 @@
...
             style={{ backgroundImage: 'url(' + image + ')' }} />
     );
 };
 
 ProductTileMedia.displayName = 'ProductTile.Media';
 
 ProductTileMedia.propTypes = {
-    image: PropTypes.string.isRequired,
-    className: PropTypes.string
+    image: PropTypes.string.isRequired
```
greg-a-smith: In the `propDescriptions` for this prop, it should probably say "URL of the background image" so it's clearer to the consumer what is to be expected.

greg-a-smith:
@hertweckhr1 @jbadan  Ok, I've been looking at this some more.  It's sort of a strange assembly of components and sub-components.  Looking at the docs site examples (and playground), there seems to be three types: simple, media and product.  They are very similar, but the media aspect of media and product tiles is different -- one accepts children, which seem best suited for an `Image` component while the other accepts a URL for a background image.  Because of this, they don't follow a very good pattern for component structure (or at least not reusable).

What about something like this?

**Simple:**
```jsx
<Tile>
    <Tile.Content title="Text">
        Some content
    </Tile.Content>
</Tile>
```

**Media:**
```jsx
<Tile>
    <Tile.Media>
        <Image />
    </Tile.Media>
    <Tile.Content title="Text">
        Some content
    </Tile.Content>
</Tile>
```

**Product:**
```jsx
<Tile backgroundImage="http://some.url/image" productTile>
    <Tile.Content title="Text">
        Some content
    </Tile.Content>
</Tile>
```

That would get rid of all the `ProductTile` components and sub-components (and subsequently the need to find a way to document them) and just support the product tile classes with the `productTile` prop on `Tile`.  The `_TileMedia` component would then only be used for the media tiles.

Thoughts?

src/Tile/_ProductTileContent.js
hertweckhr1: Yup this was my bad, I was testing this out when I was having issues with a warning message and forgot to remove.

hertweckhr1:
@greg-a-smith @jbadan - this sounds good to me.  Would it end up being 4 different types with Tile actions being added on as well? Or is having actions on a tile an option for all 3 simple, media, and product?

greg-a-smith:
@hertweckhr1 To me, I think this really converts the different "types" of tiles to patterns rather than components.  In this single component model (with sub-components), you could _technically_ put actions or media (image) on a product tile, but that's not a supported pattern.  I think the examples on the docs site should just show how to use the components to create each option:

* Simple tile
* Simple tile with an image (media)
* Simple tile with actions
* Simple tile with an image (media) and actions
* Product tile

src/Tile/Tile.Component.js
```diff
@@ -74,24 +74,50 @@ export const TileComponent = () => {
...
+                            }
+                            control={<Button glyph='vertical-grip' option='light' />}
+                            placement='bottom-end' />
+                    </Tile.Actions>
+                </Tile>
+            </Example>
+
             <Example
                 title='Product Tile'>
                 <div>
```
jbadan: Can we get rid of the `<div>` and just replace with a fragment if it's even needed?

src/Tile/Tile.Component.js
```diff
@@ -74,24 +74,50 @@ export const TileComponent = () => {
...
+                    <Tile productTile role='button'>
+                        <Tile.Media backgroundImage='https://placeimg.com/820/480/any' />
+                        <Tile.Content title='Tile Title'>
                             <p>Tile Description</p>
-                        </ProductTile.Content>
-                    </ProductTile>
+                        </Tile.Content>
+                    </Tile>
 
                     <br />
```
jbadan: Can we remove the `<br />`'s to clean up the example output html a bit?

src/Tile/Tile.Component.js
```diff
@@ -74,24 +74,50 @@ export const TileComponent = () => {
...
+            </Example>
+
             <Example
                 title='Product Tile'>
                 <div>
-                    <ProductTile role='button'>
-                        <ProductTile.Media image='https://techne.yaas.io/images/product-thumbnail-wide.png' />
-                        <ProductTile.Content title='Tile Title'>
+                    <Tile productTile role='button'>
+                        <Tile.Media backgroundImage='https://placeimg.com/820/480/any' />
```
jbadan: I think we might need to bring in an image that's consistent for this - or at least figure out the best dimensions here and maybe limit to nature? It's get's a little funky looking on wide screens. 

Screen Shot 2019-06-03 at 8 40 41 AM

src/Tile/Tile.Component.js
```diff
@@ -74,24 +74,50 @@ export const TileComponent = () => {
...
+                </Tile>
+            </Example>
+
             <Example
                 title='Product Tile'>
                 <div>
-                    <ProductTile role='button'>
-                        <ProductTile.Media image='https://techne.yaas.io/images/product-thumbnail-wide.png' />
-                        <ProductTile.Content title='Tile Title'>
+                    <Tile productTile role='button'>
```
jbadan: When we're allowing the productTile to be used as a button, should we swap out for actual html button syntax? This is more a question for fundamental-styles but we should probably think about it.

src/Tile/_TileMedia.js
```diff
@@ -3,21 +3,31 @@ import PropTypes from 'prop-types';
...
-    return <div {...rest} className={tileMediaClasses}>{children}</div>;
+    return (
+        <div {...rest} className={tileMediaClasses}
+            style={{ backgroundImage: 'url(' + backgroundImage + ')' }}>{children}</div>);
 };
 
 TileMedia.displayName = 'Tile.Media';
 
 TileMedia.propTypes = {
+    backgroundImage: PropTypes.string,
```
greg-a-smith: I was thinking this prop would/could be on the `Tile` component since that already has a prop called `backgroundColor`.  To me, `TileMedia` is more designed as a wrapper for children (images) rather than a URL string prop.

src/Tile/Tile.test.js
```diff
@@ -89,6 +89,26 @@ describe('<Tile />', () => {
         </Tile>
     );
 
+    const productTile = (
+        <Tile className='pink' isButton
```
jbadan: What is the `isButton` prop? I'm not seeing it documented anywhere.

src/Tile/Tile.test.js
```diff
@@ -89,6 +89,26 @@ describe('<Tile />', () => {
...
+            productTile>
+            <Tile.Media backgroundImage='https://techne.yaas.io/images/product-thumbnail-wide.png' />
```
jbadan: We should probably update these to an internal image since that one doesn't exist anymore or just a dummy url since nothing is getting rendered.

src/Tile/_TileMedia.js
```diff
@@ -3,21 +3,31 @@ import PropTypes from 'prop-types';
...
+            'fd-tile__media': !productTile,
+            'fd-product-tile__media': productTile
+        },
         className
     );
 
-    return <div {...rest} className={tileMediaClasses}>{children}</div>;
+    return (
+        <div {...rest} className={tileMediaClasses}
+            style={{ backgroundImage: 'url(' + backgroundImage + ')' }}>{children}</div>);
```
jbadan: Should we only have this style block render if a background image exists? 
Otherwise we end up with this in the markup: 
`<div class="fd-tile__media" style="background-image: url("undefined");"><span class="fd-image--l fd-image--circle" style="background-image: url("https://placeimg.com/400/400/nature");"></span></div>`

See related comment on snapshots.

src/Tile/__snapshots__/Tile.test.js.snap
```diff
@@ -45,6 +45,11 @@ exports[`<Tile /> create tile component 3`] = `
 >
   <div
     className="fd-tile__media green"
+    style={
+      Object {
+        "backgroundImage": "url(undefined)",
```
jbadan: See my comment on `_TileMedia`

src/_playground/documentation/Playground/Playground.js
```diff
@@ -533,9 +533,9 @@ class Playground extends Component {
               } else if (this.state.component === 'product') {
                   componentToGenerate = (
                       <Tile>
-                          <ProductTile>
-                              <ProductTile.Media image='https://techne.yaas.io/images/product-thumbnail-wide.png' />
-                              <ProductTile.Content
+                          <Tile>
```
jbadan: I think we should either remove the playground component for this page or fix it to work with the new changes so that all the options are there.

src/Tile/Tile.js
```diff
@@ -34,7 +36,15 @@ const Tile = props => {
         <div
             {...rest}
             className={tileClasses}>
-            {children}
+            {React.Children.toArray(children).map(child => {
+                if (child.type !== TileActions) {
```
greg-a-smith: I think it's slightly safer to do this equality (or non-equality) check using the `displayName` (see `TreeView` code).

src/Tile/_TileMedia.js
hertweckhr1: @greg-a-smith would you have the style tag for backgroundImage go on Tile itself then instead of media or would you pass down the backroundImage prop to media?  When I originally put it on Tile, it did not format correctly: the image took up the entire background of the Title portion as well and was not big enough, so thought to keep it in Media where there is the 'fd-product-tile__media' class has previously lived maybe making it easier for the consumer to associate media with a photo in general? (whether it be background of small icon). 

Either if fine with me - can you confirm if you want it on tile I should put the style tag on tile and add the fd-product-tile__media class name as well?

src/Tile/Tile.Component.js
greg-a-smith: I think this entire concept of tiles being buttons needs to be addressed in fd-styles and fd-react going forward.  The current props/implementation seems half-baked.

src/Tile/_TileMedia.js
greg-a-smith: I think the final emitted markup would be the same as you have it now, but `Tile` would have the prop and would render the child `<div>` where it needs it to go.  You are right, the way the styles are written, placing the `style` attribute on the top-level tile `<div>` would not be correct.

src/Tile/Tile.Component.js
hertweckhr1: Sounds good - looks like 4:1 ratio is good for header image and will switch it to nature 👍

src/Tile/Tile.Component.js
hertweckhr1: Sounds good - Looks like 4:1 is a good ratio for header images and will switch it to nature too 👍

src/Tile/Tile.test.js
hertweckhr1: This was in the previous productTile tests so I assumed it was a prop passed down 😮 - will take it out.

src/Tile/Tile.Component.js
hertweckhr1: @jbadan @greg-a-smith should I comment this as a topic for our thursday meeting?

src/Tile/Tile.js
```diff
@@ -34,7 +37,20 @@ const Tile = props => {
         <div
             {...rest}
             className={tileClasses}>
-            {children}
+            {productTile &&
+                <div className='fd-product-tile__media' style={{ backgroundImage: 'url(' + backgroundImage + ')' }} />
+            }
```
hertweckhr1: I created a conditional for product tile rather than if the background image exists bc product tiles should have background images with them always. This will eliminate the issue Jenna mentioned with background image rendering undefined for tiles that are not product.

greg-a-smith: Well, _technically_ it won't.  There are really two props needed for a product tile:  `productTile` and `backgroundImage` so it's possible the consumer passes one, but not the other.  That said, this has become more of a pattern-based component so I think this is a good change.  We _could_ make `productTile` a shape prop with a required `backgroundImage`, but that might be overkill here.  As we've said in this PR, the consumer just wants to know which props to pass and the example page will help them with that.

PR #566, opened 2019-06-06
feat: switch to using Fundamental Styles repo

jbadan:
### Description

Switch from fiori-fundamentals to fundamental styles. 

fixes #issueid


hertweckhr1:
It looks like the styling is getting tweaked a little bit for some pages: 
* placemet of *EN (right is local branch) 
image

* color of alert (right is local branch)
image

jbadan:
@hertweckhr1 These changes are bugs we should open in fundamental-styles. We have version 1.5.1 of fiori fundamentals on master, here is using 1.5.8. Something changed in between those versions that is causing problems:

Screen Shot 2019-06-07 at 8 49 09 AM

jbadan:
LocalizationEditor button alignment fix: https://github.com/SAP/fundamental-styles/pull/35

PR #593, opened 2019-06-12
chore: remove babel-jest

jbadan:
### Description

Babel-jest is unused. Removing. 

fixes #issueid


jbadan:
@bcullman done! Can't merge without an actual approval though :)

PR #632, opened 2019-06-27
feat: work needed to bring in self contained styles

jbadan:
### Description

This will be a **breaking change** for consumers as they will need to update their webpack config as detailed in the updated readme.

This pr brings in the self-contained styles for Button, updates/simplifies the webpack configuration, updates jest config to allow for parsing scss.

Size-limit was not setup to handle scss files, I copied their internal config as much as possible with the addition of scss loaders. Caveat: the output now contains our peerDependencies (? I can't figure out why) and is reporting our **parsed size, not gzipped**

MASTER:
Screen Shot 2019-06-26 at 2 53 30 PM

This PR (note - peer dependencies like react-dom and react and some devdeps for webpack are now being included on the node_modules side - I've attempted to troubleshoot this for a while, but I think at this point it's out of range of the spike and should be a followup):
Screen Shot 2019-06-27 at 8 44 32 AM

Build:index works again! Thanks @greg-a-smith


.size-limit
```diff
@@ -1,8 +1,7 @@
 [
   {
     name: "Fundamental-React Size",
-    webpack: true,
-    path: "lib/index.js",
-    limit: "50 KB"
+    config: "config/webpack.config.sizeLimit.js",
+    limit: "550 KB"
```
bcullman: whoa. sad.

jbadan: did you read the description? 😈 

50kb was gzipped, using the new custom webpack config I haven't been able to get it to return gzipped size to size-limit, so it is reporting the parsed size. While we're now at double the gzipped size (including the peer dependency weirdness).

We're really only 12kb larger than we were previously.

config/webpack.config.sizeLimit.js
```diff
@@ -0,0 +1,59 @@
...
+            },
+            {
+                test: /\.(woff(2)?|ttf|eot|svg)(\?v=\d+\.\d+\.\d+)?$/,
+                use: [{
+                    loader: 'file-loader'
+                }]
+            }
+        ]
+    },
+    externals: {
```
jbadan: Wooo! 🎉 Thanks @bcullman

src/Button/Button.js
```diff
@@ -1,3 +1,4 @@
+import '../../node_modules/fundamental-styles/components/button.scss';
```
greg-a-smith: Does this need to include the `../../node_modules` in the path or could the import be simplified to `import 'fundamental-styles/components/button.scss';`?

Issue #498, opened 2019-04-02
Unable to get date from DatePicker


avalkowsky:
### Description
I am trying to use DatePicker, however the docs forget to mention the most important thing about DatePicker (actually, it is the same for other input/input-like components, but for others I was able to make it work) : 
1) how to get date  out of the DatePicker
2) how to set date programmatically in the DatePicker

I tried to use onChange and value, but they did not work so the components renders but I cannot use the set date in any way:
`<DatePicker
									onChange={(event) => onActiveToChange(event.target.value)}
									value={activeTo}
								/>`

### Versions
    "fiori-fundamentals": "^1.4.9",
    "fundamental-react": "^0.5.0"

edit: I checked the source code carefully, but I don't see there any option to set and get the date selected on the DatePicker. 

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

avalkowsky:
btw. placeholder is 'mm/dd/yyyy' while date is formatted to 'dd/mm/yyyy' ;)

avalkowsky:
btw. same issue for TimePicker

brunohaetinger:
@avalkowsky 
The way I did to get the date value is like this:

On the datePicker Parent component, use the "ref" attribute.
```
<DatePicker ref={datePicker => this.myDatePicker = datePicker}/>
```

Then, you can use that reference from any parent component function just calling:
```
this.myDatePicker.state.formattedDate
```
--------------

Complete component example:
ParentComponent.js
```
import React, { Component } from 'react';
import { DatePicker } from 'fundamental-react/lib/DatePicker';

class ParentComponent extends Component {
    constructor(props) {
        super(props);
        this.state = {
        }
    }
    
    printDate = () =>{
        console.log(this.myDatePicker.state.formattedDate);
    }

    render() {
        return (
            <React.Fragment>
                <DatePicker ref={datePicker => this.myDatePicker = datePicker}/>
                <button onClick={this.printDate}>Print Date</button>
            </React.Fragment>
        );
    }
}


export default ParentComponent;
```

avalkowsky:
[...]

PR #712, opened 2019-08-12
fix: CORUI-6949: add onBlur callback to DatePicker

jbadan:
### Description

Without doing a full rewrite at the moment, this is as close to a functional callback as we can get. Tests were a nightmare, so I decided to forgo them for this bandaid. 

This behaves slightly erratically as the input is not wired up to the calendar/sharing validation. 

Known bugs: Changes to the input don't necessarily update the date/trigger any event.

fixes https://github.com/SAP/fundamental-react/issues/498

### DIFF
--- a/src/DatePicker/DatePicker.js
+++ b/src/DatePicker/DatePicker.js
@@ -367,7 +367,8 @@ DatePicker.defaultProps = {
 
 DatePicker.propDescriptions = {
     ...Calendar.propDescriptions,
-    enableRangeSelection: 'Set to **true** to enable the selection of a date range (begin and end).'
+    enableRangeSelection: 'Set to **true** to enable the selection of a date range (begin and end).',
+    onBlur: 'Callback function for onBlur events. In the object returned, date is the date object and formattedDate is the formatted date.'
 };
 
 export default DatePicker;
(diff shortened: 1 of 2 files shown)


src/DatePicker/DatePicker.js
```diff
@@ -303,6 +308,8 @@ class DatePicker extends Component {
                             <input
                                 {...inputProps}
                                 className={datePickerInputClasses}
+                                onBlur={() => onBlur({ date: this.state.selectedDate,
+                                    formattedDate: this.state.formattedDate })}
```
greg-a-smith: This would be better written as a separate function in the component and then referenced here like `onBlur={this._handleBlur}`.  Otherwise, this actually creates a new instance of the function each time it goes through the `render` function.

src/DatePicker/DatePicker.js
```diff
@@ -350,12 +357,18 @@ DatePicker.propTypes = {
...
+
+DatePicker.defaultProps = {
+    onBlur: () => {}
 };
 
 DatePicker.propDescriptions = {
     ...Calendar.propDescriptions,
-    enableRangeSelection: 'Set to **true** to enable the selection of a date range (begin and end).'
+    enableRangeSelection: 'Set to **true** to enable the selection of a date range (begin and end).',
+    onBlur: 'Callback function for onBlur events. In the object returned, date is the date object and formattedDate is the formatted date.'
```
greg-a-smith: Maybe add backticks around `date` and `formattedDate` when describing the returned object.

PR #739, opened 2019-09-10
Fix unique `key` prop error in Menu.Item and SearchInput

### LINKED ISSUE #512: Shellbar causes multiple React keys warnings
### Description
When using the Shellbar component, multiple errors come up in the console.

Upon mount:
```
Warning: Each child in a list should have a unique "key" prop. See https://fb.me/react-warning-keys for more information.
    in div
    in div
    in div (created by SearchInput)
    in SearchInput
    in div
    in div
    in div
    in div
    in Shellbar
...
```
Upon click on the user menu:
```
Warning: Each child in a list should have a unique "key" prop. See https://fb.me/react-warning-keys for more information.
    in Menu.Item
    in ul (created by Menu.List)
    in Menu.List
    in nav (created by Menu)
    in Menu
    in div (created by InnerPopper)
    in InnerPopper (created by Context.Consumer)
    in Popper
    in span (created by Foco)
    in Foco
    in Manager
    in Popper
    in Popover
    in div
    in div
    in div
    in div
    in div
    in Shellbar 
...
```

The mounted shellbar has a search input as well as a user menu. 

### Versions
**fundamental-react:** 0.5.0
**fiori-fundamentals:** 1.5.1
**react**: 16.8.1

---

_**DISCLAIMER:**

parostatkiem:
### Changes
- add and document `uniqueKey` prop in Menu.Item (fixes error when used)
- add unique `key` prop to SearchInput's internal Button element (fixes error)
- Add key to input element inside Toggle component (fixes error)

fixes #512

### DIFF
--- a/src/Menu/Menu.test.js
+++ b/src/Menu/Menu.test.js
@@ -141,6 +141,15 @@ describe('<Menu />', () => {
         });
     });
 
+    describe('MenuItem', () => {
+        test('create menu item with uniqueKey', () => {
+            const key = 'testkey';
+            const component = mount(<Menu.Item uniqueKey={key}></Menu.Item>).find('li');
+
+            expect(component.at(0).key()).toBe(key);
+        });
+    });
+
     describe('Prop spreading', () => {
         test('should allow props to be spread to the Menu component', () => {
             const element = mount(<Menu data-sample='Sample' />);
--- a/src/Menu/_MenuItem.js
+++ b/src/Menu/_MenuItem.js
@@ -74,6 +74,7 @@ MenuItem.propDescriptions = {
     children: 'component - can be used to pass React Router <Link> or any other component which emits an <a>.',
     isLink: 'Set to **true** to style as a link.',
     separator: 'Set to **true** to add a horizontal line (separator).',
+    uniqueKey: 'A value that is used as React\'s `key` property in the inner `<li>` element.',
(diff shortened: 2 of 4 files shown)


bcullman:
Thanks for the PR couple of thoughts

Should probably should set default prop with https://github.com/SAP/fundamental-react/blob/master/src/utils/shortId.js

Need to add a backing unit test.

parostatkiem:
@bcullman Thanks for looking at it.
`shortId` generates a new ID with every execution of `shortId.generate()`, am I right?
If yes, every outter component render would cause them to regenerate. This would have negative impact on React's performance - the whole idea of list items' keys would be gone (they have to be static and different for each item). 

I'll take care of unit tests.

sjanota:
I've been playing a little with this PR and it looks like components are working perfectly well when I copy the JSX code to my project. However, when I use a compiled version from npm it throws the react error.

parostatkiem:
@sjanota This may be connected with the fact that none of these errors is thrown on the Documentation Website, even in dev mode but they are thrown when the same components are used in another app.

sjanota:
I've been playing with it a while, but haven't found anything concrete. I'm pretty sure it's in the way the library is compiled, though. I left a comment [here](https://github.com/SAP/fundamental-react/issues/512#issuecomment-531192816).

parostatkiem:
fixed by https://github.com/SAP/fundamental-react/pull/741

PR #749, opened 2019-09-26
feat: wrap all components with withStyles HOC

jbadan:
### Description
The new `withStyles` higher order component wraps each component and dynamically requires the relevant styles, or disable the styles based on props.

Pass custom styling via require statement in `customStyles` prop.
Disable all styles using `disableStyles` prop

This pr introduces storybook! 📖 🎉 

Functional components now with wrapped with `forwardRefs`

`npm run start:playground` or `npm run storybook` to start it

> **NOTE** that you will need to refresh the page to see different style combos because the style will persist across pages.

**To do**:

- [x] docs site - markdown for code samples not working. This has to do with the forwarding refs
- [x] bug: Numbers in the calendar not showing up in storybook but are in docs site
- [x] update all tests that use .state() - not compatible with wrapped components 💀 😭 
- [x]  audit subcomponents to see if they need `withStyles` wrapper
- [x] finish remaining components
- [x] figure out why docs site in netlify is bringing in the `customTestStyles` 😆 - hot pink navigation anyone? 
- [x] localizationEditor cascade disableCSS
- [x] prop description tables


src/ActionBar/__stories__/ActionBar.stories.js
```diff
@@ -0,0 +1,40 @@
...
+            <ActionBar.Header description={'Action Bar Description'} title={'Page Title'} />
+            <ActionBar.Actions>
+                <Button disableStyles>Button</Button>
+                <Button disableStyles option='emphasized'>Button</Button>
+            </ActionBar.Actions>
+        </ActionBar>
+    ))
+    .add('custom styles', () => (
+        <ActionBar customStyles={require('../../utils/WithStyles/customStylesTest.css')}>
+            <ActionBar.Back disableStyles />
```
greg-a-smith: I don't think you need the `disableStyles` on the child component, do you?

src/utils/WithStyles/WithStyles.js
```diff
@@ -0,0 +1,69 @@
...
+                        require('fundamental-styles/dist/' + defaults.cssFile + '.css');
+                } else {
+                    customStyles;
+                }
+            }
+
+            return (
+                <WrappedComponent
+                    customStyles={customStyles}
+                    disableStyles={disableStyles}
```
greg-a-smith: Do both `customStyles` and `disableStyles` need to be passed through here?  I guess `disableStyles` is useful so components can pass that along to child components, but `customStyles` is only used right here, right?

src/ActionBar/ActionBar.js
```diff
@@ -1,31 +1,39 @@
...
 import ActionBarActions from './_ActionBarActions';
 import ActionBarBack from './_ActionBarBack';
 import ActionBarHeader from './_ActionBarHeader';
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
 import React from 'react';
+import withStyles from '../utils/WithStyles/WithStyles';
 
-const ActionBar = ({ children, className, ...props }) => {
+const ActionBar = React.forwardRef(({ children, className, customStyles, disableStyles, ...props }, ref) => {
```
greg-a-smith: I think we can get rid of the `customStyles` in this prop pruning use case since this never gets passed here from `withStyles`.  It should stay in the PropTypes definition, however, because consumers are unaware that the component is wrapped in an HOC.

src/TimePicker/TimePicker.js
```diff
@@ -3,6 +3,7 @@ import Popover from '../Popover/Popover';
 import PropTypes from 'prop-types';
 import Time from '../Time/Time';
 import TimePickerItem from './_TimePickerItem';
+import withStyles from '../utils/WithStyles/WithStyles';
```
greg-a-smith: Does `TimePicker` need the ref forwarding?

src/Toggle/Toggle.js
```diff
@@ -1,11 +1,10 @@
-import 'fundamental-styles/dist/toggle.css';
 import classnames from 'classnames';
 import FormItem from '../Forms/FormItem';
 import FormLabel from '../Forms/FormLabel';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { TOGGLE_SIZES } from '../utils/constants';
-
+import withStyles from '../utils/WithStyles/WithStyles';
 class Toggle extends React.Component {
```
greg-a-smith: Does `Toggle` need the ref forwarding?

src/TreeView/TreeView.js
```diff
@@ -1,11 +1,11 @@
...
 import Tree from './_Tree';
 import TreeBranch from './_TreeBranch';
 import TreeCol from './_TreeCol';
 import TreeHead from './_TreeHead';
 import TreeItem from './_TreeItem';
 import TreeRow from './_TreeRow';
+import withStyles from '../utils/WithStyles/WithStyles';
 import React, { Component } from 'react';
 
 class TreeView extends Component {
```
greg-a-smith: Does `TreeView` need the ref forwarding?

src/DatePicker/DatePicker.test.js
```diff
@@ -23,7 +23,7 @@ describe('<DatePicker />', () => {
     });
 
     test('open/close by calendar icon button', () => {
-        wrapper = mount(defaultDatePicker);
+        wrapper = mount(defaultDatePicker).children().children();
```
markpalfreeman: There are ~80 instances of this `.children().children()` pattern—can we write a single mount function with a comment for why we need to dive down 2 levels?

src/utils/WithStyles/WithStyles.js
```diff
@@ -0,0 +1,73 @@
...
+        customStyles: PropTypes.object,
+        disableStyles: PropTypes.bool,
+        forwardedRef: PropTypes.oneOfType([
+            PropTypes.func,
+            PropTypes.shape({ current: PropTypes.any })
+        ])
+    };
+
+    //HOCs do not automatically pass refs as they are not technically props
+    function forwardRef(props, ref) {
```
markpalfreeman: This function name makes it sound like an imperative side-effecty thing, when really this is a component. Is there a better name?

Also, can this just return the "with ref" version so the hoisted thing below is a little clearer? Maybe something like:
```javascript
const componentWithRef = (props, ref) => React.forwardRef(<WithStyles {...props} forwardedRef={ref} />);

const componentWithStaticMethods = hoistNonReactStatic(componentWithRef, WrappedComponent);
```

src/ActionBar/ActionBar.js
```diff
@@ -1,31 +1,39 @@
...
+    customStyles: PropTypes.object,
+    disableStyles: PropTypes.bool
 };
 
 ActionBar.Actions = ActionBarActions;
 ActionBar.Back = ActionBarBack;
 ActionBar.Header = ActionBarHeader;
 
-export default ActionBar;
+export { ActionBar as __ActionBar };
```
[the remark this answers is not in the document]
greg-a-smith: To have access to the "plain" component for the purposes of building the docs site (property tables and import statements).

src/utils/WithStyles/WithStyles.js
jbadan: We can definitely try that out in the next iteration - I took this from the react docs to rather be safe than sorry.

src/DatePicker/DatePicker.test.js
jbadan: Definitely should be on the to do for the next iteration ✅

src/ActionBar/ActionBar.js
```diff
@@ -1,31 +1,39 @@
...
     const actionBarClasses = classnames(
         'fd-action-bar',
         className
     );
 
     return (
-        <div {...props} className={actionBarClasses}>{children}</div>
+        <div {...props}
+            className={actionBarClasses}
+            ref={ref}>
```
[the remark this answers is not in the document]
jbadan: The reasoning was to go ahead and add them to every functional component being wrapped in `withStyles` as it will be a breaking change to add them going forward. See: https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers

PR #761, opened 2019-10-02
fix: withStyles followup including typo fix and new test utility

jbadan:
### Description
Follow up to https://github.com/SAP/fundamental-react/pull/749

fixes typo

adds `mountComponent` test util to reduce instances of `.children().children()` needed to traverse the HOCs. 

> NOTE: I only replaced the obvious instances. Anything that had complex click events needed the `.children().children()` to continue working correctly. This is most likely a bug and should have a followup issue. I wasn't comfortable changing tests simply for beauty. 

After the updated eslint version, some lint rules are now failing for sorting - fixed here.

fixes #issueid

### DIFF
--- a/src/Time/Time.test.js
+++ b/src/Time/Time.test.js
@@ -1,5 +1,5 @@
 import { mount } from 'enzyme';
-import { mountComponent } from '../utils/testUtils';
+import { mountComponentWithStyles } from '../utils/testUtils';
 import React from 'react';
 import renderer from 'react-test-renderer';
 import Time from './Time';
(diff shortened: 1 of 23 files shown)


src/Time/Time.test.js
```diff
@@ -610,30 +611,24 @@ describe('<Time />', () => {
...
             .find(
                 'button.fd-button--light.sap-icon--navigation-up-arrow'
             )
             .at(2)
             .simulate('click');
-        expect(wrapper.state('time').hour).toEqual(1);
-        expect(wrapper.state('time').minute).toEqual(0);
+        expect(wrapper.state('time').hour).toEqual('01');
+        expect(wrapper.state('time').minute).toEqual('00');
         expect(wrapper.state('time').second).toEqual('01');
```
jbadan: I don't want to keep this change in this pr, but just included it to get some context. Why are the tests expecting strings sometimes, numbers other times?? See commented out test on line 614 and test on 613 - testing the same thing except string versus number. How is this even happening?

PR #764, opened 2019-10-08
fix: Change focus lock library

### LINKED ISSUE #765: Dropdown click doesn't work in Modal
### Description
The recently added _focus-lock_ feature added to `Modal` component seems to break popover's functionality when it's inside `Modal`. The `Dropdown` which is inside modal opens the `Popover` properly (no issues here) but any item (`Menu.Item` or any other element) that is inside `Popover` has its click listener removed. 

Screenshot 2019-10-09 at 09 22 04

This problem happens only if **all** of the below requirements are fulfilled:
- The `Popover` is inside modal
- The `Modal` is inside Portal (`usePortal` param is set)

Every _click_ listener is simply removed from the list item (I can see `noop` in browser DOM inspector). During long-lasting research it turned out that the faulty bit is `focus-trap-react` library. 

### Versions
**fundamental-react:** 0.7.0-rc.73 but happens also in 0.6.1
**fundamental-styles:** None

---

_**DISCLAIMER:**

parostatkiem:
### Description
Change `focus-trap-react` library to `react-focus-lock` due to problems when it was used in Portal. 
One test was changed because it was performed on the first rendered `<div>`. `react-focus-lock` renders 3 `<div>` elements (2 helpers and the main one) and the third one is the actual Modal.

It looks like people are having similar issues:
https://github.com/davidtheclark/focus-trap-react/issues/27

fixes #765


jbadan:
@parostatkiem We will merge it, we were waiting to talk with the other fundamental-* libraries so that we stay as consistent as possible.

PR #767, opened 2019-10-14
feat: refactor input group

### LINKED ISSUE #518: Refactor Input Group
### Description
Input Group should be refactored 
* Follow input group pattern from HIG.concur.com
  * Possible subcomponents for different types of input groups. 
* No callbacks needed. 

### Versions
**fundamental-react:** v0.5.1
**fiori-fundamentals:** v1.5.3

---

hertweckhr1:
### Description
Refactored Input Group to be a bit more clean and act as wrapper to form input and input group add on. 
- consumers will now put their own form input and input group addon 

Added Storybook examples.

Deleted callbacks.

Will create follow-up issue to create component for up/down number form input.

fixes #518


src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,79 @@
...
+                className={inputGroupClasses}
+                compact={compact}>
+                {children}
+            </div>
+        );
     }
 }
 
+InputGroup.Addon = InputGroupAddon;
+InputGroup.Input = FormInput;
```
jbadan: Is `InputGroup.Input` necessary? It looks like the classes get added by checking for FormInput so we don't need to rename it here.

src/InputGroup/_InputGroupAddon.js
```diff
@@ -0,0 +1,74 @@
...
+
+InputGroupAddon.propDescriptions = {
+    addon: 'The value of the add-on.',
+    addonClassNames: 'CSS class(es) to add to the addon element.',
+    glyph: 'The icon to include if the addon is an icon. See the icon page for the list of icons.'
+};
+
+export { InputGroupAddon as __InputGroupAddon };
+
+export default withStyles(InputGroupAddon, { cssFile: 'input-group', fonts: true });
```
jbadan: This doesn't need it's own styles/fonts.

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,79 @@
...
 import FormInput from '../Forms/FormInput';
-import Icon from '../Icon/Icon';
+import { INPUT_GROUP_ADDON_TYPES } from '../utils/constants';
+import InputGroupAddon from './_InputGroupAddon';
 import PropTypes from 'prop-types';
 import withStyles from '../utils/WithStyles/WithStyles';
-import { INPUT_GROUP_ADDON_POSITIONS, INPUT_GROUP_TYPES } from '../utils/constants';
 import React, { Component } from 'react';
-
+/*eslint-disable*/
```
jbadan: Can we get rid of this?

src/InputGroup/__stories__/InputGroup.stories.js
```diff
@@ -33,4 +33,67 @@ storiesOf('Components|InputGroup', module)
                 glyph='navigation-down-arrow'
                 option='light' />
         </InputGroup>
```
jbadan: Can you update the other stories for InputGroup so they continue to work?

src/_playground/Routes.js
```diff
@@ -400,12 +400,12 @@ export class Routes extends Component {
...
                         <div className={sideBarClasses}>
                             <div className='frDocs-Search'>
-                                <InputGroup
-                                    glyph='search'
-                                    inputPlaceholder='Search'
-                                    inputProps={{ 'aria-label': 'Search' }}
-                                    inputValue={this.state.query}
-                                    onChange={this.onChangeHandler} />
+                                <InputGroup type='icon'>
+                                    <InputGroup.Addon>$</InputGroup.Addon>
```
jbadan: I think this was just an oversight - can it be changed back to the search sap icon?

greg-a-smith:
@hertweckhr1 The PR title (and subsequent merge commit message) should still be prefixed with either `fix:` or `feat:` (I think `fix:` in this case).  The `BREAKING CHANGE:` is added when merging in the description (not title) field as the first line.  Having the breaking change text in the PR title and merge commit title won't do anything with standard version.

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,79 @@
...
-                const inputNumberClasses = classnames(
-                    inputClasses,
-                    'fd-input--no-number-spinner'
-                );
+        children = React.Children.map(children, (child) => {
+            return React.cloneElement(child, {
+                actions: actions,
+                addon: addon,
+                compact: compact,
+                className: (child.type === FormInput) ? inputClasses : '',
```
greg-a-smith: I wouldn't match on `FormInput`.  I would match on `InputGroupAddon` and then put the `inputClasses` on anything that _didn't_ match that.  Keep in mind, the "input" here also needs to support a `FormTextarea`.

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,79 @@
...
-                        </div>
-                    );
-                }
-            }
-        }
+        return (
+            <div
+                {...props}
+                className={inputGroupClasses}
+                compact={compact}>
```
greg-a-smith: Placing `compact` on a `<div>` is not only not going to work, but will throw a console warning as an unexpected DOM attribute.

src/InputGroup/InputGroup.js
greg-a-smith: Not only is it not necessary, it _shouldn't_ be here.  Keep in mind, the "input" here also needs to support a `FormTextarea`.

src/InputGroup/_InputGroupAddon.js
```diff
@@ -0,0 +1,74 @@
...
+}) => {
+
+    let returnObject = children;
+
+    if (glyph && addon === 'icon') {
+        returnObject = (
+            <Icon
+                disableStyles={disableStyles}
+                glyph={glyph}
+                role='presentation' />
```
greg-a-smith: This _seems_ wrong.  I think the consumer should own what `Icon` or other children are within the `InputGroupAddon`.  The only "extra" thing we should need is knowing when the child is a button (or children are buttons).

src/InputGroup/_InputGroupAddon.js
```diff
@@ -0,0 +1,74 @@
...
+            {returnObject}
+        </span>
+    );
+};
+
+InputGroupAddon.displayName = 'InputGroupAddon';
+
+InputGroupAddon.propTypes = {
+    actions: PropTypes.bool,
+    addon: PropTypes.oneOf(INPUT_GROUP_ADDON_TYPES),
```
greg-a-smith: Might be able to just get away with a boolean `isButton` prop.  There's really nothing special we would do with any other children.

src/InputGroup/InputGroup.Component.js
```diff
@@ -20,70 +20,43 @@ export const InputGroupComponent = () => {
...
+                            </InputGroup>
                         </FormItem>
                     </FormGroup>
                     <br />
                     <FormGroup>
                         <FormLabel>Right Aligned Text Addon</FormLabel>
                         <FormItem>
-                            <InputGroup
-                                addon='€'
-                                addonPos='after'
```
markpalfreeman: I hate weird abbreviations anyway! 🎉

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,79 @@
...
-    handleTextChange = e => {
-        e.preventDefault();
-        this.setState({
-            value: e.target.value
-        });
-    };
-
     render() {
-        const {
+        let {
```
markpalfreeman: Can we change this back so we're not modifying `children` props below?

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,79 @@
...
-            'fd-input-group'
-        );
-
-        const addonClasses = classnames(
-            addonClassNames,
-            'fd-input-group__addon',
-            [{ 'fd-input-group__addon--button': !!actions || inputType === 'number' }]
-        );
-
         const inputClasses = classnames(
```
markpalfreeman: nit: this probably doesn't need to call `classnames` anymore if it's one static string.

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,79 @@
...
-            inputClassName,
             'fd-input-group__input'
         );
 
-        switch (inputType) {
-            case 'number':
+        const inputGroupClasses = classnames(
+            className,
+            'fd-input-group',
+            [{ 'fd-input-group--compact': compact }]
```
markpalfreeman: nit: you shouldn't need the redundant `[]` (just the object should work).

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,79 @@
...
+        );
 
-                const inputNumberClasses = classnames(
-                    inputClasses,
-                    'fd-input--no-number-spinner'
-                );
+        children = React.Children.map(children, (child) => {
```
markpalfreeman: Modifying (resetting) props variables in React components always makes me nervous, even just so we know exactly what we're dealing with with no side effects. Can we either:
1. make a new variable for this to designate it's a newly created map (something like `childrenWithProps`?)
2. just do the `React.children.map()` inline inside the `<div>` (doesn't necessarily have to be a variable up here)

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,79 @@
...
     }
 }
 
+InputGroup.Addon = InputGroupAddon;
+InputGroup.Input = FormInput;
+
 InputGroup.displayName = 'InputGroup';
 
 InputGroup.propTypes = {
     actions: PropTypes.bool,
```
markpalfreeman: @SAP/fundamental-react-team Would anyone be in favor of renaming this while we have a breaking change on our hands?
IMO this would be clearer as `hasActions`; right now it sounds like it should take an array of actions.

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,79 @@
...
+            [{ 'fd-input-group--compact': compact }]
+        );
 
-                const inputNumberClasses = classnames(
-                    inputClasses,
-                    'fd-input--no-number-spinner'
-                );
+        children = React.Children.map(children, (child) => {
+            return React.cloneElement(child, {
+                actions: actions,
```
markpalfreeman: nit: you don't need to list these twice if the key and variable names match:
```
{
  actions,
  addon,
  compact
}
```

src/InputGroup/InputGroup.js
markpalfreeman: OR do we need `actions` at all? Is this always just related to "is this thing a button?" 😛

jbadan: 🔥

hertweckhr1: no actions for the winnnnnn 🏅

hertweckhr1: 🔥

src/InputGroup/InputGroup.test.js
```diff
@@ -215,100 +139,54 @@ describe('<InputGroup />', () => {
...
+            <InputGroup.Input />
+        </InputGroup>);
+    };
 
-            expect(
-                element.getDOMNode().attributes['data-sample'].value
-            ).toBe('Sample');
+    describe('Default Rendering', () => {
+        let element = setup({
+            children: 'foo' //prop children is required
```
jbadan: This seems weird - shouldn't `InputGroup.Addon` and the `FormInput` be the children?

src/InputGroup/InputGroup.test.js
```diff
@@ -215,100 +139,54 @@ describe('<InputGroup />', () => {
...
-            ).toBe('Sample');
+            expect(() => element).not.toThrow();
         });
 
-        test('should allow props to be spread to the InputGroup component for type number\'s down button element', () => {
-            const element = mount(<InputGroup inputType='number' numberDownButtonProps={{ 'data-sample': 'Sample' }} />);
+        it('should allow data attribute to be passed to span element', () => {
+            const element = setup({
+                children: [<InputGroup.Input />, <InputGroup.Addon>@</InputGroup.Addon>],
+                'data-sample': 'Sample'
```
jbadan: The way this is setup, shouldn't this `data-sample` be being passed to the input group, not the addon?

jbadan: ah ok - the more I look at this - the test is right - just the description should be updated.

src/InputGroup/InputGroup.test.js
```diff
@@ -215,100 +139,54 @@ describe('<InputGroup />', () => {
...
-            ).toBe('Sample');
+                element.getDOMNode().className
+            ).toContain('fd-input-group');
         });
+    });
 
-        test('should allow props to be spread to the InputGroup component for type number\'s up button element', () => {
-            const element = mount(<InputGroup inputType='number' numberUpButtonProps={{ 'data-sample': 'Sample' }} />);
+    describe('Rendering with Props', () => {
+        it('should not throw error if child.type is undefined', () => {
```
jbadan: Should we add an additional test for adding the `fd-input-group__input` logic?

src/InputGroup/_InputGroupAddon.js
```diff
@@ -0,0 +1,48 @@
...
+};
+
+InputGroupAddon.displayName = 'InputGroupAddon';
+
+InputGroupAddon.propTypes = {
+    addonClassNames: PropTypes.string,
+    children: PropTypes.node,
+    className: PropTypes.string,
+    compact: PropTypes.bool,
+    customStyles: PropTypes.object,
```
jbadan: since this isn't wrapped with `withStyles`, it shouldn't have `customStyles` or `disableStyles` props as they wouldn't do anything

src/InputGroup/InputGroup.Component.js
```diff
@@ -145,31 +124,29 @@ export const InputGroupComponent = () => {
                 description={`The Input with add-on supports actions. Actions can be shown with a text
                     label or icon.`}
                 title='Input with actions'>
+                {/* ACTIONS NEEDS TO BE FIXED TO BE GIVEN TO CHILD FROM PARENT */}
```
jbadan: Can you remove this comment if it's no longer true?

jbadan:
One more: the styling seems off - looks like we're missing a class on the buttons `fd-input-group__button`

Screen Shot 2019-10-15 at 9 57 24 AM

src/_playground/Routes.js
```diff
@@ -400,12 +402,14 @@ export class Routes extends Component {
                         </header>
                         <div className={sideBarClasses}>
                             <div className='frDocs-Search'>
-                                <InputGroup
-                                    glyph='search'
-                                    inputPlaceholder='Search'
-                                    inputProps={{ 'aria-label': 'Search' }}
-                                    inputValue={this.state.query}
-                                    onChange={this.onChangeHandler} />
+                                <InputGroup type='icon'>
```
greg-a-smith: `InputGroup` doesn't have a `type` prop.

src/_playground/Routes.js
```diff
@@ -400,12 +402,14 @@ export class Routes extends Component {
...
+                                    <InputGroup.Addon>
+                                        <Icon glyph='search' />
+                                    </InputGroup.Addon>
+                                    <FormInput
+                                        onChange={this.onChangeHandler} placeholder='Search'
```
greg-a-smith: Very minor nit:  when stacking props on multiple lines, there should only be one per line -- `placeholder` should be on its own line.

src/InputGroup/InputGroup.js
greg-a-smith: Agreed -- should be `const`.

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,56 @@
...
             className,
-            'fd-input-group'
-        );
-
-        const addonClasses = classnames(
-            addonClassNames,
-            'fd-input-group__addon',
-            [{ 'fd-input-group__addon--button': !!actions || inputType === 'number' }]
+            'fd-input-group',
+            { 'fd-input-group--compact': compact }
```
greg-a-smith: The class `fd-input-group--compact` doesn't exist.  Child components have the "compact" variations.

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,56 @@
...
-        const {
-            actions,
-            addonClassNames,
-            addonPos,
-            addon,
+        let {
             children,
             className,
             compact,
             disableStyles,
```
greg-a-smith: We should prune `customStyles` here as well so it doesn't get spread to the top-level `<div>`.

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,56 @@
...
-            inputClassName,
-            'fd-input-group__input'
+        return (
+            <div
+                {...props}
+                className={inputGroupClasses}>
+                {React.Children.map(children, (child) => (
+                    React.cloneElement(child, {
+                        compact,
+                        className: (child.type === InputGroup.Addon) ? '' : inputClasses
```
greg-a-smith: This comparison might be better/safer using `displayName`.

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,56 @@
...
-    inputPlaceholder: 'Value for the `placeholder` attribute on the `<input>` element.',
-    inputType: 'Value for the `type` attribute on the `<input>` element.',
-    inputValue: 'Value for the `value` attribute on the `<input>` element.',
-    localizedTextShape: {
-        clear: 'Value for aria-label on the clear <button> element.',
-        down: 'Value for aria-label on the down <button> element.',
-        up: 'Value for aria-label on the up <button> element.'
-    },
-    numberDownButtonProps: 'Additional props to be spread to the down `<button>` element (for inputType=\'number\').',
-    numberUpButtonProps: 'Additional props to be spread to the up `<button>` element (for inputType=\'number\').'
```
greg-a-smith: Lots of 🔥 and I love it! 👍

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,56 @@
...
 InputGroup.displayName = 'InputGroup';
 
 InputGroup.propTypes = {
-    actions: PropTypes.bool,
-    addon: PropTypes.string,
-    addonClassNames: PropTypes.string,
-    addonPos: PropTypes.oneOf(INPUT_GROUP_ADDON_POSITIONS),
     children: PropTypes.node,
     className: PropTypes.string,
     compact: PropTypes.bool,
```
greg-a-smith: We have a choice to make regarding `compact` and I think there are merits for each approach.  If we take a `compact` prop here, it is simply to pass along to its children (and subsequently through `InputGroupAddon` to its children) the prop so the proper classes get added to the proper elements.  The other option would be to _not_ take the prop here and just have the consumer pass their `compact` prop to their respective inputs and buttons manually.

src/InputGroup/_InputGroupAddon.js
```diff
@@ -0,0 +1,45 @@
...
+    compact,
+    isButton,
+    ...otherProps
+}) => {
+
+    const addonClasses = classnames(
+        addonClassNames,
+        'fd-input-group__addon',
+        { 'fd-input-group__addon--button': isButton },
+        { 'fd-input-group__addon--button--compact': compact && isButton }
```
greg-a-smith: The class `fd-input-group__addon--button--compact` doesn't exist so this can be removed.  See comment in `InputGroup.js` about how to handle `compact` prop.

src/InputGroup/_InputGroupAddon.js
```diff
@@ -0,0 +1,45 @@
...
+        <span className={addonClasses} {...otherProps} >
+            {children}
+        </span>
+    );
+};
+
+InputGroupAddon.displayName = 'InputGroup.Addon';
+
+InputGroupAddon.propTypes = {
+    addonClassNames: PropTypes.string,
```
greg-a-smith: There no need for `addonClassNames` -- these are for _this_ component so we can just use `className`.

src/InputGroup/_InputGroupAddon.js
```diff
@@ -0,0 +1,45 @@
...
+    children: PropTypes.node,
+    className: PropTypes.string,
+    compact: PropTypes.bool,
```
greg-a-smith: There are no defined CSS classes for a "compact" addon.  This should be passed to its children (if we keep the prop).  See comment in `InputGroup.js` about how to handle `compact` prop.

src/InputGroup/InputGroup.js
hertweckhr1: I've kept the prop for now to get it simple for users based on team feedback, but can be up for further debate.

src/InputGroup/_InputGroupAddon.js
```diff
@@ -0,0 +1,49 @@
...
+const InputGroupAddon = ({
+    children,
+    className,
+    compact,
+    isButton,
+    ...otherProps
+}) => {
+
+    const addonClasses = classnames(
+        'fd-input-group__addon',
```
greg-a-smith: The prop `className` should also be included in here.

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,56 @@
...
-        );
-
-        const inputClasses = classnames(
-            inputClassName,
-            'fd-input-group__input'
+        return (
+            <div
+                {...props}
+                className={inputGroupClasses}>
+                {React.Children.map(children, (child) => (
```
greg-a-smith: To be safe, we should use the `toArray` syntax (like is being used in `InputGroupAddon`).
```jsx
{React.Children.toArray(children).map(child => {
```

src/InputGroup/InputGroup.js
```diff
@@ -1,272 +1,56 @@
...
+                    React.cloneElement(child, {
+                        compact,
+                        className: (child.type.displayname === InputGroup.Addon) ? '' : inputClasses
```
greg-a-smith: I don't believe this comparison will ever match -- `InputGroup.Addon` is a function.  I think it should be this:
```jsx
className: (child.type.displayname === InputGroupAddon.displayName) ? '' : inputClasses
```

src/InputGroup/_InputGroupAddon.js
```diff
@@ -0,0 +1,49 @@
...
+
+InputGroupAddon.propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    compact: PropTypes.bool,
+    isButton: PropTypes.bool
+};
+
+InputGroupAddon.propDescriptions = {
+    isButton: 'Set to **true** if add on is button.'
```
greg-a-smith: Minor nit: put an "a" before "button".

src/InputGroup/_InputGroupAddon.js
```diff
@@ -0,0 +1,49 @@
...
+InputGroupAddon.displayName = 'InputGroup.Addon';
+
+InputGroupAddon.propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    compact: PropTypes.bool,
+    isButton: PropTypes.bool
+};
+
+InputGroupAddon.propDescriptions = {
```
greg-a-smith: I think the `compact` prop description should be overwritten here since we are accepting `compact` at the `InputGroup` level and the consumer really shouldn't be passing it here (although nothing bad would happen if they did).
```js
compact: '_INTERNAL USE ONLY._',
```

src/InputGroup/__stories__/InputGroup.stories.js
```diff
@@ -1,36 +1,111 @@
...
-                compact
-                glyph='navigation-down-arrow'
-                option='light' />
+            <FormInput />
+            <InputGroup.Addon>
+                <Button
+                    compact
+                    glyph='navigation-down-arrow'
+                    option='light' />
+            </InputGroup.Addon>
```
greg-a-smith: A couple things I noticed:

1. When the `displayName` comparison is fixed in `InputGroup`, these will start working better.
1. Examples that are using buttons need to have the `isButton` prop set on `InputGroup.Addon`.
1. The `compact` prop can be removed from the `Button` components.
1. I see a few of the examples using knobs for the `compact` prop on `InputGroup` -- I think all of these should have that.

src/InputGroup/InputGroup.js
hertweckhr1: `className: (child.type.displayname === InputGroupAddon.displayName) ? '' : inputClasses` does not work but className: (child.type.displayName === 'InputGroup.Addon') ? '' : inputClasses does 👍

src/InputGroup/__stories__/InputGroup.stories.js
hertweckhr1: compact has been added to all example for stories, except for textArea.

src/InputGroup/InputGroup.js
hertweckhr1: Sounds good - what the extra benefit is with using `React.Children.toArray(children).map` vs. `React.children.map` in this case?

hertweckhr1:
> Very close. Just one last request. I think we should add additional snapshot testing for `InputGroupAddon`. Something that exercises both the `compact` and `isButton` props.

@greg-a-smith -> can do so for `isButton`.  Since compact is being passed down from inputgroup, isn't this shown in the inputgroup snapshot tests that already include snapshots for compact?

src/InputGroup/InputGroup.js
greg-a-smith: @markpalfreeman There was actually a case that someone found in `TabGroup` where the `React.Children.map` _didn't_ protect against `null` or `undefined` children, which is where this `toArray` pattern was introduced.  Whether it's actually _needed_ here could be up for debate, but I think this is a safe pattern.

src/InputGroup/InputGroup.test.js
```diff
@@ -216,101 +140,69 @@ describe('<InputGroup />', () => {
...
+        return mount(
+            <InputGroup {...props}>
+                <InputGroup.Addon>$</InputGroup.Addon>
+                <FormInput />
+            </InputGroup>);
+    };
+
+    describe('Default Rendering', () => {
+        let element = setup({
+            children: [<FormInput />, <InputGroup.Addon>@</InputGroup.Addon>]
```
jbadan: Is this necessary? Your setup on line 143 already has children in it. I would either edit the setup to not include any default children or edit your other tests to remove this and just call `let element = setup()`

PR #769, opened 2019-10-15
feat: Move FormGroup to forms

hertweckhr1:
### Description
BREAKING CHANGE
As a side story to the breaking change we're making by refactoring inputgroup -> moving over FormGroup to Forms instead of InputGroup.

Used git mv in order to preserve git history.


src/Forms/__stories__/FormGroup.stories.js
```diff
@@ -1,7 +1,7 @@
 import FormGroup from '../FormGroup';
 import FormItem from '../../Forms/FormItem';
 import FormLabel from '../../Forms/FormLabel';
```
greg-a-smith: Minor nit:  both `FormItem` and `FormLabel` could have paths simplified to `../FormItem` or `../FormLabel`, respectively.

src/index.js
```diff
@@ -28,7 +28,7 @@ export { default as Identifier } from './Identifier/Identifier';
 export { default as Image } from './Image/Image';
 export { default as InlineHelp } from './InlineHelp/InlineHelp';
 export { default as InputGroup } from './InputGroup/InputGroup';
-export { default as FormGroup } from './InputGroup/FormGroup';
+export { default as FormGroup } from './Forms/FormGroup';
```
bcullman: Probably worth reordering this list by alpha - it will put `FormGroup` up by the other `Form*` exports.

src/Forms/Forms.Component.js
```diff
@@ -16,42 +16,42 @@ export const FormsComponent = () => {
                     and noted by an asterisk (*).`}
                 title='Inputs'>
                 <div>
-                    <FormSet>
+                    <FormGroup>
```
hertweckhr1: Added FormGroup to an example so that we are using FormGroup in the Forms page

PR #770, opened 2019-10-15
chore: change main file path for bundlephobia

jbadan:
### Description

I believe this change might fix the bundlephobia errors we've been having. Because we `cd lib` then `npm publish`, there is no `lib/index.js`, just `index.js`.

I don't know how to test this besides merging and checking the build in bundlephobia 🤷‍♀

This might also fix the problem  @parostatkiem had in codepen.

fixes #issueid

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": false,
   "description": "SAP Fundamentals, implemented in React",
   "license": "Apache-2.0",
-  "main": "lib/index.js",
+  "main": "index.js",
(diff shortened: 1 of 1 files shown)


markpalfreeman:
I could see how this might fix publishing, but isn't the main entrypoint still `lib/index.js`?

_Needs to study up more on package consumption_

jbadan:
@markpalfreeman I'll revert if it breaks anything :)

PR #772, opened 2019-10-16
fix: create umd build

jbadan:
### Description
- [x] publish umd file so everything works in code sandboxes (see: https://github.com/mui-org/material-ui/blob/master/packages/material-ui/package.json#L28)


markpalfreeman:
I'm curious: do we need to include Rollup _in addition to_ webpack? It seems like we may be creating more overhead just to publish a different lib type.

Can we do this with some changes to our current build config? https://webpack.js.org/guides/author-libraries/#expose-the-library

jbadan:
@markpalfreeman That's a good idea - feel free to hop on this branch and try it out!

scripts/rollup.config.js
```diff
@@ -0,0 +1,78 @@
+import babel from 'rollup-plugin-babel';
```
ChristianKienle: rollup <3

markpalfreeman:
I'm not trying to avoid this; the only reason I haven't approved this is because I just don't fully understand it 🤷‍♂ 

I'd feel more comfortable to merge the `npm publish lib` change and investigate the library files separately... but maybe we can discuss briefly as a team?

PR #789, opened 2019-11-05
fix: Align popover inside dropdown

### LINKED ISSUE #788: Popover is always stretched to the right edge of the viewport
### Description
This happens on the documentation website as well as in my application that uses Fundamental-react.
image

### Versions
**fundamental-styles:** none

---

_**DISCLAIMER:**

parostatkiem:
#### This is a temporary solution.
The whole problem requires a major refactoring on popovers, as @jbadan said on Slack (100% agree).
This is just a "patch" to make Popover working with Dropdown or any other element.
⚠️ It **does not** introduce any backwards-incompatible changes.
### Description
- Add ability to align Popover left and right edge with its parent control element edges. 
- Add ability to set Popover right edge to be not-further-left or not-further-right than parent control right edge.

 It works well with Dropdown as well as any other type. ~Sadly, the solution requires creating a reference on the client side.~

#### Before:
image

#### After:
image

fixes #788

### DIFF
--- a/src/Popover/Popover.js
+++ b/src/Popover/Popover.js
@@ -1,9 +1,9 @@
 import chain from 'chain-function';
 import classnames from 'classnames';
 import Popper from '../utils/_Popper';
-import { POPPER_PLACEMENTS } from '../utils/constants';
 import PropTypes from 'prop-types';
 import withStyles from '../utils/WithStyles/WithStyles';
+import { POPPER_PLACEMENTS, POPPER_SIZING_TYPES, POPPER_SIZING_TYPES_DESCRIPTION } from '../utils/constants';
 import React, { Component } from 'react';
 
 class Popover extends Component {
@@ -13,6 +13,7 @@ class Popover extends Component {
         this.state = {
             isExpanded: false
         };
+        this.placementTargetRef = React.createRef();
     }
 
     triggerBody = () => {
@@ -44,7 +45,7 @@ class Popover extends Component {
             className,
             placement,
             popperProps,
-            parentRef,
+            widthSizingType,
             ...rest
         } = this.props;
 
@@ -54,8 +55,10 @@ class Popover extends Component {
         }
 
         const referenceComponent = React.cloneElement(control, {
-            onClick: onClickFunctions
+            onClick: onClickFunctions,
+            ref: this.placementTargetRef
         });
+
         const popoverClasses = classnames('fd-popover', className);
 
         return (
@@ -66,13 +69,14 @@ class Popover extends Component {
                     noArrow={noArrow}
                     onClickOutside={chain(this.handleOutsideClick, onClickOutside)}
                     onEscapeKey={chain(this.handleOutsideClick, onEscapeKey)}
-                    parentRef={parentRef}
+                    placementTargetRef={this.placementTargetRef}
                     popperPlacement={placement}
                     popperProps={popperProps}
                     referenceClassName='fd-popover__control'
                     referenceComponent={referenceComponent}
                     show={this.state.isExpanded && !disabled}
-                    usePortal>
+                    usePortal
+                    widthSizingType={widthSizingType}>
                     {body}
                 </Popper>
             </div>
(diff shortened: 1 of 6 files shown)


src/Popover/Popover.js
```diff
@@ -90,22 +91,24 @@ Popover.propTypes = {
...
+    onClickOutside: () => { },
+    onEscapeKey: () => { }
 };
 
 Popover.propDescriptions = {
     body: 'Node(s) to render in the overlay.',
     control: 'Node to render as the reference element (that the `body` will be placed in relation to).',
     disableEdgeDetection: 'Set to **true** to render popover without edge detection so popover will not flip from top to bottom with scroll.',
     noArrow: 'Set to **true** to render a popover without an arrow.',
+    parentElement: 'A reference (useRef) to an element of which the Popover will be positioned and sized relatively. Leave it empty to let it be done automatically.',
```
markpalfreeman: Can we call this `parentRef` to be even more explicit?

And if we're going to mention `useRef()`, can we clarify the function itself shouldn't be passed in, but rather the return object from calling it?

src/utils/_Popper.js
```diff
@@ -94,6 +95,11 @@ class Popper extends React.Component {
                     if (!show) {
                         return null;
                     }
+                    const currentParentElement = parentElement && parentElement.current;
```
markpalfreeman: Regarding my other comment, `parentRef` would be _slightly_ clearer here that we could access a property `.current`.

src/Popover/Popover.js
markpalfreeman: Also, the description seems to imply ("let it be done") that positioning and sizing **relative to its parent** _can_ happen automatically if left empty. Is that the case?

bcullman:
@parostatkiem - can you add before and after screen caps?

src/Popover/Popover.js
parostatkiem: Sadly, it's not. If it was, this PR wouldn't need to exist at all ;)
I had no idea how to let the developer know he should put useRef's _result_ there... Your idea of `parentRef` seems reasonable.

I tried to put some comments into the code displayed in the documentation website but it's impossible :(

parostatkiem:
@bcullman you can clearly see it live on the deployed preview.
https://deploy-preview-789--fundamental-react.netlify.com/dropdown

Open the _Width limited to parent's_ dropdown, then open any other one. You'll see the difference :)

parostatkiem:
@jbadan ✅done

bcullman:
@parostatkiem - I could imagine a better long term fix would be to address this here by adding a opt-in prop for `widthSizingType' here:

https://github.com/SAP/fundamental-react/blob/master/src/utils/_Popper.js

that accepted the following 

    ‘none’,
   ‘matchTarget’,
   ‘minTarget’,
   ‘maxTarget’

it would fix this component, and be flexible enough to handle all other circumstances, including other copmponets relying on `_popper.js`

parostatkiem:
@bcullman - where it would take the target's width from? The problem is Popper has no "idea" of his parent. That's why I introduced `parentRef` which has an obvious disadvantage of being defined on the client side.

... or do you mean adding `widthSizingType` **on top of** my solution?

greg-a-smith:
@parostatkiem The `_Popper.js` component has a prop called `referenceComponent`, which is the element to be sized in relation to.

parostatkiem:
@greg-a-smith it was the first place I tried to find a solution at but I couldn't get the real DOM node of the `referenceComponent`, so I abandoned the idea and did it the other way.
However, today you inspired me to give it another chance and... here it is 🍾
Thank you for the inspiration!

@bcullman I implemented your solution in an unchanged form. I finally got rid of a ref on the client side :)

jbadan:
@parostatkiem I tried to resolve some conflicts for you here, but I can't push to your branch to fix the failing test. Ping the team when the tests pass and we will merge :)

parostatkiem:
@jbadan I added you to collaborators, you should receive an invitation email.
I'm quite busy this night (it's midnight right now) and will be able to fix the tests on Monday at the earliest.

jbadan:
closing for #804

PR #790, opened 2019-11-13
fix: No Popover Arrow

### LINKED ISSUE #740: Bug: Arrow missing on Popover
### Description
The arrow on Popover is "missing" (the element is present, but not displaying an arrow)

### Versions
Current, as it is missing even in the docs
https://sap.github.io/fundamental-react/popover

prsdthkr:
### Description

Cause: Mismatch between FD-style expected attribute name and FD-react property applied leading to appropriate styles not being applied to the popover.

Solution: Change popover `<div>`'s property names to match FD-style attribute selectors seen [here](https://github.com/SAP/fundamental-styles/blob/master/src/popover.scss#L210). 
* `data-placement` becomes `x-placement`
* `data-x-out-of-boundaries` becomes `x-out-of-boundaries`
* Storybook entry for popover placement [added](http://localhost:12123/?path=/story/components-popover--placement)

**Before** (chrome)
image

**After** (chrome)
image

fixes #740

### DIFF
--- a/src/Popover/__stories__/Popover.stories.js
+++ b/src/Popover/__stories__/Popover.stories.js
@@ -31,24 +31,39 @@ storiesOf('Components|Popover', module)
         <>
             <Popover
                 body={someText}
+                control={<Button glyph='navigation-up-arrow' option='light' >Pop to Top-Start</Button>}
+                placement='top-start'
+                style={{ position: 'fixed', top: '10em', left: '35em' }} />
+            <Popover
+                body={someText}
                 control={<Button glyph='navigation-up-arrow' option='light' >Pop to Top</Button>}
                 placement='top'
                 style={{ position: 'fixed', top: '10em', left: '50em' }} />
             <Popover
                 body={someText}
+                control={<Button glyph='navigation-up-arrow' option='light' >Pop to Top-End</Button>}
+                placement='top-end'
+                style={{ position: 'fixed', top: '10em', left: '65em' }} />
+            <Popover
+                body={someText}
                 control={<Button glyph='navigation-right-arrow' option='light' >Pop to Right-Top</Button>}
                 placement='right-start'
-                style={{ position: 'fixed', top: '15em', left: '60em' }} />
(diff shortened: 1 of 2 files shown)


src/Popover/__stories__/Popover.stories.js
```diff
@@ -18,13 +18,59 @@ const bodyContent = (
...
+                control={<Button glyph='navigation-left-arrow' option='light' >Pop to Left</Button>}
+                placement='left'
+                style={{ position: 'fixed', top: '20em', left: '40em' }} />
+            <Popover
+                body={someText}
+                control={<Button glyph='navigation-left-arrow' option='light' >Pop to Left-Start</Button>}
+                placement='left-start'
+                style={{ position: 'fixed', top: '15em', left: '40em' }} />
+        </>
+    ))
```
greg-a-smith: Since you're making examples of each placement, we should probably also add `top-start`, `top-end`, `bottom-start` and `bottom-end`.

prsdthkr: Done. IDK how I missed these. Thanks!

src/Popover/__stories__/Popover.stories.js
```diff
@@ -18,13 +18,79 @@ const bodyContent = (
...
             body={bodyContent}
             control={<Button glyph='navigation-up-arrow' option='light' />} />
     ))
+    .add('Placement', () => (
+        <>
+            <Popover
+                body={someText}
+                control={<Button glyph='navigation-up-arrow' option='light' >Pop to Top-Start</Button>}
+                placement='top-start'
+                style={{ position: 'fixed', top: '10em', left: '35em' }} />
```
greg-a-smith: The placement `style` for each of these is strange.  When I have the inspector open, I am not seeing the bottom row and no scrollbar appears.  How about building a table that is centered?

| Layout | | | | |
|--|--|--|--|--|
|  |  Top Start | Top | Top End |  |
| Left Start | | | | Right Start |
| Left | | | | Right |
| Left End | | | | Right End |
| | Bottom Start | Bottom | Bottom End | |

PR #792, opened 2019-11-14
feat: add accessibility features to popover control

meganaconley:
### Description

Add accessibility features to the `Popover` `control` prop. This way, any element can be passed as a control, but it will still work as a button.

* `tabIndex`: puts normally non-tabbable elements into the normal tab flow
* `aria-haspopup`: indicates the button triggers a popup
* `role="button"`: screenreader will announce element as button
* `onKeyPress` handler: if the `control` is not a button, adds `enter` and `space` as a way to trigger the popover.

### DIFF
--- a/src/ComboboxInput/__snapshots__/ComboboxInput.test.js.snap
+++ b/src/ComboboxInput/__snapshots__/ComboboxInput.test.js.snap
@@ -14,12 +14,8 @@ exports[`<ComboboxInput /> create combobox input 1`] = `
         className="fd-popover__control"
       >
         <div
-          aria-haspopup={true}
           className="fd-input-group"
           onClick={[Function]}
-          onKeyPress={[Function]}
-          role="button"
-          tabIndex={0}
         >
           <input
             className="fd-input fd-input-group__input"
@@ -56,12 +52,8 @@ exports[`<ComboboxInput /> create combobox input 2`] = `
         className="fd-popover__control"
       >
         <div
-          aria-haspopup={true}
           className="fd-input-group"
           onClick={[Function]}
-          onKeyPress={[Function]}
-          role="button"
-          tabIndex={0}
         >
           <input
             className="fd-input fd-input--compact fd-input-group__input"
--- a/src/ComboboxInput/__snapshots__/ComboboxInput.test.js.snap
+++ b/src/ComboboxInput/__snapshots__/ComboboxInput.test.js.snap
@@ -14,8 +14,12 @@ exports[`<ComboboxInput /> create combobox input 1`] = `
         className="fd-popover__control"
       >
         <div
+          aria-haspopup={true}
           className="fd-input-group"
           onClick={[Function]}
+          onKeyPress={[Function]}
+          role="button"
+          tabIndex={0}
         >
           <input
             className="fd-input fd-input-group__input"
@@ -52,8 +56,12 @@ exports[`<ComboboxInput /> create combobox input 2`] = `
         className="fd-popover__control"
       >
         <div
+          aria-haspopup={true}
           className="fd-input-group"
           onClick={[Function]}
+          onKeyPress={[Function]}
+          role="button"
+          tabIndex={0}
         >
           <input
             className="fd-input fd-input--compact fd-input-group__input"
--- a/src/LocalizationEditor/__snapshots__/LocalizationEditor.test.js.snap
+++ b/src/LocalizationEditor/__snapshots__/LocalizationEditor.test.js.snap
@@ -22,8 +22,12 @@ exports[`<LocalizationEditor /> create localization editor 1`] = `
         className="fd-popover__control"
       >
         <div
+          aria-haspopup={true}
           className="fd-input-group fd-input-group--after"
           onClick={[Function]}
+          onKeyPress={[Function]}
+          role="button"
+          tabIndex={0}
         >
           <input
             className="fd-input fd-input-group__input"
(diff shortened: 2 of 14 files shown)


src/Popover/Popover.js
```diff
@@ -53,7 +79,11 @@ class Popover extends Component {
         }
 
         const referenceComponent = React.cloneElement(control, {
-            onClick: onClickFunctions
+            onClick: onClickFunctions,
+            tabIndex: 0,
+            role: 'button',
+            'aria-haspopup': true,
+            onKeyPress: (event) => this.handleKeyPress(event, control, onClickFunctions)
```
prsdthkr: I'm working on an additional fix for the popover target that adds `aria-haspopup='dialog' or 'grid' or 'listbox' or 'menu' or 'tree'` based on a new `type` property for the Popover component. `type` will default to `true`. This is to align with with [this spec](https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup). It also sets `aria-expanded` to `true` or `false` based on state. It also sets appropriate value for `aria-controls`.

I will open a PR with those changes after this is merged as it essentially builds on top of your work.

src/Shellbar/__snapshots__/Shellbar.test.js.snap
```diff
@@ -42,9 +42,12 @@ exports[`<Shellbar /> create shellbar 1`] = `
               className="fd-popover__control"
             >
               <span
+                aria-haspopup={true}
                 className="fd-identifier fd-identifier--xs fd-identifier--circle fd-has-background-color-accent-8"
                 onClick={[Function]}
+                onKeyPress={[Function]}
                 role=""
```
prsdthkr: This `span` should have `role=button`, but that's an issue with the `Identifier` component used in the `ShellBar`. It resets the role. Hence VoiceOver on Chrome doesn't announce the `Identifier` as a pop-up button.

src/SearchInput/__snapshots__/SearchInput.test.js.snap
```diff
@@ -16,8 +16,12 @@ exports[`<SearchInput /> create SearchInput 1`] = `
         className="fd-popover__control"
       >
         <div
+          aria-haspopup={true}
           className="fd-input-group fd-input-group--after"
           onClick={[Function]}
+          onKeyPress={[Function]}
```
prsdthkr: Similar behavior as `ComboboxInput` where `space` key presses are consumed to open the popup.

src/LocalizationEditor/__snapshots__/LocalizationEditor.test.js.snap
```diff
@@ -22,8 +22,12 @@ exports[`<LocalizationEditor /> create localization editor 1`] = `
         className="fd-popover__control"
       >
         <div
+          aria-haspopup={true}
           className="fd-input-group fd-input-group--after"
           onClick={[Function]}
+          onKeyPress={[Function]}
+          role="button"
+          tabIndex={0}
         >
```
prsdthkr: Similar behavior as `ComboboxInput` where `space` key presses are consumed to open the popup.

src/ComboboxInput/__snapshots__/ComboboxInput.test.js.snap
```diff
@@ -14,8 +14,12 @@ exports[`<ComboboxInput /> create combobox input 1`] = `
         className="fd-popover__control"
       >
         <div
+          aria-haspopup={true}
           className="fd-input-group"
           onClick={[Function]}
+          onKeyPress={[Function]}
+          role="button"
```
prsdthkr: `space` keypress is consumed by this `div` wrapper and hence it's not possible to type out an actual `space` in the `input` below. The dropdown `button` also behaves weirdly when we press `space`.

jbadan:
Once the Popover is open - how do I navigate inside of it with keyboard? Or is that out of scope for this pr.
I forgot to turn keycastr on but I tabbed to the button dropdown and then was using the arrow keys - not the mouse like it looks. Bad gif sorry!

jacobdevera:
@jbadan keyboard navigation for Popover is through tabbing in https://github.com/SAP/fundamental-react/pull/793

I would think that a basic Popover should navigate via tabbing since it can contain any content, but it makes sense that certain compositions (Dropdown, ComboboxInput, etc.) would use arrow keys. Probably out of scope for now.

src/Popover/Popover.js
```diff
@@ -78,13 +79,22 @@ class Popover extends Component {
...
-            role: 'button',
-            'aria-haspopup': true,
-            onKeyPress: (event) => this.handleKeyPress(event, control, onClickFunctions)
-        });
+        let controlProps = {
+            onClick: onClickFunctions
+        };
+
+        if (!disableKeyPressHandler) {
+            controlProps = {
```
prsdthkr: Is it unsafe to add  `tabindex:0` and `aria-haspopup:true` when `disableKeyPressHandler` is `false/unset`?

src/Popover/Popover.js
```diff
@@ -78,13 +79,22 @@ class Popover extends Component {
...
+                ...controlProps,
+                tabIndex: 0,
+                role: 'button',
+                ariaRole: 'button',
```
prsdthkr: I'm not sure if we need `aria-role` at all if `role` is present. Also, React converts `ariaRole` to `ariarole`.

src/Popover/Popover.js
meganaconley: I think it may be, as these controls are compound components (such as ComboboxInput where the input is both a FormInput and a Button combined). None of the components I gave this prop to use anything besides a Button to trigger the Popover, so they're still accessible.

src/Popover/Popover.js
meganaconley: I've fixed the role handling.

PR #796, opened 2019-11-20
fix: Add aria attributes to popover control

### LINKED ISSUE #791: Bug: Popover control should have aria attributes for accessibility
### Description
The popover control is not being announced as such by screen readers.
To fix this, the popover control (the trigger) should have following aria attributes:
* [aria-controls](https://www.w3.org/TR/wai-aria-1.1/#aria-controls)
* [aria-expanded](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded)
* [aria-haspopup](https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup)

VoiceOver on Chrome just announces 'button' without indication if it has an associated popover or not and if that popover is open or close:
ACC-1424_no_aria

### Versions
**fundamental-react:** Current
**fundamental-styles:** N/A

prsdthkr:
### Description
Add following aria attributes to popover control/target.
* [aria-controls](https://www.w3.org/TR/wai-aria-1.1/#aria-controls)
* [aria-expanded](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded)
* [aria-haspopup](https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup)

This allows screen readers and other assistive technologies to understand the popover interactions semantically.

fixes #791

### DIFF
--- a/src/Popover/Popover.js
+++ b/src/Popover/Popover.js
@@ -2,9 +2,10 @@ import chain from 'chain-function';
 import classnames from 'classnames';
 import keycode from 'keycode';
 import Popper from '../utils/_Popper';
-import { POPPER_PLACEMENTS } from '../utils/constants';
 import PropTypes from 'prop-types';
+import shortId from '../utils/shortId';
 import withStyles from '../utils/WithStyles/WithStyles';
+import { POPOVER_TYPES, POPPER_PLACEMENTS } from '../utils/constants';
 import React, { Component } from 'react';
 
 class Popover extends Component {
@@ -14,6 +15,10 @@ class Popover extends Component {
         this.state = {
             isExpanded: false
         };
+
+        //A generated shortId as fallback, incase props.popperProps.id is unset.
+        //This ID binds the popover and it's control by 'aria-controls'.
+        this.ariaControls = shortId.generate();
     }
 
     isButton = (node) => {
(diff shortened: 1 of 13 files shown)


src/Popover/Popover.js
```diff
@@ -79,6 +85,12 @@ class Popover extends Component {
             onClickFunctions = chain(this.triggerBody, control.props.onClick);
         }
 
+        let newPopperProps = !!popperProps ? popperProps : {};
+        // eslint-disable-next-line no-undefined
+        if (newPopperProps.id === undefined || newPopperProps.id === null) {
+            Object.assign(newPopperProps, { id: this.ariaControls });
+        }
```
markpalfreeman: I wonder if this could be simplified and avoid the safe checking by:
1) Setting `defaultProps` for `popperProps = {}`
2) Just assigning an `id` inline?
```js
const id = popperProps.id || this.ariaControls;
...
popperProps={{ ...popperProps, id }}
```

src/Popover/Popover.js
```diff
@@ -88,7 +100,9 @@ class Popover extends Component {
                 ...controlProps,
                 tabIndex: 0,
                 role: 'button',
-                'aria-haspopup': true,
+                'aria-controls': newPopperProps.id,
+                'aria-expanded': this.state.isExpanded,
+                'aria-haspopup': !!type ? type : true,
```
markpalfreeman: Is this right? The positive/first ternary return type seems like it could end up being a string.
I imagine the intention is a boolean, but when should this be `false`? (there should probably be a test for this)

src/Popover/Popover.js
```diff
@@ -14,6 +15,10 @@ class Popover extends Component {
         this.state = {
             isExpanded: false
         };
+
+        //A generated shortId as fallback, incase props.popperProps.id is unset.
+        //This ID binds the popover and it's control by 'aria-controls'.
+        this.ariaControls = shortId.generate();
```
meganaconley: `ariaControls` is a bit confusing - maybe `controlId` would be better?

prsdthkr: Since it is not the control's ID how about `popoverId` instead? or `popoverContainerId`?

markpalfreeman: Yeah, I think `popoverId` makes sense. It's a bit difficult to distinguish between that and `popperProps.id`, but I think it's clear enough that one comes from props and the other is defined here as a fallback.

PR #820, opened 2019-12-10
chore: lint all JS

markpalfreeman:
### Description

I _nearly_ got rid of the whole `.eslintignore` file so we don't need to keep adding to it, except it ignores dot-directories by default (needed to add by `.storybook/`).

This also reduces lint scripts significantly; I don't see the value in linting all the folders individually 🤷‍♂


package.json
```diff
@@ -15,30 +15,25 @@
...
     "build:cjs": "cross-env NODE_ENV=production BABEL_ENV=cjs babel src --out-dir lib --ignore \"src/**/*.spec.js\",\"src/**/*.test.js\",\"src/**/*.Component.js\",\"src/_playground/*\",\"src/**/*.stories.js\"",
     "build": "npm run build:index && rm -rf lib && npm run build:cjs",
-    "config:lint": "eslint 'config/**' --ext .js,.jsx --env browser,node",
     "deploy": "gh-pages -d build",
-    "devtools:lint": "eslint 'devtools/**' --ext .js,.jsx --env browser,node",
     "docs:dev": "FUNDAMENTAL_REACT_PLAYGROUND=true npm start",
     "dry-run": "npm run build && npm run build:copy-files && npm publish lib --dry-run",
-    "lint:fix": "npm run build:lint:fix  && npm run scripts:lint:fix && npm run config:lint:fix",
+    "lint": "npm run lint:code && npm run lint:style",
+    "lint:code": "eslint . --ext .js,.jsx",
```
greg-a-smith: Does this need the `--env browser,node` argument?

package.json
```diff
@@ -15,30 +15,25 @@
...
+    "lint:style": "stylelint 'src/**/*.scss' 'src/**/*.css'",
+    "lint:fix": "npm run lint:code -- --fix && npm run lint:style -- --fix",
```
greg-a-smith: Maybe add a `lint:code:fix` and `lint:style:fix` so they could be run independently?  And then have this command run both of those.

markpalfreeman: I can do that, but does anyone actually run those? 😛

package.json
markpalfreeman: It's already in `.eslintrc`: https://github.com/SAP/fundamental-react/blob/master/.eslintrc#L1

skvale:
@greg-a-smith 
@jbadan 
I moved this PR to https://github.com/SAP/fundamental-react/pull/847

PR #832, opened 2019-12-31
fix: handling of `false` checked in Checkbox

sjanota:
I want to use in a controlled manner. If the checked value in Checkbox is initially set to false then defaultChecked is used instead. But as I want it to be a controlled component I never set the default value and undefined gets passed to underlying input. Because of that react assumes it's an uncontrolled component. When a user toggles the checkbox to checked an error is logged:
```
index.js:1 Warning: A component is changing an uncontrolled input of type checkbox to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
    in input (created by Checkbox)
    in label (created by FormLabel)
    in FormLabel (created by FormLabel)
    in FormLabel (created by FormLabel)
    in FormLabel (created by Checkbox)
    in div (created by FormItem)
...
```

Becuase checked is a boolean we need to explicitly check for undefined here.

### Description

fixes #issueid


jbadan:
@sjanota We had a publishing issue and removed a commit from master, can you remove commit `01a4c6c` from this pr.

PR #824, opened 2019-12-20
fix: allow adding classes to both modal body and backdrop

jbadan:
### Description

similar to https://github.com/SAP/fundamental-react/pull/816, which was removed due to a publishing conflict.

fixes #800

PR #845, opened 2020-01-06
feat: allow adding classes to both modal body (removing extraneous div) and backdrop 

### LINKED ISSUE #800: Modal does not honor provided CSS class in "className" property
### Description
CSS class provided in Modal "className" property does not have any effect.

Currently bypassing this issue by directly manipulating underlying "fundamental-style" CSS property of Modal.

`.fd-modal{
    max-width: none!important;
}
.modal-demo-bg{
    width: 90%;
}`

### Versions
**fundamental-react:** 0.7.1
**fundamental-styles:** 0.3.0

jbadan:
### Description

Copy of https://github.com/SAP/fundamental-react/pull/824, removing commit `01a4c6c` that causes publishing issues.

fixes #800


src/Modal/Modal.js
```diff
@@ -51,9 +52,14 @@ class Modal extends Component {
...
 
-        const modalClasses = classnames(
+        const backdropClasses = classnames(
             'fd-overlay',
             'fd-overlay--modal',
+            backdropClassName
+        );
+
+        const modalClasses = classnames(
+            'modal-demo-bg',
```
greg-a-smith: I know this class was here before, but should it be?  This seems like something used as a selector for unit tests and there is an entry in `customstyles.css` from `fundamental-styles`.  It's fine to use some custom styles for a playground or such, but it shouldn't be part of the base component's classes.

jbadan: Good catch! I'll remove it

jbadan: @greg-a-smith turns out that entire div was just for the styles docs site. I'm removing it here and going to make a pr there to make that more obvious.

greg-a-smith:
@jbadan Although this _is_ a feature, should the prefix be `fix:` since we are still on version 0?

jbadan:
I did `feat` because I think this technically a breaking change since I removed a div? What do you think?

greg-a-smith:
@jbadan Yeah, fair enough.  I forgot about that. 👍

Issue #821, opened 2019-12-15
Create Notification component


greg-a-smith:
### Description
In `fundamental-styles@0.3.0-rc.9`, a new component style was added (`fd-notification`).  There should be a component in `fundamental-react` to consume these styles.

### Versions
**fundamental-react:** Latest
**fundamental-styles:** 0.3.0-rc.9 (and above)

PR #846, opened 2020-01-08
feat: update fundamental-styles to 0.3.0

meganaconley:
### Description

[BREAKING CHANGE]

First in a series of PRs to catch up fundamental-styles. The primary changes in this set that are relevant to fundamental-react (up to 0.3.0) were markup changes to the Menu and Shellbar ProductSwitch components. 

1. Menu
- `separator` prop and `<hr>` removed from `MenuItem`
- `separated` prop added to `MenuList`
- `addonBefore` prop passed down into `MenuItem` in order to render `addon` span
- `addon` node changed to `span` and is now wrapped within the `a`

2. Shellbar
- `ProductSwitcher` renamed to `ProductSwitch`
- Import `ProductSwitch` css file
- Removed `nav` node from `ProductSwitch`
- Updated classnames in `ProductSwitch`
- Use glyph in `ProductList` instead of images

Known issues
- `ProductSwitch` button in `Shellbar` has a white background. Further PRs to catch up on `Button` classname changes will fix the issue.

A new component was added which is out of scope for this PR. Issue #821 has already been created for the new Notification component.

https://github.com/SAP/fundamental-styles/blob/master/CHANGELOG.md

Two more PRs are planned to bring it up to 0.3.1 and then 0.4.0-rc.3 to unblock calendar a11y work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "@babel/runtime": "^7.6.3",
     "chain-function": "^1.0.1",
     "classnames": "^2.2.6",
-    "fundamental-styles": "^0.3.0-rc.16",
+    "fundamental-styles": "^0.3.0",
     "hoist-non-react-statics": "^3.3.0",
     "keycode": "^2.2.0",
     "moment": "^2.24.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fundamental-react",
-  "version": "0.8.0-rc.9",
+  "version": "0.8.0-rc.10",
   "private": false,
   "description": "SAP Fundamental Library, implemented in React",
   "license": "Apache-2.0",
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "@babel/runtime": "^7.6.3",
     "chain-function": "^1.0.1",
     "classnames": "^2.2.6",
-    "fundamental-styles": "^0.3.0",
+    "fundamental-styles": "~0.3.0",
     "hoist-non-react-statics": "^3.3.0",
     "keycode": "^2.2.0",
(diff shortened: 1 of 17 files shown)


src/Menu/_MenuList.js
```diff
@@ -2,20 +2,33 @@ import classnames from 'classnames';
...
+                });
+            })}
+        </ul>
+    );
 };
 
 MenuList.displayName = 'Menu.List';
 
 MenuList.propTypes = {
+    addonBefore: PropTypes.bool,
```
jbadan: Can you add `separated` and `addonBefore` to the prop description list.

src/Shellbar/Shellbar.js
```diff
@@ -479,8 +482,8 @@ Shellbar.propDescriptions = {
     },
     notifications: 'Information about pending notifications.',
     productMenu: 'Holds product titles and navigation.',
-    productSwitcher: 'For navigating between products.',
-    productSwitcherList: 'List of the products.',
+    productSwitch: 'For navigating between products.',
+    productSwitchList: 'List of the products.',
```
jbadan: Should this be more descriptive - that it's an array of objects? Name the key options might be a good idea and that title and callback are required with subtitle being optional.

package.json
```diff
@@ -52,7 +52,7 @@
     "@babel/runtime": "^7.6.3",
     "chain-function": "^1.0.1",
     "classnames": "^2.2.6",
-    "fundamental-styles": "~0.3.0-rc.4",
+    "fundamental-styles": "^0.3.0",
```
jbadan: Can you change the `^` to a `~` so we don't accidentally take breaking changes here?

PR #847, opened 2020-01-10
chore: lint all styles and js

skvale:
Migrated from @markpalfreeman 's PR https://github.com/SAP/fundamental-react/pull/820
### Description

I _nearly_ got rid of the whole `.eslintignore` file so we don't need to keep adding to it, except it ignores dot-directories by default (needed to add by `.storybook/`).

This also reduces lint scripts significantly; I don't see the value in linting all the folders individually 🤷‍♂


src/_playground/_content.scss
```diff
@@ -58,7 +58,7 @@
     }
 
     &__tile-background {
-        background: #F3F4F5;
+        background: #f3f4f5;
```
bcullman: switching to lowercase for HEX looks weird to me.

skvale: It's easy enough to change.  I suppose it would make sense to keep the stylelint consistent with [fundamental-styles](https://github.com/SAP/fundamental-styles/blob/master/.stylelintrc.json)
They [use upper case there](https://github.com/SAP/fundamental-styles/blob/fac41376f8eba6c4e55f884172c754886f76adb6/docs/_sass/_docs-utils.scss#L4)

bcullman:
@skvale - good point. let's go ahead and merge.

PR #849, opened 2020-01-15
fix: Do not use defaultChecked if checked is false in Checkbox

skvale:
@sjanota Thanks for the PR!  

I opened this one to get past the blocker in https://github.com/SAP/fundamental-react/pull/832 

### Description
Copied from https://github.com/SAP/fundamental-react/pull/832 

> I want to use in a controlled manner. If the checked value in Checkbox is initially set to false then defaultChecked is used instead. But as I want it to be a controlled component I never set the default value and undefined gets passed to underlying input. Because of that react assumes it's an uncontrolled component. When a user toggles the checkbox to checked an error is logged:
```
index.js:1 Warning: A component is changing an uncontrolled input of type checkbox to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
    in input (created by Checkbox)
    in label (created by FormLabel)
    in FormLabel (created by FormLabel)
    in FormLabel (created by FormLabel)
    in FormLabel (created by Checkbox)
    in div (created by FormItem)
...
```

> Becuase checked is a boolean we need to explicitly check for undefined here.

### DIFF
--- a/src/Forms/Checkbox.js
+++ b/src/Forms/Checkbox.js
@@ -41,7 +41,7 @@ const Checkbox = React.forwardRef(({ checked, className, defaultChecked, disable
                 <input
                     {...inputProps}
                     aria-checked={getCheckStatus(checked, indeterminate)}
-                    checked={checked || defaultChecked}
+                    checked={typeof checked === 'undefined' ? defaultChecked : checked}
                     className={classes}
                     disabled={disabled}
                     id={id}
--- a/src/Forms/Checkbox.js
+++ b/src/Forms/Checkbox.js
@@ -28,6 +28,8 @@ const Checkbox = React.forwardRef(({ checked, className, defaultChecked, disable
         'fd-checkbox'
     );
 
+    const isChecked = typeof checked === 'undefined' ? defaultChecked : checked;
+
     return (
         <FormItem
             {...props}
@@ -40,8 +42,8 @@ const Checkbox = React.forwardRef(({ checked, className, defaultChecked, disable
                 disabled={disabled}>
                 <input
                     {...inputProps}
-                    aria-checked={getCheckStatus(checked, indeterminate)}
-                    checked={typeof checked === 'undefined' ? defaultChecked : checked}
+                    aria-checked={getCheckStatus(isChecked, indeterminate)}
+                    checked={isChecked}
                     className={classes}
                     disabled={disabled}
                     id={id}
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,3 @@
+build
+lib
 !/.storybook
--- a/src/Forms/Checkbox.test.js
+++ b/src/Forms/Checkbox.test.js
@@ -25,7 +25,26 @@ describe('<Checkbox />', () => {
                 onChange: () => {}
             });
 
-            expect(element.props().checked).toBe(true);
+            expect(element.find('input').props().checked).toBe(true);
+        });
+
+        test('should add checked attribute of false to override the defaultChecked', () => {
+            let element = setup({
+                checked: false,
+                defaultChecked: true,
+                onChange: () => {}
+            });
+
+            expect(element.find('input').props().checked).toBe(false);
+        });
+
+        test('should add checked attribute if defaultChecked is true', () => {
+            let element = setup({
+                defaultChecked: true,
+                onChange: () => {}
+            });
+
+            expect(element.find('input').props().checked).toBe(true);
         });
 
         test('should set disabled to true when passed', () => {
@@ -33,7 +52,7 @@ describe('<Checkbox />', () => {
                 disabled: true
             });
 
-            expect(element.props().disabled).toBe(true);
+            expect(element.find('input').props().disabled).toBe(true);
         });
 
         test('should add inline class when inline is passed', () => {
--- a/src/Forms/Checkbox.test.js
+++ b/src/Forms/Checkbox.test.js
@@ -26,6 +26,7 @@ describe('<Checkbox />', () => {
             });
 
             expect(element.find('input').props().checked).toBe(true);
+            expect(element.find('input').props()['aria-checked']).toBe('true');
         });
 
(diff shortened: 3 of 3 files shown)


src/Forms/Checkbox.js
```diff
@@ -41,7 +41,7 @@ const Checkbox = React.forwardRef(({ checked, className, defaultChecked, disable
                 <input
                     {...inputProps}
                     aria-checked={getCheckStatus(checked, indeterminate)}
-                    checked={checked || defaultChecked}
+                    checked={typeof checked === 'undefined' ? defaultChecked : checked}
```
bcullman: Weird we don't have a test for indeterminate.

skvale: Are you talking about `aria-checked={getCheckStatus(checked, indeterminate)}`?
```js
const isChecked = {typeof checked === 'undefined' ? defaultChecked : checked;
// ...
aria-checked={getCheckStatus(isChecked, indeterminate)}
```

skvale: I added a test for indeterminate and others for `aria-checked`

src/Forms/Checkbox.js
```diff
@@ -40,8 +42,8 @@ const Checkbox = React.forwardRef(({ checked, className, defaultChecked, disable
                 disabled={disabled}>
                 <input
                     {...inputProps}
-                    aria-checked={getCheckStatus(checked, indeterminate)}
-                    checked={checked || defaultChecked}
+                    aria-checked={getCheckStatus(isChecked, indeterminate)}
+                    checked={isChecked}
```
greg-a-smith: Do the prop descriptions for `checked` and `defaultChecked` need to be adjusted?

src/Forms/Checkbox.js
```diff
@@ -28,6 +28,8 @@ const Checkbox = React.forwardRef(({ checked, className, defaultChecked, disable
         'fd-checkbox'
     );
 
+    const isChecked = typeof checked === 'undefined' ? defaultChecked : checked;
```
greg-a-smith: Couldn't the same thing happen if both `checked` and `defaultChecked` are undefined?  Wouldn't you just end up with undefined being sent to the `<input>`?

skvale:
Maybe this change isn't needed.

The issue assumes that you'd be using the component like this:
```jsx
    .add('Controlled', () => {
        const [isChecked, setChecked] = useState(false);
        return (
            <>
                <button onClick={() => setChecked(!isChecked)}>Toggle</button>
                <Checkbox checked={isChecked} defaultChecked />
            </>
        );
    })
```
But it really doesn't make sense for the CheckBox to be controlled by the checked prop and uncontrolled with defaultChecked.  

If one is used the other shouldn't be.

bcullman:
So, i think the tests for indeterminate are worth keeping

But I wonder the what is the value of keeping the `uncontrolled component` this is a react component library - users came here to use react components. If you want an uncontrolled checkbox, just use a plain old HTML checkbox.

If that rings true, `defaultChecked` should just be striken from the library (here and on radio buttons)

greg-a-smith:
I agree with @bcullman here.  There is no need to have both of these props.  Let's give `defaultChecked` the 👢.

skvale:
I'm not convinced that removing defaultChecked the right decision.

I see other component libraries having this prop

IBM's Carbon components has the concept of `defaultChecked`
- http://react.carbondesignsystem.com/?path=/story/checkbox--unchecked

Microsoft's ui-frabric has the concept of `defaultChecked`
- https://developer.microsoft.com/en-us/fabric/#/controls/web/checkbox

Microsoft describes it like

> Default checked state. Mutually exclusive to "checked". Use this if you want an uncontrolled component, and want the Checkbox instance to maintain its own state.

---

Imagine the state controlled by a wrapping `<form />`


```jsx
<form onSubmit={(formValues) => /* do something with the checkbox/radio values */}>
	<Checkbox defaultChecked />
    <FormRadioGroup>
       <RadioGroupItem defaultChecked />
       <RadioGroupItem defaultChecked />
    </FormRadioGroup>
</form>
```

These controls are "uncontrolled", but the values are still used by the form's onSubmit.  

If you were to use a `plain old HTML checkbox`, you'd lose the styling, aria-checked, inline support, and having `checked` as the second argument of the onChange callback.

or this

```jsx
<CheckBox defaultChecked onChange={(e, isChecked) => /* API call */} />
```

Would have to maintain its own state

```jsx
const [toggle, setToggle] = useState(true)
<CheckBox 
  checked={toggle} 
  onChange={(e, isChecked) => { setToggle(isChecked); /* API call */ }} 
/>
```

greg-a-smith:
@skvale 🤔 Given that, don't we have to manage some internal React state within the component (when it's uncontrolled) so we don't lose the `aria-selected` (and potentially `checked` depending on what might cause a re-render)?

greg-a-smith:
@skvale I also just looked at your radio button example.  It doesn't even seem like `FormRadioItem` is the right component to have `defaultChecked` as you can't have multiple checked items within a radio group.

skvale:
Using office-ui-fabric as an example, it looks like that's [how they do it](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx#L61)

FormRadioItem and FormRadioGroup is a separate discussion, but yeah, you'd imagine an api like
```jsx
    <FormRadioGroup defaultCheckedId='item-one'>
       <FormRadioItem id='item-one' />
       <FormRadioItem id='item-two' />
    </FormRadioGroup>
```

skvale:
I'm closing this PR.

I think the initial PR was addressing an issue that happens when using the component with `checked` as well as `defaultChecked`.  These should not be used together.

How CheckBox and RadioGroups handle the checked state and keep that in sync with the `aria` attributes will need more thought though. It's are on the radar now.

PR #851, opened 2020-01-17
chore: Add tests to Checkbox for indeterminate and defaultChecked

skvale:
# Description

We wanted to keep the unit tests that came out of this PR: https://github.com/SAP/fundamental-react/pull/849

## Note

Maybe we don't want to keep the TODO comment in the test, but I wanted to leave it in so reviewers could see the issue with using the component in an uncontrolled way.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,3 @@
 build
 lib
 !/.storybook
-lib
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,4 @@
+build
+lib
 !/.storybook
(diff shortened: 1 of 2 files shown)


.eslintignore
```diff
@@ -1,2 +1,4 @@
+build
+lib
 !/.storybook
 lib
```
jbadan: did you mean to duplicate lib?

skvale: nope, somebody beat must have me to it.  I'll take it out

Issue #850, opened 2020-01-16
Datepicker triggers form Submission when clicked on month/year button


coding-my-passion:
### Description
If you use a DatePicker inside the form and select year/month button, form get submitted. It seems like month and year buttons, do not have type attribute. If we set type attribute to button form should not get submitted. 

### Versions
**fundamental-react:**
**fundamental-styles:**

---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

PR #852, opened 2020-01-21
fix: Datepicker triggers form submission when clicked on month, year & calendar icon buttons

coding-my-passion:
### Description

fixes https://github.com/SAP/fundamental-react/issues/850

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,3 @@
+build
+lib
 !/.storybook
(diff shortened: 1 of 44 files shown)


bcullman:
seems like the better fix would be to set the default `typeAttr ` of `Button` to `button`

coding-my-passion:
I have pushed updated snapshots and also email address is set correctly for this commit.

PR #854, opened 2020-01-21
chore: update babel and storybook dependencies

### LINKED ISSUE #853: Update babel to 7.8.0
### Description

Update to babel 7.8.0 - https://babeljs.io/blog/2020/01/11/7.8.0

Why? 
https://babeljs.io/blog/2020/01/11/7.8.0#ecmascript-2020-default-support-10811-https-githubcom-babel-babel-pull-10811-10817-https-githubcom-babel-babel-pull-10817-10819-https-githubcom-babel-babel-pull-10819-10843-https-githubcom-babel-babel-pull-10843

Update build pipeline to use new `--copy-ignored` flag (https://github.com/SAP/fundamental-react/blob/master/package.json#L16)

### Versions
**fundamental-react:**
**fundamental-styles:**

---

_**DISCLAIMER:**

jbadan:
### Description
dependabot is not capable of bumping these in sync with each other, so this must be done manually for now. Also updated the necessary packages after running `npm audit`.

~Note: Only relevant breaking change from standard-version:
https://github.com/conventional-changelog/standard-version/blob/master/CHANGELOG.md#breaking-changes~

~“we now bump the minor rather than major if version < 1.0.0; --release-as can be used to bump to 1.0.0.”~

~@greg-a-smith I did some research and it seems this _shouldn't_ interfere with github assistant, but it might. Any thoughts on how to test?~ removed standard-version from this pr. Will do separately in a followup.

fixes #853


package.json
```diff
@@ -143,13 +143,13 @@
     "sass-loader": "7.3.1",
     "size-limit": "^1.3.6",
     "sort-by": "^1.2.0",
-    "standard-version": "^4.4.0",
+    "standard-version": "^7.1.0",
```
jbadan: @greg-a-smith Do you remember if it was on the standard-version end that the api changed for making our release notes? Is this upgrade ok?

greg-a-smith:
I see the `standard-version` bump was removed from this PR so this comment is no longer relevant to the PR itself.  I will put it here simply because the PR description makes some references to it.

I think getting to a version of `standard-version` that understands how to increment version numbers for 0th version libraries is a good thing.  That said, if the only way to then get to the official 1.0.0 release is to pass the `--release-as` option, we should have a script defined in the `package.json` to do that.  Definitely something that can be ironed out when we do bump to higher version of `standard-version`.

jbadan:
@greg-a-smith created an issue! https://github.com/SAP/fundamental-react/issues/874

Issue #829, opened 2019-12-29
DatePicker's Calendar Popup should close when pressing Esc Key


codefactor:
### Description

If the user presses the "Esc" key the Calendar Popup should close.

### Versions
**fundamental-react:**
**fundamental-styles:**

---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

Issue #827, opened 2019-12-29
DatePicker uses componentWillMount which has been renamed, and not recommended for use.


codefactor:
### Description

The following warning is displayed in the console:

```
react-dom.development.js:12357 Warning: componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.

* Move code with side effects to componentDidMount, and set initial state in the constructor.
* Rename componentWillMount to UNSAFE_componentWillMount to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

Please update the following components: DatePicker
```

### Versions
**fundamental-react:** 1.0.0.rc.0, all versions
**fundamental-styles:**

---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

jacobdevera:
fixed by #858

Issue #787, opened 2019-11-01
Update Calendar markup for a11y


jacobdevera:
### Description
The markup should be updated to reflect changes made in `fundamental-styles`.
https://github.com/SAP/fundamental-styles/pull/410
- Convert month/year selector `ul` into `table` with role=“grid”
- Currently selected date: `<td>`: aria-selected=“true”  and `<span>` role "button"

Additional changes:
- Month/year header: aria-live=“polite” (to announce month/year changes)
- Navigation buttons: Needs localizable `aria-label` (e.g. “Next Year”, "Next Month")
- Selectable dates: `aria-label` should be used to read out full date
- Disabled dates: `<td>`: `aria-disabled="true"`

jacobdevera:
fixed by #858

PR #858, opened 2020-01-27
feat: keyboard navigation in calendar

jacobdevera:
### Description
Adds the following characteristics to Calendar:
* Focus is set on the current date by default
* Press the LEFT and RIGHT arrow keys to navigate the row by week day.
* Press the UP and DOWN arrow keys to navigate between weeks on the same week day.
* Will go to next/previous month if:
    * Passing first/last day of the month with left/right arrow keys OR
    * Passing first/last week of the month with up/down arrow keys
* Press the HOME and END keys to jump to the beginning or end of the current row
* Press the PAGEDOWN and PAGEUP keys to navigate backwards or forwards by month
* Press SHIFT+PAGEDOWN and SHIFT+PAGEUP to navigate backwards or forwards by year
* Press the ENTER key to activate the selected date
* Mouse users can click the desired date buttons as usual
* Tab cycles between date grid and month/year navigation keys
* Should be able to activate month/year header to enter month/year selection mode with similar keyboard interactions to selecting dates

calendar-kb 2020-01-26 22_45_07

### Known issues

- When a cell is both focused and selected, the white text against light grey background is a contrast accessibility issue. It looks awkward for range selection as well with the border-radius difference. Will require a styles update to address.

### Desired

- [x] Use Popover for DatePicker to allow for Escape to close the popover, and trap focus within the Calendar: some issues with date validation currently that is making this difficult

- [ ] Announce instructions for arrow key navigation: accessibility-wise, it's best for this to be onscreen and recognizable by both sighted and blind users, though if we want to hide this, would probably require a styles update.

### References
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html

### Fixes
fixes https://github.com/SAP/fundamental-react/issues/829 & https://github.com/SAP/fundamental-react/issues/827 & #787


src/Calendar/Calendar.js
```diff
@@ -118,8 +188,7 @@ class Calendar extends Component {
     changeMonth = (month) => {
         const newDate = moment(this.state.currentDateDisplayed)
             .locale(this.props.locale)
-            .month(month)
-            .date(1);
+            .month(month);
```
bcullman: careful: This was *probably* put in place so that you don't accidentally jump from 1/31/2020 to 2/31/2020

jacobdevera: With moment, it will now set it to the latest day if the new month doesn't have enough days. Can go either way on this though.

meganaconley:
When I use PAGEUP/PAGEDOWN it selects the day it lands on in the next/previous month, even if I had already selected a day.

Screen Recording 2020-01-27 at 1 48 41 PM

Another note is the odd interaction at first. I did a shift-tab back into the calendar and it put my focus on Feb 1 at first, but hitting left arrow got me on Jan 30 instead of Jan 31. I haven't been able to repro though.

jbadan:
Screen Shot 2020-01-31 at 11 06 46 AM


Should the Calendar popover close once I move focus to a new DatePicker input?

greg-a-smith:
Not specifically related to keyboard navigation, but when going from April 2020 to May 2020, the size of the calendar grows because of how the days line up.  There are actually 6 weeks with days in May.  Should we add something such that 6 weeks always show in the Calendar (and then DatePicker) so there is no "jumping"?

greg-a-smith:
Overall, it seems pretty good.  Some random thoughts/observations:

* The currently selected/focused element (in the day, month or year view) is gray (sometimes with white text).  It's difficult to tell that it's where your cursor will start from -- it looks disabled.
* When the year view is invoked, the PAGEUP and PAGEDOWN change months.  I understand this is the behavior in the day or month view (which makes sense), but it _seems_ odd when you're looking at years.
* When the focused day/month/year is on the currently selected value, it loses the blue background and it's just gray.
* There's no way to dismiss the month or year view other than selecting a month or year.  It _seems_ like ESC should switch back to the day view.

jacobdevera:
> Should the Calendar popover close once I move focus to a new DatePicker input?
For sure. May probably take another stab at changing Calendar to a Popover.

> Should we add something such that 6 weeks always show in the Calendar (and then DatePicker) so there is no "jumping"?

I don't see a specification in Fiori for always showing 6 weeks, and I would think it would be odd to include an extra week entirely from the next month if the month only has 5 weeks.

>The currently selected/focused element (in the day, month or year view) is gray (sometimes with white text). It's difficult to tell that it's where your cursor will start from -- it looks disabled.

>When the focused day/month/year is on the currently selected value, it loses the blue background and it's just gray.

These are styles issues that were not initially made in the first pass through changing calendar styles, because there wasn't a spec for focus indication in Fiori. I just opted to use the hover style, which isn't ideal in hindsight. This PR (https://github.com/SAP/fundamental-styles/pull/506) seems to have a good solution for focus styles. Wonder if we should wait for this and upgrade fundamental-styles before merging this in.

> When the year view is invoked, the PAGEUP and PAGEDOWN change months. I understand this is the behavior in the day or month view (which makes sense), but it seems odd when you're looking at years.

If PGUP/PGDN changed the year in the year view, then we probably have to change what SHIFT+PGUP/PGDN does too. I'm not sure making those keys do the same thing, do nothing, or making it change the month is a good idea.

>There's no way to dismiss the month or year view other than selecting a month or year. It seems like ESC should switch back to the day view. Should month and year views be popovers/overlays instead?

I think changing the month and year views to popover would be quite a big refactor (needing to remove the box shadow, trying to make it work within the Calendar popover, etc.). Ultimately other date pickers I have seen just close the whole widget altogether if there is a separate month/year picker, which doesn't seem to be a bad idea.

bcullman:
@jacobdevera - we do 6 week cal in the exist datepicker. its important to have a static height for the calendar, otherwise, while flipping through the months in a near page edge situations, popper will flip the calendar up when it encounters 6 weeks, and flip back when it goes back to 5 weeks. 

a static height assures that popper wont reposition the calendar while navigating through months.

src/DatePicker/DatePicker.test.js
```diff
@@ -25,70 +25,6 @@ describe('<DatePicker />', () => {
         mount(compactRangeDatepicker);
     });
 
-    test('open/close by calendar icon button', () => {
```
jacobdevera: These essentially test opening/closing the calendar which should now be covered by using the Popover component itself.

jbadan: 🔥

jbadan:
Once you merge latest master into here, can you also remove these lines: https://github.com/SAP/fundamental-react/blob/master/src/DatePicker/DatePicker.js#L25

Because you replaced the hardcoded html for popover in here with the actual Popover component, it doesn't need this style block again

PR #873, opened 2020-02-15
feat: replace withStyles wrapper with useEffect and remove customStyles property

jbadan:
### Description

By removing dynamic requires, consumers will not be bringing in all of fundamental-styles with each component. 

"...This means dynamic requires are supported but will cause all matching modules to be included in the bundle." https://webpack.js.org/guides/dependency-management/#require-with-expression

I decided to remove `withStyles` instead of repurposing it. Two reasons:

1. passing `cssFile: require('fundamental-styles/dist/button.css)` to `withStyles` will cause the css to always be loaded and not respect `disableStyles` prop.
1. passing `cssFile: 'fundamental-styles/dist/button.css'` and then calling it like `require(cssFile)` in `withStyles` does not work because of the following: 

"Fully dynamic statements, such as import(foo), will fail because webpack requires at least some file location information.The import() must contain at least some information about where the module is located, so bundling can be limited to a specific directory or set of files.
Just in case someone wonders, the same rule applies for dynamic require statements"
https://stackoverflow.com/questions/44166393/uncaught-error-cannot-find-module-when-using-dynamic-import-for-javascript

 On the remaining class components (which we need to convert at some point), I used `componentDidMount` and on functional components `useEffect`. 

I also removed `customStyles` prop. Because we are not using a hoc anymore, we can just import our custom styles in our wrapper inside nui-widgets.


src/DatePicker/DatePicker.js
```diff
@@ -23,6 +22,12 @@ class DatePicker extends Component {
         this.calendarRef = React.createRef();
     }
 
+    componentDidMount() {
+        if (!this.props.disableStyles) {
+            require('fundamental-styles/dist/popover.css');
```
jbadan: This whole section can be removed if https://github.com/SAP/fundamental-react/pull/858 is merged first. That rebuilds this with a popover so we get those styles from popover component.

src/Shellbar/Shellbar.js
```diff
@@ -498,6 +503,4 @@ Shellbar.propDescriptions = {
     subtitle: 'Displays an application context. Should be used rarely.'
 };
 
-export { Shellbar as __Shellbar };
-
-export default withStyles(Shellbar, { cssFile: ['shellbar', 'product-switch'], fonts: true });
```
skvale: should this file also `require('fundamental-styles/dist/fonts.css');` since fonts was true?

src/Image/__stories__/Image.stories.js
```diff
@@ -15,10 +15,4 @@ storiesOf('Components|Image', module)
         <Image disableStyles
             photo='https://content.fortune.com/wp-content/uploads/2019/07/hippocorn.jpg'
             size='l' />
-    ))
-    .add('custom styles', () => (
-        <Image
-            customStyles={require('../../utils/WithStyles/customStylesTest.css')}
-            photo='https://content.fortune.com/wp-content/uploads/2019/07/hippocorn.jpg'
```
skvale: just so we know, this is what we're losing :(
hippocorn

src/Dropdown/Dropdown.js
```diff
@@ -1,10 +1,15 @@
...
 import PropTypes from 'prop-types';
-import React from 'react';
-import withStyles from '../utils/WithStyles/WithStyles';
+import React, { useEffect } from 'react';
 
 const Dropdown = React.forwardRef(({ children, className, disableStyles, standard, ...props }, ref) => {
 
+    useEffect(() => {
+        if (!disableStyles) {
+            require('fundamental-styles/dist/dropdown.css');
```
skvale: should I see `dropdown.css` in fundamental-styles?  https://github.com/SAP/fundamental-styles/tree/master/src ?  Because I don't

skvale: looks like it exists in 0.4.0 though, nevermind

src/Shellbar/Shellbar.js
jbadan: Nope - there's a bunch of other components imported into this component that already require fonts. That was an oversight to include them in the first place with `withStyles`

PR #875, opened 2020-02-18
POC: global fonts + icons, theming 

jbadan:
### Description

This is the idea I had in mind to handle global styles. Note that this doesn't work perfectly by any means, just an example of what we could do.

fixes #issueid

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.8.0",
+    "@sap-theming/theming-base-content": "^11.1.16",
     "chain-function": "^1.0.1",
     "classnames": "^2.2.6",
     "fundamental-styles": "~0.4.0-rc.3",
--- a/src/ThemeProvider/ThemeProvider.js
+++ b/src/ThemeProvider/ThemeProvider.js
@@ -1,0 +1,46 @@
+/* eslint-disable default-case */
+import PropTypes from 'prop-types';
+import React, { useEffect } from 'react';
+
+const ThemeContext = React.createContext(null);
+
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -47,6 +47,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.8.0",
+    "@sap-theming/theming-base-content": "^11.1.16",
```
jbadan: This would ideally come from fundamental-styles so we don't need an extra dependency.

src/ThemeProvider/ThemeProvider.js
```diff
@@ -0,0 +1,46 @@
...
+            console.log(theme)
+            switch (theme) {
+                case 'dark':
+                    require('@sap-theming/theming-base-content/content/Base/baseLib/sap_fiori_3_dark/css_variables.css');
+                    break;
+                case 'light-dark':
+                    require('@sap-theming/theming-base-content/content/Base/baseLib/sap_fiori_3_light_dark/css_variables.css');
+                    break;
+                case 'default':
+                    require('@sap-theming/theming-base-content/content/Base/baseLib/sap_fiori_3/css_variables.css');
```
greg-a-smith: I get where you're going with this and I think there's some merit here.  I am curious if there's a good way to essentially "unrequire" when toggling themes (or anything for that matter).  The `require` statement loads more styles, but it never really unloads any.

jbadan: Yeah, the more research I do, I don't think this is going to be the solution for theming. What do we think about it for fonts and icons though?

PR #878, opened 2020-02-21
feat: upgrade fundamental-styles to 0.4.0-rc.24

jbadan:
### Description

I was originally going to upgrade only to `0.4.0-rc.10`, but there were no additional breaking changes between `0.4.0-rc.10` and `rc-24`.

Changes **bold denotes breaking change**:
  
FormInput
- changed `state` prop options to match new validation states

Checkbox
- new `state` prop: added validation states

FormRadioGroup
- **`div` to Fragment to match html from Fundamental Styles**

FormItem
- new `isHorizontal` prop
- **`isInline` becomes internal only prop as it is only meant for Checkboxes and Radios**

FormRadioItem
- new `compact` prop
- new `state` prop: added validation states
- **`inline` prop becomes internal use only - should only be applied to `FormRadioGroup`**

FormSelect
- new `state` prop: added validation states
- new `compact` prop

FormTextarea
- new `compact`, `disabled` and `readOnly` props
- new `state` prop: added validation states

Combobox Input
- rebuilt with InputGroup
- new `state` prop: added validation states
 
SearchInput
- rebuilt with InputGroup
- new `state` prop: added validation states

InputGroup
- new `disabled` prop
- new `state` prop: added validation states
-  **fixed `compact` prop so that appropriate `fd-input-group__addon—compact` class is added to InputGroup.Addon**

MultiInput
- rebuilt with InputGroup
- **changed `placeHolder` prop to `placeholder`**
- new `state` prop: added validation states

FormLabel
- reworked `required` prop
- new `is-Toggle` prop - internal use only
- new `is-Checkbox` prop - internal use only
- new `is-Radio` prop - internal use only
- new `is-InlineHelp` prop - add this to any labels used with InlineHelp component

FormMessage
- **changed types options: help -> information**

TimePicker
- rebuilt with InputGroup

Alert
- **`div` for text changed to `p`**

Known Issues out of the scope of this pr:
* Dropdown missing top button border
* InlineHelp doesn’t actually work when paired with FormLabel + FormInput
* Spacing inside TimePicker popover with TimeItems
* checkboxes and radios styling looks bad, but this is fixed in a later rc


package.json
```diff
@@ -49,8 +49,7 @@
     "@babel/runtime": "^7.8.0",
     "chain-function": "^1.0.1",
     "classnames": "^2.2.6",
-    "fundamental-styles": "~0.4.0-rc.3",
-    "hoist-non-react-statics": "^3.3.0",
```
jbadan: hoist-non-react-statics should have been removed with `withStyles`

src/ComboboxInput/__stories__/ComboboxInput.stories.js
```diff
@@ -3,41 +3,43 @@ import Menu from '../../Menu/Menu';
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 import {
+    boolean,
+    select,
     withKnobs
 } from '@storybook/addon-knobs';
 
+const menu = () => (<Menu>
```
skvale: This story doesn't work if is menu is a function, looks like you have to call it below `menu()` or just define it as `const menu = (<Menu>...`

src/ComboboxInput/ComboboxInput.js
```diff
@@ -61,7 +50,8 @@ ComboboxInput.propTypes = {
     disableStyles: PropTypes.bool,
     inputProps: PropTypes.object,
     placeholder: PropTypes.string,
-    popoverProps: PropTypes.object
+    popoverProps: PropTypes.object,
+    state: PropTypes.oneOf(FORM_STATES)
 };
 
 ComboboxInput.propDescriptions = {
```
skvale: it's a separate issue, but should all props have a corresponding propDescription?  Should you be adding one for `state`?

src/Forms/Forms.Component.js
```diff
@@ -55,6 +55,48 @@ export const FormsComponent = () => {
...
+                            </FormLabel>
+                            <FormInput id='input-2' placeholder='Field placeholder text'
+                                type='text' />
+                        </FormItem>
+                    </FormGroup>
+
+                    <FormGroup>
+                        <FormItem isHorizontal>
+                            <FormLabel htmlFor='input-3' required>
+                                Password
```
skvale: > When a field is required, the label should include an asterisk (*).

Does that mean the user is responsible for adding an `*` and this should be `Password *` ?

src/Forms/__stories__/FormRadioGroup.stories.js
```diff
@@ -3,20 +3,40 @@ import FormRadioItem from '../FormRadioItem';
...
+});
+
+const createRadioProps = (overrides) => ({
+    disabled: boolean('disabled', false),
+    state: select('Validation State', {
+        'default': null,
+        'valid': 'valid',
+        'invalid': 'invalid',
+        'information': 'information',
+        'warning': 'warning'
```
skvale: How about we just make validation state in a story util file so we can import it in all these different places.  Makes it easier to update all the spots when a new validationState comes along.
```js
// some-storybook-util-file.js
export const validationStateKnob = select('Validation State', {
  'default': null,
  'valid': 'valid',
  'invalid': 'invalid',
  'information': 'information',
  'warning': 'warning'
});
```

src/ComboboxInput/__stories__/ComboboxInput.stories.js
jbadan: oops! Good catch!

src/Forms/Forms.Component.js
jbadan: Nah, our docs need an entire story just to rewrite. I'll fix this now though.

src/ComboboxInput/ComboboxInput.js
jbadan: There's a global prop definition for state since it's shared amongst so many components.

jbadan:
Closing this for now - I'm going to upgrade to 21 since a lot of the validation states were removed between 10 and 20, but added back in 21.

src/Forms/__stories__/FormRadioGroup.stories.js
jbadan: That's a good idea! I would prefer to leave it as a followup for when we convert these stories to the new syntax though - I didn't want to do that here and create such a huge diff.

src/Forms/Checkbox.js
```diff
@@ -30,7 +48,8 @@ const Checkbox = React.forwardRef(({ checked, className, defaultChecked, disable
 
     const classes = classnames(
         className,
-        'fd-checkbox'
+        'fd-checkbox',
+        { [`is-${state}`]: state }
```
skvale: I want to entertain this thought:  Should we be defining classes with full strings instead of interpolation.

If we use string interpolation we wouldn't work with something like [PurgeCSS](https://purgecss.com/#table-of-contents) in case a consumer wants to configure their project to minimize the imported styles.

So we'd want some utility to add the full strings here
```js
 { ['is-warning']: state.warning, ['is-invalid']: state.invalid, ... }
```

jbadan: I like the idea of this - maybe as a follow up if we want to use purgeCSS in nui-widgets/floorplans. We could go through all the components with these issues.

src/Forms/FormItem.js
```diff
@@ -33,11 +34,13 @@ FormItem.propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
     disableStyles: PropTypes.bool,
+    isHorizontal: PropTypes.bool,
```
jbadan: I don't love the name of this prop 🤷‍♀

src/Forms/FormLabel.js
```diff
@@ -13,7 +24,14 @@ const FormLabel = React.forwardRef(({ required, children, className, disabled, d
 
     const formLabelClasses = classnames(
         'fd-form-label',
-        { 'is-disabled': disabled },
+        {
+            'is-disabled': disabled,
+            'fd-form-label--toggle': isToggle,
+            'fd-form-label--checkbox': isCheckbox,
+            'fd-form-label--radio': isRadio,
```
jbadan: I don't really love this pattern - but checkbox and radio are going to change significantly in a couple more rc rev's so there's no point in really diving into it right now.

src/Forms/Checkbox.js
bcullman: I also prefer to avoid string interpolation for classnames - they don't show up in search and replaces efforts, which inevitably leads to bugs.

greg-a-smith: I can see @bcullman's point and in our legacy applications I completely agree.  However, I will say that in the world of self-contained styles, I don't think this is nearly as big of an issue anymore.

src/ComboboxInput/__stories__/ComboboxInput.stories.js
```diff
@@ -3,41 +3,43 @@ import Menu from '../../Menu/Menu';
...
-                        <Menu.Item url='#'>Strawberry</Menu.Item>
-                        <Menu.Item url='#'>Raspberry</Menu.Item>
-                        <Menu.Item isLink url='#'>
-                            + New Item
-                        </Menu.Item>
-                    </Menu.List>
-                </Menu>
-            }
-            placeholder='Select Fruit' />
+            {...createProps()}
```
greg-a-smith: Going forward, it would be good to have one "default" story for each component that contains all the storybook knobs, but all additional stories should probably disable knobs as the stories really become specific use cases with all the appropriate props set and shouldn't be "played with" anymore.  There are a few exceptions to this, yes, but it would be good if that was the general pattern.  That said, not part of this PR.

src/Forms/FormItem.js
greg-a-smith: It looks like `isHorizontal` puts the label and the input side-by-side on the same visual row.  That seems more like `isInline` to me, but I also see that prop existed before.  I'm not quite sure what the difference is between `isHorizontal` and `isInline`.  Looking at the `fundamental-styles` page on forms, there are no uses of `fd-form-item--inline`. 🤷‍♂

src/Forms/FormRadioItem.js
```diff
@@ -1,16 +1,41 @@
...
         if (!disableStyles) {
             require('fundamental-styles/dist/radio.css');
         }
     }, []);
 
+    const inputClassNames = classnames(
+        'fd-radio',
+        {
+            'fd-radio--compact': compact,
+            [`is-${state}`]: state
```
greg-a-smith: Not exactly "self-contained", but that's a _WHOLE OTHER_ issue.

src/Forms/Forms.Component.js
```diff
@@ -12,8 +12,8 @@ export const FormsComponent = () => {
             title='Forms'>
 
             <Example
-                description={`Inputs are used to collect data from the user. When a field is required, the label is displayed in bold
-                    and noted by an asterisk (*).`}
+                description={`Inputs are used to collect data from the user. When a field is required,
+                the <code>required</code> prop will include an asterisk (*).`}
```
greg-a-smith: I think the playground components are setup to support markdown if you didn't want to hardcode markup tags.

src/Forms/FormItem.js
jbadan: `--inline` is only for inline labels for radio and checkbox. I think we can just go forward like this, then with the rc that changes how checkboxes and radios work can revisit that?

src/Forms/Checkbox.js
jbadan: Let's 🅿️ this today... the more I think about it, the less I want to make this change

src/ComboboxInput/__stories__/ComboboxInput.stories.js
jbadan: Agree! I started to do that... then realized it's soooo out of scope. What do you think about a story in the backlog to 1) upgrade storybook to use the "new" pattern we've established in floorplans 2) setup the stories consistently for the components?

greg-a-smith: Yeah, that sounds like a good plan.

PR #884, opened 2020-02-25
fix: a few issues caused by directly using fundamental-styles classes instead of the respective components

BodhiHu:
### Description

Fix a few issues caused by directly using fundamental-styles classes instead of the respective components.

### Fixed issues:

* fix ```DatePicker``` popover hidden by following components (caused by not rendering the popover with the ```Popover``` compoment);
* DatePicker, SearchInput: set default width to "100%" to match parent element width;
* SearchInput: render Menu.Item instead of ```<li>``` tag to prevent the react-dom props check warnings (passing unknown props to ```<li>``` tag);

### DIFF
--- a/src/DatePicker/DatePicker.js
+++ b/src/DatePicker/DatePicker.js
@@ -133,8 +133,6 @@ class DatePicker extends Component {
             blockedDates,
             buttonLabel,
             buttonProps,
-            popoverStyle,
-            popoverProps,
             className,
             compact,
             disableAfterDate,
@@ -150,6 +148,7 @@ class DatePicker extends Component {
             locale,
             localizedText,
             onBlur,
+            popoverStyle,
             ...props
         } = this.props;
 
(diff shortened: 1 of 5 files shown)


src/DatePicker/DatePicker.js
```diff
@@ -220,14 +222,18 @@ DatePicker.propTypes = {
...
 };
 
 DatePicker.defaultProps = {
     buttonLabel: 'Choose date',
     defaultValue: '',
     locale: 'en',
-    onBlur: () => {}
+    onBlur: () => {},
+    popoverStyle: {
+        width: '100%'
```
jacobdevera: 1. Can you show the issue (screenshot) that was present before putting this style in? 
2. Does it make sense to set a default style in here rather than making a change in `fundamental-styles`? And once a consumer sets this prop, they would need to either include the `width` change or it would be overwritten.

jbadan: Is this trying to solve the Popover width sizing issues? We can set a default of one of these options instead: https://fundamental-react.netlify.com/popover#_width-_sizing-_type

jbadan:
Thanks for your contribution! Make sure to run `npm run test -- -u` to update the snapshots so that Travis will pass.

BodhiHu:
Has fixed the snapshots, please help review again. - Thanks

src/DatePicker/DatePicker.js
BodhiHu: input width not matching that of parent

Hi as you can see above, each input field is given a width of 240px, the ```SearchInput``` and ```DatePicker```'s width would not match that of their parents.

jbadan: Did `popperProps={{ widthSizingType: "matchTarget" }}` not work for this use case?

BodhiHu: input width not matching that of parent even with widthSizingType as 'matchTarget'

As you can see from above, passing ```widthSizingType={"matchTarget"}``` actually did not fix this issue.

jacobdevera: @jbadan There's a mixup between the "control" element (InputGroup) and the popper content itself. The issue here is the DatePicker InputGroup isn't aligning widths with the rest of the form fields in the screenshot. 

The `popoverStyle` prop here only affects the styling on the main popover div (with the control element), whereas `widthSizingType` is for the popper content itself and wouldn't affect it.

image

The naming of the prop is a little misleading.

jbadan: Ah ok - sorry for the misunderstanding. This looks like a change needs to happen in fundamental-styles https://github.com/SAP/fundamental-styles/blob/master/src/popover.scss#L49

jbadan:
Because this change should take place in the fundamental-style repository, I'm going to close this pr. Feel free to reopen if you disagree.

PR #888, opened 2020-02-28
feat: update fundamental-styles to 0.6.0-rc.26

jbadan:
### Description

This is a continuation of https://github.com/SAP/fundamental-react/pull/885

I'm opening these so that reviewers can see the changes in "smaller" increments. There will a final pr to review with the final changes. Feel free to leave comments here though.

- Tab
    - remove `disabled` prop
- Shellbar
    - logo `<a>` is now `<span>`
    - added `<div>` around Identifier component
- Alert
    - renamed MessageStrip
    - classname changes 
    - removed Icon from underlying html
- Dropdown
    - renamed `Select`
    - remove `standard` prop
- ListGroup
    - renamed “List”
    - classname changes
    - so many changes! basically a new component!
- MultiInput
    - underlying html changes
    - `Popover` body now built with `List` component
    - lots of underlying html changes for validation states
    - `state` prop replaced with `validationState` - takes object
    - `localizedText` prop removed
- Combobox Input
    - underlying html changes
    - `menu` prop renamed `list` - now takes `List` component
    - `state` prop now `validationState` - takes an object 
- FormMessage
    - now a `<div>` from `<span>`
- InputGroup
    - `state` prop now `validationState` - takes an object
    - underlying html changes to include `FormMessage` when passed `validationState`
- SearchInput
    - `state` prop now `validationState` - takes an object
    - underlying html changes to include `FormMessage` when passed `validationState`
- FormLabel
    - removed `isToggle` prop
- Toggle
    - renamed to “Switch”
    - removed size options, now use `compact` property
    - `labelProps` no longer needed, additional props will be spread to the `<label>` element
    - underlying html changes - no longer wrapped  in `FormItem`


src/ComboboxInput/ComboboxInput.js
```diff
@@ -1,61 +1,121 @@
...
+                validationState ? (
+                    <>
+                        <FormMessage
+                            disableStyles={disableStyles}
+                            type={validationState.state}>
+                            {validationState.text}
+                        </FormMessage>
+                        {list}
+                    </>
+                ) : list}
```
skvale: This seems a tiny bit cleaner to me
```jsx
<>
  {validationState && 
    <FormMessage
      disableStyles={disableStyles}
      type={validationState.state}>
        {validationState.text}
    </FormMessage>
  }
  {list}
</>
```

jbadan:
Known Issues:
    * Toggle/Switch
        * internalLabel prop not working in prop descriptions
        * not an accessible pattern
        * this is going to change ASAP as it is currently being redesigned 
        * icons are misaligned inside switch
    * Token: 
        * selected state - did not implement - what is this even used for?
    * Shellbar
        * see screenshot - `fd-shellbar__button` needs more specificity
    * ListText
        * —no-wrap class only works in selects
    * Select
        * compact not working because `—fd-forms-height-compact` is not part of our postcss pipeline anymore - just SAP variables
    * MultiInput
        * `fd-tokenizer__input` does not take priority over `fd-input` - getting the fd-input outline box (needs more specificity)
    * Validation states
        * validation state styling - FS has the validation message as a list item instead of a div

PR #890, opened 2020-02-28
feat: update fundamental-styles to 0.7.0-rc.13

jbadan:
### Description

Modal
   - renamed `Dialog`
   - removed `closeProps`
   - `actions` now required
Button
   - underlying class name changes
ButtonGroup
   - classname changed to `fd-segmented-button`
- Tab
    - remove `disabled` prop
   - change underlying html - added `span` around tab text
- Shellbar
    - logo `<a>` is now `<span>`
    - added `<div>` around Identifier component
- Alert
    - renamed MessageStrip
    - classname changes 
    - removed Icon from underlying html
- Dropdown
    - renamed `Select`
    - remove `standard` prop
- ListGroup
    - renamed “List”
    - classname changes
    - so many changes! basically a new component!
- MultiInput
    - underlying html changes
    - `Popover` body now built with `List` component
    - lots of underlying html changes for validation states
    - `state` prop replaced with `validationState` - takes object
    - `localizedText` prop removed
- Combobox Input
    - underlying html changes
    - `menu` prop renamed `list` - now takes `List` component
    - `state` prop now `validationState` - takes an object 
- FormMessage
    - now a `<div>` from `<span>`
- InputGroup
    - `state` prop now `validationState` - takes an object
    - underlying html changes to include `FormMessage` when passed `validationState`
- SearchInput
    - `state` prop now `validationState` - takes an object
    - underlying html changes to include `FormMessage` when passed `validationState`https://github.com/SAP/fundamental-react/pull/885
- Toggle
    - renamed to “Switch”
    - removed size options, now use `compact` property
    - `labelProps` no longer needed, additional props will be spread to the `<label>` element
    - underlying html changes - no longer wrapped  in `FormItem`
* Checkbox
    * new `children` prop
    * `value` prop changed
    * underlying html has changed significantly 
* FormRadioItem
    * underlying html has changed significantly 
* FormMessage
    * now an internal component
* FormInput
    * removed `state` property
    * new `validationState` property
* FormItem
    * remove `isHorizontal` prop
    * remove `isInline` prop (was internal only)
* FormRadioGroup
    * now outputs `<div>` instead of Fragment
* FormLabel
    * remove `isCheckbox` prop
    * remove `isRadio` prop
    - removed `isToggle` prop

See https://github.com/SAP/fundamental-react/pull/888, https://github.com/SAP/fundamental-react/pull/885 for a divided pr

Known Issues:
Toggle/Switch
        - internalLabel prop not working in prop descriptions
        - not an accessible pattern
        - this is going to change ASAP as it is currently being redesigned 
        - icons are misaligned inside switch
Token: 
        - selected state - did not implement - what is this even used for?
Shellbar
        - `fd-shellbar__button` needs more specificity
  ListText
        - `—no-wrap` class only works in selects - this should probably be part of select, not list
Select
        - compact not working because `—fd-forms-height-compact` is not part of our postcss pipeline anymore - just SAP variables: https://github.com/SAP/fundamental-styles/pull/711
MultiInput
        - `fd-tokenizer__input` does not take priority over `fd-input` - getting the fd-input outline box (needs more specificity)
Validation states
        - validation state styling - FS has the validation message as a list item instead of a div:https://github.com/SAP/fundamental-styles/issues/712
        - “onClickOutside” is not being handled correctly


.size-limit
```diff
@@ -3,6 +3,6 @@
     name: "Fundamental-React Size",
     webpack: true,
     path: "lib/index.js",
-    limit: "180 KB"
+    limit: "200 KB"
```
jbadan: it's only 186 but I'm giving myself some headroom

src/Dialog/Dialog.Component.js
```diff
@@ -0,0 +1,63 @@
+import path from 'path';
+import { Button, Dialog } from '..';
+import { ComponentPage, Example } from '../_playground';
+import React, { Component } from 'react';
+
+export class DialogComponent extends Component {
```
skvale: Could this be a functional component?

src/Dialog/Dialog.js
```diff
@@ -1,12 +1,11 @@
-import Button from '../Button/Button';
 import classnames from 'classnames';
 import CustomPropTypes from '../utils/CustomPropTypes/CustomPropTypes';
 import FocusLock from 'react-focus-lock';
 import PropTypes from 'prop-types';
 import ReactDOM from 'react-dom';
 import React, { Component } from 'react';
-class Modal extends Component {
-    // select body element to add Modal component too
+class Dialog extends Component {
```
skvale: Doesn't need to happen here, but this could be a functional component pretty easy too.

src/Dialog/Dialog.Component.js
jbadan: oh this is copied from the previous ModalComponent - it just looks new from the name change. I don't really want to rework all the docs 😆

src/List/ListIcon.test.js
```diff
@@ -0,0 +1,26 @@
...
+    );
+
+    test('create list Icon item', () => {
+        let component = renderer.create(ListIcon);
+        let tree = component.toJSON();
+        expect(tree).toMatchSnapshot();
+    });
+
+    describe('Prop spreading', () => {
+        test('should allow props to be spread to the ListIcon component', () => {
```
skvale: tiny thing, this could say `List.Icon component` so it's not confused with the ListIcon thing right above

src/MessageStrip/__stories__/MessageStrip.stories.js
```diff
@@ -0,0 +1,54 @@
...
+    ...overrides
+});
+
+storiesOf('Components|MessageStrip', module)
+    .addDecorator(withKnobs)
+    .add('Dev', () => (
+        <MessageStrip {...createProps()}>Default MessageStrip</MessageStrip>
+    ))
+    .add('Dismissable', () => (
+        <MessageStrip dismissable>MessageStrip</MessageStrip>
```
skvale: I think it's spelled `dismissible`.  But this does seem like the right spelling

src/MessageStrip/MessageStrip.test.js
```diff
@@ -0,0 +1,95 @@
...
+        const component = renderer.create(nonDismissibleMessageStrip);
+        const tree = component.toJSON();
+
+        expect(tree).toMatchSnapshot();
+
+        // const wrapper = shallow(basicMessageStrip);
+        // expect(wrapper.state(['isActive'])).toBeTruthy();
+
+        // wrapper.find('button.fd-message-strip__close').simulate('click');
+        // expect(wrapper.state(['isActive'])).toBeFalsy();
```
skvale: ?? Can we take these out now?

src/MessageStrip/__stories__/MessageStrip.stories.js
jbadan: oh good catch - I can't spell 🤣

src/MessageStrip/MessageStrip.test.js
jbadan: 🔥

bcullman:
modal looks like it is missing some padding:
image

TabGroup looks... not right.
image

validation state on multiinput cause field to shift around:
valstates

^ same issues in selects

popover placement in docs seems smushed:
image

src/MessageStrip/__stories__/MessageStrip.stories.js
jbadan: `dismissible` is the correct spelling - I looked through the history and I've had to fix this a couple times because we(me) can't get it right. Maybe we change the prop name in the future 😆

jbadan:
@bcullman 

1. Dialog by default has no padding in the body 🤷‍♀ I added the additional sizes so that consumers can add padding as they wish. See https://fundamental-styles.netlify.com/components/dialog.html#dialog-size-modifiers
I set the defaulted to large to prevent any confusion.

1. TabGroup is updated with the new span class and I've now added the sizes now (defaulted to large like Dialog since the default from fundamental-styles has no padding). Other features see: https://github.com/SAP/fundamental-react/issues/891

1.  Popover & validation state jumping fixed on docs site - was a flexbox issue.

PR #908, opened 2020-03-03
feat: remove fonts from package, update ActionBar, Button and InfoLabel

jbadan:
### Description

* Fonts are no longer included in each component, follow instruction in readme to include fonts in your own project. 
* New peer dependency on @sap-theming/theming-base-content
* Counter
    * moved from `Badge` folder to `Counter` folder - import as `import { Counter } from 'fundamental-react/Counter'`
* Badge, Label
    * removed, use new InfoLabel component
* ActionBar
    * Removed ActionBar.Back, ActionBar.Header, ActionBar.Actions component, now built into ActionBar via props
* Button
    * `option=‘light’` to `option=‘transparent’`


.storybook/preview-head.html
```diff
@@ -0,0 +1,29 @@
+<style type="text/css">
+@font-face {
+    font-family: '72';
+    src: url('./72-Regular.woff') format('woff');
```
jbadan: This is stupid and annoying, but I couldn't get the fonts from @sap-theming to load from node_modules in here 😢

scripts/build.js
```diff
@@ -47,10 +47,6 @@ if (!checkRequiredFiles([paths.appHtml, paths.appIndexJs])) {
     process.exit(1);
 }
 
-fs.copyFile('README.md', 'src/_playground/documentation/Home/README.md', (err) => {
-    if (err) throw err;
-});
-
```
jbadan: moved to `scripts/copy-assets.js`

src/ActionBar/ActionBar.js
```diff
@@ -1,42 +1,121 @@
...
+};
+
+ActionBar.propDescriptions = {
+    actionClassNames: 'Classnames to spread to the action Button container.',
+    actionProps: 'Props to spread to the action Button container',
+    actions: 'Button components to add to the ActionBar.',
+    buttonContainerClassNames: 'Classnames to spread to the back Button container.',
+    description: 'Localized text for the description.',
+    descriptionProps: 'Additional props to be spread to the description\'s `<p>` element.',
+    headingLevel: 'Heading level. `<h1>` is reserved for the page title.',
```
skvale: if h1 is reserved, should the propType be `CustomPropTypes.range(2, 6),` then?

src/ActionBar/ActionBar.js
```diff
@@ -1,42 +1,121 @@
...
+
+ActionBar.defaultProps = {
+    headingLevel: 3
+};
+
+ActionBar.propDescriptions = {
+    actionClassNames: 'Classnames to spread to the action Button container.',
+    actionProps: 'Props to spread to the action Button container',
+    actions: 'Button components to add to the ActionBar.',
+    buttonContainerClassNames: 'Classnames to spread to the back Button container.',
```
skvale: Seems weird to have `actionClassNames` and `buttonContainerClassNames` plural when we use `className` which isn't plural for the same idea.

skvale: Looks like there is a mixture of both ways of doing it in fundamental-react

```
actionClassNames
buttonContainerClassNames
labelClassNames
inputClassNames
listClassNames
tokenizerClassNames

vs

backdropClassName
contentClassName
inputClassName
popperClassName
tableBodyClassName
table....ClassName
referenceClassName
```

src/ActionBar/ActionBar.js
```diff
@@ -1,42 +1,121 @@
...
+    description: 'Localized text for the description.',
+    descriptionProps: 'Additional props to be spread to the description\'s `<p>` element.',
+    headingLevel: 'Heading level. `<h1>` is reserved for the page title.',
+    onClick: 'Callback to pass to the back Button container.'
```
skvale: isn't it passed to the button and not its container?

greg-a-smith: There is no subcomponent anymore.  That said, should we name this prop something to reflect that it's for the back button (_e.g._ `onBackClick`)?

src/ActionBar/ActionBar.js
jbadan: I think that came from a bunch of different people implementing it. We should do a separate story to clean it up in one go. 

I'll make this one singular for now.

jbadan: https://github.com/SAP/fundamental-react/issues/911

jbadan: Now that I look at the history on it.... it's only me that's been adding the plural ones 😆🤣

PR #909, opened 2020-03-03
fix: add DatePicker validation states, fix validation states for popover-based components

jacobdevera:
### Description

- adds validation states to DatePicker
- fixes click/click outside/esc handling for validation states for popover-based components


src/Shellbar/__snapshots__/Shellbar.test.js.snap
```diff
@@ -234,10 +234,10 @@ exports[`<Shellbar /> create shellbar 3`] = `
...
                 />
                 <span
-                  className="fd-input-group__addon fd-input-group__addon--button"
+                  className="fd-shellbar__input-group__addon fd-input-group__addon fd-input-group__addon--button"
                 >
                   <button
-                    className="fd-button fd-button--light sap-icon--search fd-input-group__button"
+                    className="fd-button fd-button--light sap-icon--search fd-shellbar__button fd-input-group__button"
                     onClick={[Function]}
                     type="button"
```
jacobdevera: I realize the `fd-shellbar__button` specificity has been addressed in styles—just need to make sure the class is added in the first place.

PR #918, opened 2020-03-09
fix:#631 Added condition for default value

### LINKED ISSUE #631: SearchInput component not support {value: 'params'} in the inputProps users passes
SearchInput component render the input element with spreading the inputProps, but if inputProps is {value:'test'}, it won't work since it will be overrides by value={this.state.value}, I am wondering whether the initial state value in the constructor can check if users passes the value in the inputprops, not directly set it to empty string.

singh-mahendra:
### Description
Added condition for default value

fixes #631


src/SearchInput/SearchInput.js
```diff
@@ -15,12 +15,16 @@ class SearchInput extends Component {
         this.state = {
             isExpanded: false,
             searchExpanded: false,
-            value: '',
-            searchList: this.props.searchList,
-            filteredResult: this.props.searchList
+            value: props.inputProps && props.inputProps.value ? props.inputProps.value : '',
```
jbadan: We can simplify this to `props.inputProps?.value ? props.inputProps.value : ''` - same with line 20 :)

singh-mahendra: Updated.Thanks.

PR #923, opened 2020-03-16
fix: DatePicker custom formats and auto formatting

prsdthkr:
### Description
 * adds custom date format for date field with property `dateFormat`
 * sane auto-formatting of input values to `dateFormat` using [Moment.js multiple formats API](https://momentjs.com/docs/#/parsing/string-formats/)
* added support for 2 digits year input
* sane autoformatting for date range picker
* more specific date range placeholder i.e. `YYYY-MM-DD - YYYY-MM-DD`. This would help anyone wanting to just type it out. 

**Demo**
ACC-2058-demo


src/DatePicker/DatePicker.js
```diff
@@ -28,6 +29,63 @@ class DatePicker extends Component {
...
+     * It will return an empty string if Moment.js is unable to create a valid
+     * date object from given string (for e.g. text only strings).
+     * @param  {String|Object} date to format.
+     * @returns {String} formatted date string.
+     */
+    getFormattedDateStr = (date) => {
+        if (date) {
+            const momentDateObj = this.getMomentDateObj(date);
+            if (momentDateObj && momentDateObj.isValid()) {
+                const { dateFormat } = this.props;
```
jacobdevera: If the user provides a locale but not a dateFormat, I would think that the locale's should be used. Right now because there's a default set (ISO), it would always override the locale (e.g. look at the localized datepickers on the docs site which are normally DD/MM/YYYY).

src/DatePicker/DatePicker.js
```diff
@@ -28,6 +29,63 @@ class DatePicker extends Component {
...
+    }
+    /**
+     * Get the different date formats allowed for user input.
+     * These formats are based on props.dateFormat
+     * If no entry for props.dateFormat found in saneDateLookup,
+     * it returns the confiured format as string.
+     *
+     * @returns {Array} collection of date formats allowed for input.
+     */
+    getSaneFormats = () =>{
```
meganaconley: It's used throughout the files but I'm not sure about `sane`. Maybe `valid` would be more clear?

jbadan: We should include one insane dateFormat- `mm-YYYY/dd` 🤣

bcullman: this is my bad - i started calling these "sane" formats starting 2 weeks ago. Not sure where it came from.

src/DatePicker/DatePicker.js
prsdthkr: Thanks, that's correct.

src/DatePicker/DatePicker.js
jbadan: I mean.. you're not wrong 🤷‍♀

jbadan:
Is this a breaking change or a new feature? If it's just a feature we can use `fix` instead. I know.. I know.. one day we will upgrade standard-version and this won't be a thing we have to worry about 😃

prsdthkr:
> Is this a breaking change or a new feature? If it's just a feature we can use `fix` instead. I know.. I know.. one day we will upgrade standard-version and this won't be a thing we have to worry about 😃

There is a visual change to the date range picker's placeholder text. Does that qualify as a breaking change?

PR #924, opened 2020-03-17
fix: expand DatePicker capabilities

meganaconley:
### Description
Adds support for the following (with unit tests):
- updating `defaultValue` prop
- `disabled` and `readOnly` mode
- added prop to spread to inner `Calendar` component
- `onChange` and `onFocus` callback
-  provide ISO-8601 formatted date

No visual changes.


src/DatePicker/DatePicker.js
```diff
@@ -247,23 +295,36 @@ DatePicker.propTypes = {
...
     locale: PropTypes.string,
+    readOnly: PropTypes.bool,
     validationState: PropTypes.shape({
         state: PropTypes.oneOf(FORM_MESSAGE_TYPES),
         text: PropTypes.string
     }),
-    onBlur: PropTypes.func
+    onBlur: PropTypes.func,
+    onChange: PropTypes.func,
+    onFocus: PropTypes.func
```
jbadan: LOL how was anyone actually using this thing before 😆

src/DatePicker/DatePicker.test.js
```diff
@@ -268,6 +292,31 @@ describe('<DatePicker />', () => {
...
+            element
+                .find('input[type="text"]')
+                .simulate('change', { target: { value: '04/14/2020' } });
+
+            expect(change).toHaveBeenCalledWith(expect.objectContaining({ formattedDate: '04/14/2020' }));
+        });
+    });
+
+    describe('onFocus callback', () => {
+        test('should call onFocus on entering input text', () => {
```
jacobdevera: Was the wording meant to be switched between these? i.e. 'should call onChange on entering input text', 'should call onFocus on focusing input'.

meganaconley: Good catch, I swapped these around at some point.

PR #927, opened 2020-03-18
fix: disable trigger popover on click for datepicker

jacobdevera:
### Description
Clicking the DatePicker will automatically focus the Calendar upon opening the Popover which isn't always desired if the user wants to directly edit the text field. Adds a prop to Popover disable the default opening of the Popover body when clicking the `control` element.

### DIFF
--- a/src/Popover/Popover.js
+++ b/src/Popover/Popover.js
@@ -94,6 +94,7 @@ class Popover extends Component {
             disableEdgeDetection,
             disableKeyPressHandler,
             disableStyles,
+            disableTriggerOnClick,
             onClickOutside,
             onEscapeKey,
             disabled,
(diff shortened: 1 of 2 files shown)


src/Popover/Popover.js
```diff
@@ -94,6 +94,7 @@ class Popover extends Component {
             disableEdgeDetection,
             disableKeyPressHandler,
             disableStyles,
+            disableTriggerOnClick,
```
bcullman: looking at https://sap.github.io/ui5-webcomponents/playground/components/DatePicker/, the default seems to be to not open the calendar (and that is how our timepicker works, as well as  as well)

so i think this prop might be backwards?

jacobdevera: I was thinking more in terms of what the default should be for a Popover. Normally the `control` component of a Popover is a Button, Menu, etc. and you would expect it to open by default when clicking.

But if it is a more complex component like InputGroup with both an input field and a Button, it makes sense to override it and not have the whole `InputGroup` trigger opening the `Popover` when clicked, especially with DatePicker because it steals focus from the input field.

jbadan: According to the design wiki: "The date picker has two components: the date input field and the date picker button. On all devices users can either use the input field to type a date **or use the date picker button to open the date picker calendar.**" 

The default should probably match fiori3 in here.

bcullman: i read this is as, 'by default, the calendar component of the datepicker *only* opens when you interact with the input group's button'.

jacobdevera: It makes sense for that to be the default behavior for DatePicker, but this prop is being defined in Popover. So I'm not sure if it makes sense to reverse this prop and enable the trigger click for all the other Popoover-based components.

jbadan: Yeah I think we're all saying the same thing now haha

PR #932, opened 2020-03-25
feat: remove formset component

jbadan:
### Description

Remove deprecated FormSet component. It was replaced by FormFieldset.

fixes #issueid


src/Forms/Forms.Component.js
```diff
@@ -102,7 +102,7 @@ export const FormsComponent = () => {
                     is displayed below the input field.`}
                 title='Input Help Elements'>
                 <>
-                    <FormSet>
+                    <FormGroup>
```
skvale: Could these be replaced with `<FormFieldSet` instead of `<FormGroup`?

jbadan: Do they need to be a FormFieldSet?

skvale: nope!  I didn't mean to suggest they should be. I just read the PR description and saw
> It was replaced by FormFieldset

So I thought I'd ask

PR #936, opened 2020-03-27
fix: datepicker onBlur, empty validation, empty isoFormattedDate

jacobdevera:
### Description

Fixes bugs:
- onBlur would run on click outside and escape key, triggering validation prematurely
- empty validation object still causes FormMessage to render; adjust so only render if text has length
- isoFormattedDate would always return 'invalid date' on empty dates instead of empty string (desired)


src/DatePicker/DatePicker.test.js
```diff
@@ -298,33 +292,11 @@ describe('<DatePicker />', () => {
...
-                .simulate('click');
-
-
-            //trigger onBlur by clicking outside
-            simluateOutsideMouseClick();
-
-            expect(blur).toHaveBeenCalledTimes(1);
-
-            expect(blur).toHaveBeenCalledWith(expect.objectContaining({ date: expect.any(moment) }));
-        });
```
jacobdevera: Don't think this test/story is necessary, and will cause validation issues if onBlur runs validation prematurely because of an outside click.

PR #937, opened 2020-03-27
fix: search addonBefore error and popover out boundaries

### LINKED ISSUE #931: Search Input getting error in the console ,not recognize the `addonBefore &  x-out-ofboundaries
### Description

**1. when opening the search popover** 
1540 Warning: React does not recognize the `addonBefore` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `addonbefore` instead. If you accidentally passed it from a parent component, remove it from the DOM element....
Screen Shot 2020-03-25 at 7 36 47

**2. when scrolling out of the screen with an open popover.**
Warning: Received `true` for a non-boolean attribute `x-out-of-boundaries`.
If you want to write it to the DOM, pass a string instead: x-out-of-boundaries="true" or x-out-of-boundaries={value.toString()}...
image001
 
Thanks 
Ben Oron 
Front End Developer, 
SAP Fieldglass,

### Versions
**fundamental-react:**
0.8
---

_**DISCLAIMER:**

smohtadi:
### Description
Fixes :
- Search error when popover is triggered. Cause by li element being created in SearchInput and passed to MenuList were it gets addonBefore prop appended. Used Menu.Item instead.
- Popover error when open and window scrolls down.

fixes #931

### DIFF
--- a/src/utils/_Popper.js
+++ b/src/utils/_Popper.js
@@ -122,7 +122,7 @@ class Popper extends React.Component {
                             ref={ref}
                             style={{ ...style, ...popperProps.style }}
                             // eslint-disable-next-line no-undefined
-                            x-out-of-boundaries={!!outOfBoundaries || undefined}
+                            x-out-of-boundaries={!!outOfBoundaries ? 'true' : undefined}
                             x-placement={placement}>
                             {children}
                             <span
--- a/src/SearchInput/SearchInput.js
+++ b/src/SearchInput/SearchInput.js
@@ -133,19 +133,18 @@ class SearchInput extends Component {
                     {this.state.filteredResult && this.state.filteredResult.length > 0 ? (
                         this.state.filteredResult.map((item, index) => {
                             return (
-                                <li
-                                    className='fd-menu__item'
+                                <Menu.Item
                                     key={index}
                                     onClick={(e) => this.handleListItemClick(e, item)}>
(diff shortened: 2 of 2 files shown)


src/utils/_Popper.js
```diff
@@ -122,7 +122,7 @@ class Popper extends React.Component {
                             ref={ref}
                             style={{ ...style, ...popperProps.style }}
                             // eslint-disable-next-line no-undefined
-                            x-out-of-boundaries={!!outOfBoundaries || undefined}
+                            x-out-of-boundaries={!!outOfBoundaries ? 'true' : undefined}
```
jbadan: Would this work if we just had  `x-out-of-boundaries={!!outOfBoundaries}`?

jbadan:
@smohtadi thank you for your contribution! Can you sign the CLA agreement linked by the comment? If you have signed it and it's not working, check that the email address you use for git commits is the same as the email address associated with your account.

src/utils/_Popper.js
smohtadi: I tried it, and it throws an error: "If you used to conditionally omit it with x-out-of-boundaries={condition && value}, pass x-out-of-boundaries={condition ? value : undefined} instead." It seems  x-out-of-boundaries expects either a string when true or an undefined when false.

PR #941, opened 2020-03-31
fix: Close SearchInput popover after selecting an item.

### LINKED ISSUE #938: search input not closed after select item
### Description

search input not closed after select item 
### Versions
**fundamental-react:**
0.8
---

_**DISCLAIMER:**

BenOron:
### Description

fixes #938

### DIFF
--- a/src/Popover/Popover.js
+++ b/src/Popover/Popover.js
@@ -53,7 +53,7 @@ class Popover extends Component {
         if (this.state.isExpanded && this.popover) {
             this.focusManager = new FocusManager(this.popover, this.controlRef, this.props.useArrowKeyNavigation);
         }
-    }
+    };
 
     handleOutsideClick = () => {
         if (this.state.isExpanded) {
@@ -74,7 +74,7 @@ class Popover extends Component {
                 firstTabbableNode && firstTabbableNode.focus();
             }
         }
-    }
+    };
 
     handleKeyPress = (event, node, onClickFunctions) => {
         if (!this.isButton(node)) {
@@ -87,7 +87,7 @@ class Popover extends Component {
                 default:
             }
         }
-    }
+    };
 
     render() {
         const {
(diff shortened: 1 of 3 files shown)


src/Popover/Popover.js
```diff
@@ -165,7 +166,7 @@ class Popover extends Component {
                     popperProps={{ ...popperProps, id }}
                     referenceClassName='fd-popover__control'
                     referenceComponent={referenceComponent}
-                    show={this.state.isExpanded && !disabled}
+                    show={typeof this.props.show === 'boolean' ? this.props.show && !disabled : this.state.isExpanded && !disabled}
```
skvale:
```js
show={!disabled && (typeof show === 'boolean' ? show : this.state.isExpanded)}
```
To simplify it a little bit, we could take out the repeated `!disabled` as well as use `show` instead of `this.props.show` since it is already deconstructed.

benoron: on it

src/SearchInput/SearchInput.js
```diff
@@ -32,6 +32,11 @@ class SearchInput extends Component {
     };
 
     handleListItemClick = (event, item) => {
+        this.setState({
+            value: item.text,
+            isExpanded: false,
+            searchExpanded: false
+        });
         item.callback ? item.callback() : null;
```
skvale, line 40: I see let calling the onChange callback after setState, we could follow the same pattern.

Also, if you are setting the value in state, it looks like you need to recompute the filteredResult, like this:

```js
        this.setState((prevState) => ({
            value: item.text,
            isExpanded: false,
            searchExpanded: false,
            filteredResult: this.filterList(prevState.searchList, item.text)
        }), () => {
            if (item.callback) {
                item.callback();
            }
        });     
```

Otherwise it can show every item in the list prefixed with the selected text

Screen Shot 2020-04-02 at 10 42 56 AM

instead of

Screen Shot 2020-04-02 at 10 44 12 AM

BenOron: Hi @skvale  and thanks for the review I pushed a new fix.

PR #955, opened 2020-04-02
feat: added textarea support for maxLength prop

### LINKED ISSUE #880: FormTextarea: build out counter on component `fd-textarea-counter`
### Description
https://github.com/SAP/fundamental-styles/pull/419
https://fundamental-styles.netlify.com/components/form.html
Screen Shot 2020-02-20 at 4 39 02 PM

### Versions
**fundamental-react:**

---

_**DISCLAIMER:**

smohtadi:
### Description
Feature:
- Added text area counter component
- Added an onchange event on the text area that updates the counter

fixes #880


skvale:
Thank you for your pull request, but I think this is a customization that doesn't belong to this component library.

In your own codebase, you could create this wrapper around the Fundamental React Textarea, but I don't see it as functionality that this library should own.

jbadan:
@skvale This is part of fundamental-styles/fiori3 spec - see https://fundamental-styles.netlify.com/components/form.html#textarea

src/Forms/FormTextarea.js
```diff
@@ -37,10 +56,19 @@ FormTextarea.propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
     compact: PropTypes.bool,
+    counter: PropTypes.number,
```
skvale, line 59: HTML `textarea` elements have existing support for the `maxLength` prop.  

We could re-use that language as it may be more familiar to consumers.

src/Forms/Forms.Component.js
```diff
@@ -50,6 +50,15 @@ export const FormsComponent = () => {
...
                     </FormGroup>
+
+                    <FormGroup>
+                        <FormItem>
+                            <FormLabel htmlFor='textarea-2'>
+                                Text area with counter
+                            </FormLabel>
+                            <FormTextarea id='textarea-2' maxLength={150} />
+                        </FormItem>
+                    </FormGroup>
```
skvale, line 61: This is looking great.
Could we add a corresponding Storybook example for this
```jsx
//FormTextArea.stories.js
    .add('Max length', () => (
        <FormTextarea {...createProps({maxLength: 150})}>Max Length</FormTextarea>
    ))
```

skvale:
It looks like the build is failing because
> Package size limit has exceeded by 11 B

`.size-limit` just needs to be bumped, something like
```
limit: "190 KB"
```

src/Forms/FormTextarea.js
```diff
@@ -1,16 +1,58 @@
...
+        return 0;
+    };
+
+    const hasMaxLength = maxLength >= 0;
+
+    const [charCount, setCharCount] = useState(hasMaxLength ? getInitialCharCount() : 0);
+
+    const getMaxLengthText = () => {
+        let charsLeft = Math.max(maxLength - charCount, 0);
+        return `${charsLeft} character${charsLeft !== 1 ? 's' : ''} left`;
```
skvale, line 46: I don't think our current way of handling allows for strings with arguments to conditionally render plural, but to Jenna's earlier point of using localizedText, this could be
 
```js
    const getMaxLengthText = () => {
        const {localizedText} = this.props;
        const {charactersLeftPlural, charactersLeftSingular} = localizedText;

        let charsLeft = Math.max(maxLength - charCount, 0);
	    if (charsLeft !== 1) {
            return `${charsLeft} ${charactersLeftPlural}`;
        }
        return `${charsLeft} ${charactersLeftSingular}`;
    }

// ...

TextArea.propTypes = {
   // ...
   localizedText: CustomPropTypes.i18n({
      charactersLeftPlural: PropTypes.string,
      charactersLeftSingular: PropTypes.string,
   })
}

TextArea.defaultProps = {
   // ...
   localizedText: {
      charactersLeftPlural: 'characters left',
      charactersLeftSingular: 'character left',
   }
}
```

src/Forms/FormTextArea.test.js
```diff
@@ -39,4 +49,87 @@ describe('<FormTextArea />', () => {
         expect(ref.current.tagName).toEqual('TEXTAREA');
         expect(ref.current.className).toEqual('fd-textarea');
     });
+
+    describe('FormTextArea counter', () => {
+        const setup = (props, children) => {
+            return mount(<FormTextarea {...props}>
+                {children}
```
skvale, line 56: I see warnings running these tests
```
    Warning: Use the `defaultValue` or `value` props instead of setting children on <textarea>.
        in textarea (created by FormTextarea)
        in FormTextarea (created by WrapperComponent)
        in WrapperComponent
```
it could be changed to

```jsx
        const setup = (props, defaultValue) => {
            return mount(<FormTextarea {...props} defaultValue={defaultValue} />)
        }
```
to get rid of the warnings

smohtadi:
Thanks! And thank you so much for all your help!

PR #958, opened 2020-04-03
feat: add onSelect callback to ComboboxInput;Select;SearchInput

### LINKED ISSUE #921: Select component doesn't update the option on click
### Description

When you click on one of the list items inside a `Select` component, it fails to reflect the selection.

You can see the behaviour in the docs: https://sap.github.io/fundamental-react/Select
There's seems to be no api to update the selection.

### Versions
**fundamental-react:** 0.8.0

---

_**DISCLAIMER:**

skvale:
### Description

Add an onSelect callback event for `ComboboxInput`, `Select`, and `SearchInput` and reflect the changes in the `ComboboxInput` and `Select` components.

[BREAKING]
- ComboboxInput no longer takes a `list` prop, instead, it takes an options array of shape `{key, text}`
- Select no longer takes children, instead, it takes an options array of shape `{key, text}`

fixes #921


src/Select/__snapshots__/Select.test.js.snap
```diff
@@ -269,22 +269,23 @@ exports[`<Select /> create Select component 7`] = `
...
-        uvalidationState={
-          Object {
-            "state": "success",
-            "text": "Test validation state",
-          }
-        }
       >
         <div
-          className="fd-select__control"
+          className="fd-select__control is-success"
```
skvale, line 274: This snapshot is changed because the previous test misspelled `validationState` as `uvalidationState`

nickbaek:
Hi team, I just wonder how can we retrieve the value that we have selected?

Issue #879, opened 2020-02-21
BusyIndicator: new component


jbadan:
### Description
https://github.com/SAP/fundamental-styles/pull/412
https://fundamental-styles.netlify.com/components/busy-indicator.html
![Screen Shot 2020-02-20 at 4 37 08 PM](https://user-images.githubusercontent.com/29607818/74993270-52415280-53ff-11ea-83a1-081034cb6fca.png)

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

jbadan:
related to https://github.com/SAP/fundamental-react/issues/797

jbadan:
fixed by https://github.com/SAP/fundamental-react/pull/963

PR #963, opened 2020-04-10
feat: New Busy Indicator component

JaySardhara:
### Description
Added Busy Indicator component

Implements #879


src/BusyIndicator/BusyIndicator.js
```diff
@@ -0,0 +1,63 @@
...
+
+    const busyIndicatorClasses = classnames(
+        {
+            [`fd-busy-indicator--${size}`]: !!size
+        },
+        className
+    );
+
+    if (!show)
+        return (<></>);
```
skvale, line 30: Should we do
```jsx
if (!show) {
  return null;
}
```
to match the style of other components in fundamental-react?

src/BusyIndicator/BusyIndicator.js
```diff
@@ -0,0 +1,63 @@
...
+    useEffect(() => {
+        if (!disableStyles) {
+            require('fundamental-styles/dist/icon.css');
+            require('fundamental-styles/dist/busy-indicator.css');
+        }
+    }, []);
+
+    const busyIndicatorClasses = classnames(
+        {
+            [`fd-busy-indicator--${size}`]: !!size
```
skvale, line 24: I am looking at https://fundamental-styles.netlify.com/components/busy-indicator.html
and I notice the small busy indicator class doesn't have a `--s` on it.  If you toggle between `fd-busy-indicator` and `fd-busy-indicator--s` there are style differences.

Maybe that linked doc is out of date, but I think there should be a special case for `size === 's'` here

src/BusyIndicator/BusyIndicator.js
```diff
@@ -0,0 +1,63 @@
...
+    );
+
+    if (!show)
+        return (<></>);
+
+    return (
+        <div
+            {...props}
+            className={busyIndicatorClasses}
+            ref={ref} >
```
skvale, line 36: [a11y]
Should we default an aria-label attribute?  Something like
```js
 aria-label={localizedText.loading}
```
and should `aria-hidden="false"` be applied by default?

src/BusyIndicator/BusyIndicator.js
JaySardhara: Noted. Let me incorporate above suggestion.

src/BusyIndicator/BusyIndicator.js
skvale: @meganaconley ^

JaySardhara: Added above suggestions.
Did not add `aria-hidden` in propDescription as its implicit.

src/BusyIndicator/BusyIndicator.js
JaySardhara: Added above suggested changes.

src/BusyIndicator/BusyIndicator.js
JaySardhara: Added above suggested changes.

src/BusyIndicator/BusyIndicator.js
```diff
@@ -0,0 +1,77 @@
...
+    localizedText: {
+        loading: 'Loading'
+    },
+    size: 'm',
+    show: false
+};
+
+BusyIndicator.propDescriptions = {
+    show: 'Set to **true** to make Busy Indicator visible',
+    localizedTextShape: {
```
skvale, line 72: This should be `localizedText` instead of `localizedTextShape`

src/BusyIndicator/BusyIndicator.js
```diff
@@ -0,0 +1,77 @@
...
+            <div className='fd-busy-indicator--circle-1' />
+            <div className='fd-busy-indicator--circle-2' />
+        </div>
+    );
+});
+
+BusyIndicator.displayName = 'BusyIndicator';
+
+BusyIndicator.propTypes = {
+    children: PropTypes.node,
```
skvale, line 52: It looks like you can take children out of the propTypes, it isn't used

src/BusyIndicator/BusyIndicator.js
JaySardhara: @skvale description is not rendered if `localizedText` is used in propDescription. 
image

src/BusyIndicator/BusyIndicator.js
JaySardhara: Done.

JaySardhara:
Thanks for review @skvale

PR #964, opened 2020-04-14
feat:  today actions for datepicker

prsdthkr:
### Description
This change adds the ability to configure a `todayAction` for the DatePicker.

 For example, ```todayAction={{type: 'select', label: 'Today'}}```

**todayAction.type** is a string indicating the type of today button.

- `'none'` (default) today button won't be shown
- `'select'` today button as footer action that selects today's date and closes datepicker
- `'navigate'` today button as header action that navigates (i.e. sets focus) to today's date

**todayAction.label** is a localized string label for the today action button.

The button will only be rendered if:

- `todayAction.type` is `'select'` or `'navigate'`
- AND `todayAction.label` is a valid non-empty string

Other changes:
- Add compact mode for Calendar
- Use calendar compact mode for compact Datepicker

DatePicker today actions


bcullman:
should there be a boolean to `enableTodaySelect` here?

prsdthkr:
> should there be a boolean to `enableTodaySelect` here?

For this PR if `todayLabel` is specified it will automatically be used to enable selection. I'm going to create a follow-up after https://github.com/SAP/fundamental-react/pull/1094 is merged , that would allow us to use it for either selection or navigation.

jbadan:
@prsdthkr are you waiting for a review on this or it is it progress?

prsdthkr:
> @prsdthkr are you waiting for a review on this or it is it progress?

@jbadan await review.

src/Calendar/Calendar.js
```diff
@@ -453,6 +453,7 @@ class Calendar extends Component {
             this.props.localizedText.show12PreviousYears : this.props.localizedText.previousMonth;
         const nextButtonLabel = this.state.showYears ?
             this.props.localizedText.show12NextYears : this.props.localizedText.nextMonth;
+        const compact = this.props?.compact || false;
```
skvale, line 456: `this.props` should never be undefined and is this necessary or can we just use `this.props.compact` below?

src/DatePicker/DatePicker.js
```diff
@@ -37,6 +37,16 @@ class DatePicker extends Component {
         this.popoverRef = React.createRef();
     }
 
+    componentDidMount() {
+        const { disableStyles } = this.props;
+
+        if (!disableStyles && this._showTodayFooter()) {
+            //below styles are needed for footer styling
+            require('fundamental-styles/dist/dialog.css');
+            require('fundamental-styles/dist/bar.css');
```
skvale, line 46: I think this is the old way of including styles, now they're expected to be imported with the rest of the imports without conditionals.

src/Calendar/__stories__/Calendar.stories.js
```diff
@@ -96,6 +98,7 @@ export const dev = () => (
             dateKnobToDate('disable between dates (2)', disabledDateSecondDefault)]}
         locale={text('locale', 'en')}
         openToDate={dateKnobToDate('open to date', new Date())}
+        showToday={boolean('showToday', false)}
```
skvale, line 101: could add compact to as a knob to the dev story as well

src/DatePicker/DatePicker.js
```diff
@@ -309,6 +311,28 @@ class DatePicker extends Component {
...
+                && todayLabel
+                && typeof todayLabel === 'string'
+                && todayLabel.trim().length > 0;
+    }
+
+    _showTodayFooter = () => {
+        const { enableRangeSelection, todayAction: { label: todayLabel, type: todayType } } = this.props;
+        return todayType === 'select'
+                && !enableRangeSelection
+                && isEnabledDate(moment(), this.props)
```
skvale, line 326: Why do we not show the footer button if
```js
&& !enableRangeSelection
&& isEnabledDate(moment(), this.props)
```
but we don't consider these things for the header button?

src/Calendar/Calendar.js
```diff
@@ -678,7 +679,10 @@ class Calendar extends Component {
 
         const calendarClasses = classnames(
             'fd-calendar',
-            className
+            className,
+            {
+                'fd-calendar--compact': compact
+            }
```
skvale, line 685: It's possible className would override a css rule from `rd-calendar--compact` and it looks like the convention in fundamental-react is to keep the custom className last

```js
        const calendarClasses = classnames(
            'fd-calendar',
            {
                'fd-calendar--compact': compact
            },
            className
        );
```

src/DatePicker/DatePicker.js
```diff
@@ -350,6 +375,16 @@ class DatePicker extends Component {
...
 
+        const datepickerFooterClassName = classnames(
+            'fd-dialog__footer',
+            'fd-bar',
+            'fd-bar--footer',
+            {
+                'fd-bar--cosy': !compact,
+                'fd-bar--compact': compact
+            }
+        );
```
skvale, line 386: should we add a prop so a consumer can add their own className to this?  `footerClassName`?

src/DatePicker/DatePicker.js
```diff
@@ -309,6 +311,28 @@ class DatePicker extends Component {
         this.validateDates(this.props.onBlur);
     };
 
+    _showTodayHeader = () => {
```
skvale, line 314: let's remove the `_` these would be the only class methods that have them in the repo

src/DatePicker/DatePicker.js
prsdthkr: Because clicking the footer today button selects today's date and closes the popover, I think this would not be a valid interaction if a date range is to be selected or if today's date is disabled. Also, since the header today button only performs navigation we can show it regardless. What do you think?

src/Calendar/__stories__/Calendar.stories.js
prsdthkr: Added.

src/Calendar/Calendar.js
prsdthkr: Changed.

src/DatePicker/DatePicker.js
prsdthkr: Changed.

PR #966, opened 2020-04-16
fix: Blur event validates time in the TimePicker input 

### LINKED ISSUE #965: On the TimePicker component, whenever focus is lost from the field, the field gets cleared
### Description
On the TimePicker component, if there is a default time or when we set a time, if we click elsewhere such that focus is lost from the field, the field gets cleared.

### Versions
**fundamental-react:**v0.8.1

---

_**DISCLAIMER:**

coding-my-passion:
### Description
Added new isValidTime function to check if entered time is correct. 

fixes #965


src/TimePicker/_TimePickerItem.js
```diff
@@ -201,10 +201,38 @@ class TimePickerItem extends Component {
         }
     };
 
-    onBlur = () => {
-        const { isValid } = this.state;
+    isValidTime = (value) => {
```
jbadan, line 204: What is different here than `onInputValidation`? Would it make sense it reuse that function/improve it instead of reinventing it?

coding-my-passion: Thanks for your review comment.
regex for few below cases in onInputValidation were not matching input correctly.
eg.
a. 12-Hour Clock - (hh:mm:ss am/pm)
-> Here user can enter hour >= 13 and minute/second >= 60
b. 24-Hour Clock - (hh:mm:ss)
-> Here user can enter hour >= 24 and minute/second >= 60
c. Time Picker With No Seconds - (hh:mm)
-> Here user can enter minute/second >= 60
As you suggested I have modified the code to use onInputValidation function for change as well as blur event type.
Please review.

PR #967, opened 2020-04-20
fix: add backbutton to shellbar

auryn31:
### Description

add backbutton to shellbar


src/Shellbar/Shellbar.js
```diff
@@ -121,6 +122,13 @@ class Shellbar extends Component {
         return (
             <div className={shellbarClasses}>
                 <div className='fd-shellbar__group fd-shellbar__group--product'>
+                    {backAction && <Icon
+                        className='fd-shellbar__group fd-button--back'
+                        glyph='nav-back'
+                        onClick={backAction}
+                        size='l'
+                        style={{ color: 'white', marginRight: '0.5em', flex: 'none' }} />
```
jbadan, line 130: We don't like to include inline styles - is there something missing in fundamental-styles? If so, we need to make a pull request there. If the style is available in fundamental-styles, but fundamental-react isn't on the correct version, we can update.

src/Shellbar/Shellbar.js
```diff
@@ -121,6 +122,12 @@ class Shellbar extends Component {
         return (
             <div className={shellbarClasses}>
                 <div className='fd-shellbar__group fd-shellbar__group--product'>
+                    {backAction && <Button
+                        className='fd-shellbar__button'
```
jbadan, line 126: We should add an aria-label here for "Back" - see the `localizedText` prop below and copy what's done with `counterLabel`. 

Also, `npm run test:visual:update` should get you squared away with travis! 🎉

PR #973, opened 2020-04-23
fix: call onChange only if formattedDate has changed

prsdthkr:
### Description
Currently `this.props.onChange` is called from multiple places without checking if the `formattedDate` in the state has truly changed. This leads to `this.props.onChange` being called with the same payload multiple times. To prevent this we should explicitly check if formattedDate has changed before calling `this.props.onChange` with the new payload.


jacobdevera:
Not to say this change is wrong, but I'm having a hard time reproducing it. I've been logging, setting breakpoints, etc. but any `onChange` function I set or any of the related functions (updateDate, resetState, validateDates) only seem to be called once whether I'm typing in dates or selecting a date with the calendar.

prsdthkr:
> Not to say this change is wrong, but I'm having a hard time reproducing it. I've been logging, setting breakpoints, etc. but any `onChange` function I set or any of the related functions (updateDate, resetState, validateDates) only seem to be called once whether I'm typing in dates or selecting a date with the calendar.

Oops! forgot to add before/after GIF. Added.

For example, onChange is fired after onBlur with the same data, or even after selecting the same date again.

jacobdevera:
Okay, I can see those cases, that makes sense to prevent.

PR #1017, opened 2020-05-14
chore: visual testing improvements + reduce travis build time

jbadan:
### Description

Added a script to programmatically create an additional stories file that includes all stories on 1 page - this speeds up visual regression and snapshot testing.

fixes #issueid


devtools/buildVisualStories.js
```diff
@@ -0,0 +1,63 @@
...
+import * as stories from './${componentName}.stories';
+
+export default {
+    title: 'Visual'
+};
+
+export const ${componentName} = () => {
+    let storyNames = Object.keys(stories).filter(story => story !== 'default');
+
+    return (<>{storyNames.map(item => stories[item]())}</>);
```
skvale, line 49: I was looking at the Token examples
Screen Shot 2020-05-18 at 9 16 09 AM

Do you think these could be wrapped in a `div` here so they show up on new lines?

(separate issue: that readonly token shouldn't say Compact...)

jbadan: Good idea! We also do need to do an audit and cleanup a lot of the examples for things like the labels

devtools/buildVisualStories.js
```diff
@@ -0,0 +1,63 @@
...
+            // https://github.com/storybookjs/storybook/releases/tag/v6.0.0-alpha.43
+            if (componentName === 'Calendar'
+            || componentName === 'Dialog'
+            || componentName === 'Table'
+            || componentName === 'Popover'
+            || componentName === 'Shellbar') {
+                return;
+            }
+
+            fileContents += `
```
skvale, line 38: Maybe we could do
```js
let fileContents = `
import React from 'react';
export default {
    title: 'Visual'
};
`
```
Then
```js
fileContents += `
import * as stories from './${componentName}.stories';
export const ${componentName} = () => {
    let storyNames = Object.keys(stories).filter(story => story !== 'default');

    return (<>{storyNames.map(item => stories[item]())}</>);
};
${componentName}.story = {
    parameters: {
        docs: { disable: true }
    }
};
`
```
Otherwise we would import React more than once and that would throw an error

jbadan: I'm confused - this works as it is? Where are you seeing it import react more than once?

jbadan: Oh I think I see where the confusion is - I'm not making one file here - I'm making one file per component. So I need that import inside the loop

skvale: Oh, that makes sense.  Then componentName and fileContents could just be local variables in the inner loop scope
```
componentDirs.map((directory) => {
    // Loop through its files.
    directory.fileNames.map((fileName) => {
        // get only stories.js files
        if (fileName.includes('.stories.js')) {
        // Grab the component name
            const componentName = fileName.substr(0, fileName.indexOf('.'));
            // TO DO: reenable storyshots for examples using hooks in storybook@6
            // https://github.com/storybookjs/storybook/releases/tag/v6.0.0-alpha.43
            if (componentName === 'Calendar'
            || componentName === 'Dialog'
            || componentName === 'Table'
            || componentName === 'Popover'
            || componentName === 'Shellbar') {
                return;
            }

            const fileContents = `
import React from 'react';
import * as stories from './${componentName}.stories';

export default {
    title: 'Visual'
};

export const ${componentName} = () => {
    let storyNames = Object.keys(stories).filter(story => story !== 'default');

    return (<>{storyNames.map(item => <div>{stories[item]()}</div>)}</>);
};
${componentName}.story = {
    parameters: {
        docs: { disable: true }
    }
};
`;
            // write the visual story file into the directory.
            let visualPath = path.join(directory.path, `${componentName}.visual.js`);
            writeFileSync(visualPath, fileContents);
        }
    });
});
```

devtools/buildVisualStories.js
skvale: I had to delete the visual snapshots
```
rm -r storybook-testing/__image_snapshots__
```
and recreate them for it to register these changes (because the base files didn't really change)

devtools/buildVisualStories.js
jbadan: Sure - do you want to push an update and test that code?

skvale: 9b8f81e

jbadan: Thank you for the help! 🙇

PR #1018, opened 2020-05-16
chore: clean up layout

bcullman:
### Description
* fill content area more completely
* use fixed header


jbadan:
This looks good! Only one small problem - the iframe gets hidden behind the left nav:
Screen Shot 2020-05-16 at 10 20 48 AM

.storybook/custom/custom.scss
```diff
@@ -194,17 +223,3 @@ body {
...
     padding: 10px 10px;
 }
-
-.break {
-    flex-basis: 100%;
-    height: 0;
-    border-bottom: 10px solid transparent;
-}
-
-.demo-icon-wrapper {
```
jbadan, line 204: These were needed! They help with layouts - that's why you had to update snaps. Check out icons: https://deploy-preview-1018--fundamental-react.netlify.app/?path=/docs/component-api-icon--primary

bcullman: ok, i see that `.demo-icon-wrapper` is needed. 

Is `.break` needed too?

bcullman:
@jbadan i went with prefix `fddocs-`.

jbadan:
@bcullman it doesn't look like you replaced all the uses of `docs-container` with `fddocs-container`

Issue #1014, opened 2020-05-13
InlineHelp: rebuild with Popover component


athulvraj:

JaySardhara:
@athulvraj is this duplicate of #920 ?

jbadan:
@JaySardhara This is the issue to track the work associated with updating InlineHelp - it will solve several open issues so it was easier for tracking to create this.

jbadan:
fixed https://github.com/SAP/fundamental-react/pull/1094

Issue #920, opened 2020-03-11
InlineHelp inside FormLabel is not rendering correctly


JaySardhara:
### Description
InlineHelp when used with FormLabel and its prop "isInlineHelp" does not correctly render inlineHelp message, where its box boundaries are cut-off.

Fundamental React version 0.8.0
![image](https://user-images.githubusercontent.com/6151138/76470576-bbd4d100-63ad-11ea-8b83-778bd2876625.png)

Fundamental React version 0.7.0
![image](https://user-images.githubusercontent.com/6151138/76470620-d3ac5500-63ad-11ea-8a21-a25eeddbf6e3.png)

### Versions
**fundamental-react:**
0.8.0
---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

jbadan:
fixed by #1094

PR #1024, opened 2020-05-20
Feat: #1014 Inline help rebuild with popover component

athulvraj:
### Description

fixes https://github.com/SAP/fundamental-react/issues/920
fixes https://github.com/SAP/fundamental-react/issues/1014

Corresponding styles work: https://github.com/SAP/fundamental-styles/pull/1041


src/InlineHelp/InlineHelp.js
```diff
@@ -20,22 +21,18 @@ const InlineHelp = React.forwardRef(({ text, placement, className, contentClassN
...
-            </span>
-        </span>
+        <Popover
+            body={
+                <div {...props}
+                    className='fd-inline-help__content fd-no-border' >
+                    {text}
+                </div>
+            }
+            control={<span><span
```
jbadan, line 32: Can we get rid of the extra span?

athulvraj: This was added since the ref was returning null when the ref was added to the parent element. Tried to add react fragment(<></>) but it was also failing.

skvale: It seems like Popover might be overwriting the ref prop [here](https://github.com/SAP/fundamental-react/blob/InlineHelp-rebuild-with-Popover-component/src/Popover/Popover.js#L136)

It could do something like this there
```jsx
            ref: (c) => {
                this.controlRef = findDOMNode(c);
                if (control.props.ref) {
                   control.props.ref(c);
                }
            }
```

jbadan:
Once https://github.com/SAP/fundamental-react/pull/1082 is merged, we should be up to the version of fundamental-styles right before the inline help changes went in. We'll be unblocked here 🎉

src/InlineHelp/InlineHelp.js
skvale: related conversation: https://github.com/SAP/fundamental-react/pull/1082#discussion_r434924795

prsdthkr:
This work seems to have been done and merged in https://github.com/SAP/fundamental-react/pull/1094
Should we close this? @athulvraj  @jbadan

PR #1054, opened 2020-05-28
feat: Select: add readOnly state

### LINKED ISSUE #1053: Select: Readonly state for Select is not working
### Description
Dear developer,
I followed the instructions as below:
https://sap.github.io/fundamental-styles/components/select.html
> To make the Select component read-only, the readonly attribute needs to be added to the fd-select__control element. This can also be done by adding the .is-readonly class or the aria-readonly="true" attribute.

My code is like below. However the Select that rendered is still same as the usual one.

```
const { toolTip, enabled = true, visible = true, readOnly = false } = props;
const readOnlyProperty = readOnly ? "true" : "false";
import "fundamental-styles/dist/select.css";
import { Select } from "fundamental-react";
  const dropDownListBoxClass = classnames(
    "fd-select--compact",
    readOnly ? "is-readonly" : ""
  );
<Select className={dropDownListBoxClass} title={toolTip} disabled={!enabled} aria-readonly={readOnly}></Select>
```

### Versions
**fundamental-react:**
0.9.0
---

_**DISCLAIMER:**

jbadan:
### Description

See https://sap.github.io/fundamental-styles/components/select.html

fixes #1053

### DIFF
--- a/src/Select/Select.js
+++ b/src/Select/Select.js
@@ -19,6 +19,7 @@ const Select = React.forwardRef(({
     onClick,
     onSelect,
     placeholder,
+    readOnly,
     selectedKey,
     validationState,
     ...props
@@ -50,6 +51,7 @@ const Select = React.forwardRef(({
         'fd-select__control',
         {
             'is-disabled': disabled,
+            'is-readonly': readOnly,
             [`is-${validationState?.state}`]: validationState?.state
         }
     );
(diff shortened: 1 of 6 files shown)


src/Select/Select.js
```diff
@@ -62,16 +64,16 @@ const Select = React.forwardRef(({
             {...props}
             className={selectClasses}
             id={id}
-            onClick={handleClick}
+            onClick={disabled || readOnly ? null : handleClick}
```
jbadan, line 67: @SAP/fundamental-react-team Should the dropdown still open if it's readonly? The spec doesn't specify and I know native `<select>` cannot be readOnly, so there's not a lot to go on here

src/Select/Select.js
```diff
@@ -62,16 +64,16 @@ const Select = React.forwardRef(({
...
             ref={ref}>
             <div className={selectControlClasses}>
                 {selected ? selected.text : placeholder}
```
jbadan, line 70: The design spec says the exclude the placeholder if it's readOnly, but fundamental-styles does not. We also don't have a min-width set (will make pr to fundamental-styles)

jbadan:
Will update with whatever the outcome of https://github.com/SAP/fundamental-styles/pull/1066 is as well

jbadan:
closing until 1066 is done

PR #1058, opened 2020-05-31
fix: SearchInput doesn't updated when searchList property change

### LINKED ISSUE #1012: SearchInput doesn't updated when searchList property change
Version : 
I am using fundamental-react version : 0.9 
Issue : 
SearchInput doesn't update when searchList change.

Example : 
I build the SearchInput with empty list till I get response from the server,
the control remain empty, since you read the list only in the constructor or when pressing ESC.

Current Behaviour : once SearchInput render it is not updated when searchList property change 
FIX : searchList should be read again in componentDidUpdate()

I065450:
### Description
read search list from props directly

fixes #1012


BenOron:
Hi @jbadan 

Can you please review ?

PR #1082, opened 2020-06-03
feat: upgrade fundamental-styles to 0.11.0-rc.4

### LINKED ISSUE #1071: Popover: styling is incorrect on docs site
### Description

Not sure if this is from the way the examples are constructed or if we missed something on the last few styles upgrades. 

Screen Shot 2020-06-02 at 6 45 19 AM

### Versions
**fundamental-react:** 0.10.2-rc.0

---

_**DISCLAIMER:**

jbadan:
### Description

BREAKING CHANGES:
- `Panel`: renamed `LayoutPanel`
- `Select`: underlying html changes
- `StepInput`: underlying html changes
- `Tile`: removed `active` prop
- `Tile`: `onClick` prop now required - all tiles are clickable by default
- `Tile`:  removed `productTile` prop
- `Tile`: removed `tabIndex` prop (no longer necessary because they are always buttons now)
- `Tile.Actions` removed
- `Tile.Media` removed
- `Tile.Content` no longer takes a `title` prop - this should now be added as a child to `Tile.Header`
- `Tile.Content`: removed `headingLevel` prop 
- `Tile.Content`: removed `titleProps` prop
- `Tile.Content`: removed `productTile` prop
- `Tile.Content`: underlying html changes related to removal of title 

Features:
- `List.Item`: add `selected` state
- `Select`: keyboard handling + focus handling
- `StepInput`: new `compact` state
- `Tile`: new `isDouble` prop (this is a bad name Help!) - maxes tile 2x1
- `Tile`: new `size` prop
- `Tile.Header`: new subcomponent
- `Tile.Footer`: new subcomponent
- `Tile.Content`: new `twoColumns` prop - enabled Content to be split into 2 columns

Fixes:
- fix Popover styling issues on docs site

fixes #1071


src/LayoutPanel/LayoutPanel.js
```diff
@@ -48,11 +48,11 @@ Panel.propTypes = {
...
-Panel.Filters = PanelFilters;
-Panel.Footer = PanelFooter;
-Panel.Head = PanelHead;
-Panel.Header = PanelHeader;
+LayoutPanel.Actions = LayoutPanelActions;
+LayoutPanel.Body = LayoutPanelBody;
+LayoutPanel.Filters = LayoutPanelFilters;
+LayoutPanel.Footer = LayoutPanelFooter;
+LayoutPanel.Head = LayoutPanelHead;
+LayoutPanel.Header = LayoutPanelHeader;
```
jbadan, line 56: I've exported something wrong in layout panel but I can't figure out what 👀

src/Select/Select.test.js
```diff
@@ -51,7 +51,7 @@ describe('<Select />', () => {
         });
     });
 
-    test('forwards the ref to the button', () => {
+    xtest('forwards the ref to the div role="button"', () => {
```
jbadan, line 54: @skvale I've done something wrong with how to handle forwarding the ref in Select - could you take a look when you get some time?

src/Select/Select.js
```diff
@@ -26,18 +28,122 @@ const Select = React.forwardRef(({
 
     useEffect(() => {
         if (!disableStyles) {
+            require('fundamental-styles/dist/button.css');
+            require('fundamental-styles/dist/icon.css');
             require('fundamental-styles/dist/select.css');
         }
     }, []);
 
+    const divRef = ref ? ref : useRef(null);
```
skvale, line 37: We shouldn't conditionally use hooks
```js
const internalDivRef = useRef(null);
const divRef = ref ? ref : internalDivRef;
```

src/Select/Select.test.js
skvale: Popover is overriding the ref prop of this control element
```js
        let controlProps = {
            onClick: onClickFunctions,
            ref: (c) => {
                this.controlRef = findDOMNode(c);
            }
        };
```
I'll see if there's way to extend the ref, but still maintain the original ref somehow

skvale: I think we need to do something like this in Popover
```js
let controlProps = {
    onClick: onClickFunctions,
    ref: (c) => {
        this.controlRef = findDOMNode(c);
        if (typeof control.ref === 'function') {
            control.ref(c);
        } else if (typeof control.ref === 'object') {
            control.ref.current = c;
        }
    }
};
```

src/StepInput/StepInput.js
```diff
@@ -25,11 +36,19 @@ const StepInput = React.forwardRef(({
...
+
+    const stepInputClasses = classnames(
         className,
-        'fd-input-group--control',
+        'fd-step-input',
         {
+            'fd-step-input--compact': compact,
             'is-disabled': disabled,
+            'is-readonly': readOnly,
             [`is-${validationState?.state}`]: validationState ?.state
```
skvale, line 52: I wonder why eslint doesn't pick up on this line having an extra space `validationState ?.state`

src/Select/Select.js
```diff
@@ -26,18 +28,124 @@ const Select = React.forwardRef(({
...
+            case 'tab':
+                e.stopPropagation();
+                setIsExpanded(false);
+                setFocusedElement(tryFocus(divRef.current));
+                break;
+            case 'enter':
+            case 'space':
+                e.stopPropagation();
+                handleSelect(e, option);
+                setFocusedElement(tryFocus(divRef.current));
```
skvale, line 135: It looks like the intent of the `enter` and `space` is to close the select.  This isn't happening.
For some reason it is calling the `handleClick` immediately after the `handleSelect` and the `handleClick` is toggling the `isExpanded` state back to true

skvale: It works as expected if you call `e.preventDefualt();` though

src/StepInput/StepInput.js
jbadan: I'm not sure - good catch though!

src/Button/__stories__/__snapshots__/ButtonSegmented.visual.storyshot
```diff
@@ -1,60 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Storyshots Visual Button Group 1`] = `
```
jbadan, line 3: this was a duplicate of the same storyshot below

PR #1010, opened 2020-05-13
chore: cleanup warnings on docs site, small bug fixes and build out forms documentation

jbadan:
### Description
This pr does small bug fixes that were causing warnings on the docs site. 

fixes #993 
fixes #974

PR #1083, opened 2020-06-03
feat: weekdayStart for Calendar

justeban:
### Description

This is to add support for the Calendar component to be able to start on any day of the week. The new prop I'm suggesting would be `weekdayStart`. This prop would take in an integer with acceptable values of 0-6 
```
weekdayStart = 0   // Sunday
weekdayStart = 1   // Monday
weekdayStart = 2   // Tuesday
weekdayStart = 3   // Wednesday
weekdayStart = 4   // Thursday
weekdayStart = 5   // Friday
weekdayStart = 6   // Saturday
```


jbadan:
@justeban Thanks for your contribution!

@prsdthkr @jacobdevera @meganaconley You're more familiar with the DatePicker, would this be an acceptable change? While I like it and think it's a reasonable feature, I'm not seeing anything like it in the fiori specs or ui5 - https://sapui5.hana.ondemand.com/#/entity/sap.m.DatePicker/sample/sap.m.sample.DatePicker
https://experience.sap.com/fiori-design-web/date-picker/

src/Calendar/__stories__/Calendar.stories.js
```diff
@@ -85,6 +88,11 @@ export const dev = () => (
         locale={text('locale', 'en')} />
 );
 
+export const weekDayStart = () => {
+    const _weekDayStart = number('weekDayStart', 1);
+    return <Calendar weekDayStart={_weekDayStart} />;
```
jacobdevera, line 93: Could this adjustable knob instead be part of the `dev` story above, and have a set value for this story for demo purposes? On the docs site, this just appears as a regular calendar.

jbadan:
@jacobdevera Sounds good to me!

src/Calendar/Calendar.js
```diff
@@ -673,19 +696,19 @@ Calendar.displayName = 'Calendar';
 
 Calendar.propTypes = {
     /** Blocks dates that are in between the blocked dates */
-    blockedDates: PropTypes.arrayOf(PropTypes.instanceOf(Date)),
+    blockedDates: PropTypes.arrayOf(PropTypes.instanceOf(moment)),
```
justeban, line 699: The change here, and subsequent areas below, the `Date` was giving me an error in the console while writing tests and was expecting instances of a Moment. This seemed to mitigate the error in the console. Not sure if that will cause a problem 🤷

src/Calendar/__stories__/Calendar.stories.js
```diff
@@ -70,6 +73,11 @@ export const rangeSelection = () => (
     <Calendar enableRangeSelection />
 );
 
+export const weekdayStart = () => {
```
justeban, line 76: So I kept a story to showcase a different starting week in the docs but added the prop to the dev story as well.

src/Calendar/Calendar.test.js
```diff
@@ -412,6 +414,87 @@ describe('<Calendar />', () => {
         expect(wrapper.state('arrSelectedDates').length).toEqual(2);
     });
 
+    describe('weekdayStart', () => {
```
justeban, line 417: Hopefully these tests make sense but I'm trying to verify that we are starting on the right weekday and that the right corresponds to the proper weekday.

src/DatePicker/DatePicker.js
```diff
@@ -477,6 +479,8 @@ DatePicker.propTypes = {
         /** Text of the validation message */
         text: PropTypes.string
     }),
+    /** Number to indicate which day the week should start. 1 = Sunday, 2 = Monday, 3 = Tuesday, 4 = Wednesday, 5 = Thursday, 6 = Friday, 7 = Saturday */
+    weekdayStart: PropTypes.number,
```
justeban, line 483: Getting the prop up through the DatePicker

jacobdevera: @jbadan it seems Calendar.basePropTypes was removed in https://github.com/SAP/fundamental-react/pull/1010/files#diff-35d7f79fc5e005a5d077c11beac4ece9L674 . Is it better to duplicate shared Calendar propTypes over to the DatePicker, or should it even be defined in DatePicker's propTypes in the first place?

src/Calendar/Calendar.js
jacobdevera: Yeah, fair call. I do remember mixing up `Date` and `moment` props at some point.

src/DatePicker/DatePicker.js
justeban: kind of related: Would we want a PR in https://github.concur.com/coreui/nui-widgets-lab/blob/master/src/DatePicker/DatePicker.jsx as well to add the `weekdayStart` to the PropTypes. It looks like it would work as is because we spread `...otherProps` but not sure if we would want to be explicit there. I don't know if that decision would influence the decision here.

justeban:
The travis job is failing with the `npm run test:storybook` command. I'm just not familiar enough to tell if the expected and received changes are reasonable to expect. Some of the failures seem to be unrelated to the changes I made - https://travis-ci.org/github/SAP/fundamental-react/jobs/694827676

jbadan:
@justeban You'll need to run `test:storybook:update` because you added a story. We have it automated to take snapshots and visual images of each story. The others were affected because we have a bug with short id :)

src/DatePicker/DatePicker.js
justeban: @jbadan ?

src/Button/__stories__/__snapshots__/ButtonGroup.visual.storyshot
```diff
@@ -0,0 +1,278 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
jbadan, line 1: Can you remove this file? It shouldn't be in here - ButtonGroup was deleted.

src/DatePicker/DatePicker.js
jbadan: Sorry, I missed this notification. Can we use `CustomPropTypes.range` here? https://github.com/SAP/fundamental-react/blob/master/src/utils/CustomPropTypes/CustomPropTypes.js#L42

`Calendar.basePropTypes` was removed because it did not play nicely with the automated prop table creation. https://github.com/SAP/fundamental-react/blob/master/src/DatePicker/DatePicker.js#L439 is a typo and should be `...Calendar.propTypes`. Then we won't need to duplicate everything and the sharing should work again. Good catch!

jbadan:
One last request! Can we default 0 to Sunday - it looks like that's what `Date.getDay()` does as well as moment: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDay
https://momentjscom.readthedocs.io/en/latest/moment/02-get-set/06-day/

justeban:
> One last request! Can we default 0 to Sunday - it looks like that's what `Date.getDay()` does as well as moment: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDay

I think that makes sense my reasoning for starting with 1 is because nui-intl starts its numbering on Sunday with `1`, but I think it would be better to start at 0.

justeban:
These merge conflicts seem to be just different `aria-control` and `id` numbers. Am I okay just to keep my current or changes or do I need to accept incoming changes and run the storybook update again? 

Screen Shot 2020-06-08 at 8 10 04 AM
Screen Shot 2020-06-08 at 8 09 45 AM

jbadan:
@justeban I would merge and update - I have a pr up to fix that problem in the future

PR #1086, opened 2020-06-05
docs: add withKnobs decorator to all stories + add toolbar to all docs examples + fix shortid 

jbadan:
### Description
withKnobs was missing from global decorators 

reworked DocsPage to have toolbars for every example

changed shortid mock so that every snapshot does not need to be updated when one is changed

fixes #issueid


jbadan:
Opening a new pr that has these changes + storybook@6!

PR #1089, opened 2020-06-09
feat: Substring Search in SearchInput

### LINKED ISSUE #1003: Substring search doesn't work in SearchInput
**Version :** I am using fundamental-react version : **0.9**
**Issue :** Substring search doesn't work in SearchInput.
Example : Below values are present in my searchList
1) Who is a supplier user?
2) How to set the PIN?
3) Who is a buyer user?
4) How to change the PIN if I forget it?
Key to be searched : 
a)Who : return option 1,3 [CORRECT]
b)How : return option 2,4 [CORRECT]
c) PIN : empty result  **Issue** [Should return 2,4]
d) user : empty result  **Issue** [Should return 1,3]

Current Behaviour : Search key is compared only with the starting characters in the search list.
FIX :  Search key should be compared with the substring in the search list.

muthukkaruppan4695:
### Description
https://github.com/SAP/fundamental-react/issues/1003

fixes #1003


src/SearchInput/SearchInput.js
```diff
@@ -257,7 +269,8 @@ SearchInput.propTypes = {
 SearchInput.defaultProps = {
     onChange: () => { },
     onEnter: () => { },
-    onSelect: () => { }
+    onSelect: () => { },
+    subStringSearch: false
```
jbadan, line 273: Boolean props default to false so we can exclude this

jbadan: other than this small change and updating with master this looks good to me! Thanks for contributing!

src/SearchInput/SearchInput.js
```diff
@@ -207,6 +219,8 @@ SearchInput.propTypes = {
     className: PropTypes.string,
     /** Set to **true** to enable compact mode */
     compact: PropTypes.bool,
+    /** Internal use only */
+    disableStyles: PropTypes.bool,
```
prsdthkr, line 223: `disableStyles` is not needed anymore: https://github.com/SAP/fundamental-react/pull/1102

muthukkaruppan4695: Yes I have removed it.

src/SearchInput/SearchInput.js
```diff
@@ -116,6 +119,7 @@ class SearchInput extends PureComponent {
             searchBtnProps,
             popoverProps,
             validationState,
+            disableStyles,
```
jbadan, line 122: It needs to be removed here as well then CI should pass :)

muthukkaruppan4695: Yes, removed it.

jbadan:
I had to revert this as it was causing lots of failures in ci - I'm wondering if it needed to be updated with master?

muthukkaruppan4695:
@jbadan @prsdthkr  Why it is causing failure in CI? What should be done to fix this issue ?

jbadan:
@muthukkaruppan4695 Can you reopen the pull request? I can't do it in on your behalf

src/SearchInput/SearchInput.js
```diff
@@ -231,6 +241,8 @@ SearchInput.propTypes = {
             callback: PropTypes.func
         })
     ),
+    /** enable substring search */
+    subStringSearch: PropTypes.bool,
```
skvale, line 245: What do you think about keeping the capitalization consistent with JavaScript's `substring` method?
`subStringSearch` -> `substringSearch`

PR #1091, opened 2020-06-11
feat: 1020: Shellbar pass popoverProps to SearchInput

### LINKED ISSUE #1020: Shellbar does not provide a way to pass Popover props to Search Input
### Description
It looks like there's no way to pass props through the Shellbar's search box to the popover used to display the list.

I ran into this on Chrome 83.0.4103.56 on Mac 10.15.4 with the window set to 1200x600. I copied the example "Links with collapsible menu, CoPilot and Product Switch" Shellbar from the docs (https://sap.github.io/fundamental-react/shellbar) to a new React app and had the App function return only the Shellbar:

```javascript
function App() {
   return <Shellbar ... />
}
```

Running the app, then clicking in the search input box causes the search list to be displayed above the search input box. Making the window bigger causes the list to be displayed below the search input box, as expected.

It looks like this could be fixed by passing {disableEdgeDetection:true, placement:"bottom"} to the Popover, but the Shellbar's SearchInput interface doesn't pass those props through.

### Versions
**fundamental-react: 0.9.0**

---

_**DISCLAIMER:**

prsdthkr:
### Description
With this change
+ Shellbar passes popoverProps to SearchInput
  + SearchInput popoverProps default to `placement='bottom'` and `disableEdgeDetection=true`
+ Product Switch button gets aria-label from `productSwitch.label`
   + `productSwitch.label` prop is now required
   + `productSwitch.label` prop documented
+ Popover `widthSizingType ` prop documentation markdown is fixed. Earlier, using MD code 
image

fixes #1020

### DIFF
--- a/src/Shellbar/Shellbar.js
+++ b/src/Shellbar/Shellbar.js
@@ -61,7 +61,9 @@ class Shellbar extends Component {
         if (this.props.productSwitch) {
             let collapsedProductSwitch = this.props.productSwitch;
 
+            // eslint-disable-next-line react/prop-types
             collapsedProductSwitch.glyph = 'grid';
+            // eslint-disable-next-line react/prop-types
             collapsedProductSwitch.callback = () => {
                 this.setState(prevState => ({
                     showCollapsedProductSwitchMenu: !prevState.showCollapsedProductSwitchMenu
@@ -206,6 +208,7 @@ class Shellbar extends Component {
                                 inputProps={{ className: 'fd-shellbar__input-group__input' }}
                                 onEnter={searchInput.onSearch}
                                 placeholder={searchInput.placeholder}
+                                popoverProps={searchInput.popoverProps}
                                 searchBtnProps={{ className: 'fd-shellbar__button' }}
                                 searchList={searchInput.searchList} />
                         </div>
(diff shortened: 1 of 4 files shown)


src/Shellbar/__stories__/Shellbar.stories.js
```diff
@@ -198,7 +198,12 @@ export const coPilot = () => (
         profileMenu={profileMenu}
         searchInput={{
             label: 'Search',
-            placeholder: 'Search'
+            placeholder: 'Search',
+            popoverProps: {
+                //to always show search-results popup towards bottom
+                disableEdgeDetection: true,
+                placement: 'bottom'
+            }
```
jbadan, line 206: Should we add these as defaults (in shellbar only - not searchinput in general) for consumers that they can then override if they want to?

src/Shellbar/Shellbar.js
```diff
@@ -499,8 +504,11 @@ Shellbar.propTypes = {
     notifications: PropTypes.object,
     /** Holds product titles and navigation */
     productMenu: PropTypes.array,
-    /** For navigating between products */
-    productSwitch: PropTypes.object,
+    /** For navigating between products. An object that contains an accessible label for product switch button. */
+    productSwitch: PropTypes.shape({
+        /** Accessible label for product switch button */
+        label: PropTypes.string.isRequired
```
jbadan, line 510: 👏  Should we mark this as a breaking change when merging since this is now required?

src/Shellbar/__stories__/Shellbar.stories.js
prsdthkr: Makes sense

src/Shellbar/Shellbar.js
jbadan: Oh I'm sorry - it was always marked as required - I read the diff wrong 😞

PR #1096, opened 2020-06-17
feat: openToDate prop for Calendar

jacobdevera:
### Description
Adds `openToDate` prop to specify a date that will be focused when the Calendar is mounted if consumers desire to override the default, which is to focus today's date.

Official fiori spec: https://experience.sap.com/fiori-design-web/date-picker/#components

> You can change the default focus __ “today” to another date. This can save users several clicks when they create events. For an event end date, for example, the focus should propose a date in the future (after the start date).


src/DatePicker/DatePicker.js
```diff
@@ -333,6 +333,7 @@ class DatePicker extends Component {
             localizedText,
             onBlur,
             onDatePickerClose,
+            openToDate,
             popoverProps,
```
prsdthkr, line 337: Should we add `openToDate` to `DatePicker.propTypes` too?

jacobdevera: I believe that is covered by `...Calendar.PropTypes`

prsdthkr: Right.

Wouldn't the de-structuring then be
```js
const { calendarProps: openToDate } = this.props
```
and story be
```js
<DatePicker calendarProps={{openToDate: _openToDate}} />;
```
?

jacobdevera: I think there is just a discrepancy for how we handle some Calendar props. Some of them are raw props to spread to the Calendar's HTML elements:
`calendarProps: PropTypes.shape({
        monthListProps: PropTypes.object,
        tableBodyProps: PropTypes.object,
        tableHeaderProps: PropTypes.object,
        tableProps: PropTypes.object,
        yearListProps: PropTypes.object
    })`

While the others that specify logic for dates and such (`blockedDates`, `disabledDates`, etc.) are pulled up a level to make for a simpler API and not having a huge `calendarProps` object.

Seems to be out of scope for this PR, though.

PR #1097, opened 2020-06-18
fix: style sidebar

bcullman:
### Description
possible way to make the sidebar header stick to the top

### DIFF
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -2,7 +2,7 @@
     .sidebar-header {
         position: fixed !important;
         width: 210px;
-        background-color: white;
+        background-color: #fff;
         border-bottom: 2px solid #d7dbde;
         z-index: 10;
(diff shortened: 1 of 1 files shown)


.storybook/manager-head.html
```diff
@@ -0,0 +1,56 @@
...
+        position: relative;
+    }
+
+    .sidebar-header > div:first-child {
+        position: relative;
+        margin-right: 0;
+    }
+
+    .sidebar-header > div:last-child {
+        display: none;
```
jbadan, line 23: Per our discussion on slack, I think we need to keep this

bcullman: 😢

bcullman: ok done.

jbadan: I know I know it looks so much better with it hidden 😿

PR #1100, opened 2020-06-19
feat: Upgrade to popper-2

### LINKED ISSUE #1042: Upgrade to react-popper 2.x.x
https://github.com/popperjs/react-popper

skvale:
### Description

Use popper 2

fixes #1042


package.json
```diff
@@ -55,7 +56,8 @@
     "react-foco": "^1.3.0",
     "react-focus-lock": "^2.3.1",
     "react-overlays": "^2.1.0",
-    "react-popper": "^1.3.3",
+    "react-popper": "^2.2.3",
+    "react-popper-2": "npm:react-popper@^2.2.3",
```
skvale, line 60: This is temporary to fix an issue with Storybook
Storybook: https://github.com/storybookjs/storybook/issues/10982

Storybook must be doing something with the how deps are resolved, but it should work after they update react-popper-tooltip, tracking that here: https://github.com/mohsinulhaq/react-popper-tooltip/issues/92

src/utils/_Popper.js
```diff
@@ -171,9 +236,8 @@ Popper.propTypes = {
     referenceComponent: PropTypes.element.isRequired,
     disableEdgeDetection: PropTypes.bool,
     noArrow: PropTypes.bool,
-    placementTargetRef: PropTypes.shape({ current: PropTypes.any }),
     popperClassName: PropTypes.string,
-    popperModifiers: PropTypes.object,
+    popperModifiers: PropTypes.array,
```
skvale, line 240: I don't think this is a breaking change because it's not an exported component and it isn't accessible via exported components.

jbadan:
Do we need some global polyfills in storybook so our docs will work in ie11 (object.assign/array.find/maybe promise)?

jbadan:
Can we add a note in the readme about global polyfills needed? 
```
## Polyfills

Fundamental React requires a polyfill for [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign), [Array.find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) and [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise).

You will need to add your own polyfills. See [core-js](https://github.com/zloirock/core-js#commonjs-api) for instructions.
```

skvale:
blocked by https://github.com/popperjs/popper-core/issues/1127

jbadan:
@skvale Will this issue block v13 release?

skvale:
I think I can fix the issue from ui-assets styles and this can go in after that

PR #1105, opened 2020-06-24
fix: use popper for form validation messages

### LINKED ISSUE #1095: FormMessage should not consume DOM space
### Description
FormMessage in InputGroup takes space in the DOM in Fundamental React. This can cause issues especially in popover-based components (e.g. DatePicker) where DOM space is lost and regained depending on if the popper is open.

In Fundamental Styles, a popper with absolute positioning is used: https://fundamental-styles.netlify.app/components/form.html#input-states

The InputGroup itself should also use a popper.

### Versions
**fundamental-react: 0.11.0-rc.11**

jacobdevera:
### Description

- use `FormValidationOverlay` to display validation messages. Remove direct renders of `FormMessage`, which consumes DOM space
- do not display validation message until field is focused (per https://experience.sap.com/fiori-design-web/form-field-validation/)
- refactoring keyboard navigation of Select to rely on focusManager in Popover instead
- add `firstFocusIndex` prop to Popover to indicate a focusable item to focus first
#### Affected components with validation
- ComboBoxInput
- DatePicker
- FormInput
- FormTextarea
- InputGroup
- MultiSelect
- Select
- StepInput

#### Out of Scope
If a component did not already have keyboard navigation, it was not added

fixes #1095


jbadan:
This looks really good! Two things:

1) comboboxinput, multiInput, localizationeditor, searchInput, select - when I click into the input - the dropdown opens, when I tab into it only the validation opens. I'm totally fine with this being a followup if you need to get this DatePicker fix to production asap

2) something went wonky with inputgroup styles

Screen Shot 2020-06-25 at 6 37 59 AM
Screen Shot 2020-06-25 at 6 37 22 AM

jacobdevera:
> 1. comboboxinput, multiInput, localizationeditor, searchInput, select - when I click into the input - the dropdown opens, when I tab into it only the validation opens.

I believe this is consistent with previous behavior; they should only open on click. However, form validation should appear on focus as well. But a deeper dive could be done in a separate PR.

> 2. something went wonky with inputgroup styles

✅

Issue #1035, opened 2020-05-21
ObjectStatus: new component


jbadan:
see: https://sap.github.io/fundamental-styles/components/object-status.html

PR #1118, opened 2020-07-02
feat: ObjectStatus new component

athulvraj:
### Description
Added a new component - ObjectStatus  which displays the semantic status of an object
https://sap.github.io/fundamental-styles/components/object-status.html

Issue created:  #1035


src/ObjectStatus/ObjectStatus.js
```diff
@@ -0,0 +1,71 @@
...
+            {children}
+        </StatusTag>
+    );
+});
+
+
+ObjectStatus.displayName = 'ObjectStatus';
+
+ObjectStatus.propTypes = {
+
```
jbadan, line 43: Can we add something like:

```
ariaLabel: (props) => {
        if (
            (!props.children && !props.ariaLabel) ||
            (!props.children && typeof props.ariaLabel !== 'string')
        ) {
            return new Error('ariaLabel is required if there is no text inside the ObjectStatus component and must be a string');
        }
    },
```

So that we have a required aria-label for a11y for icon only status labels?

src/ObjectStatus/ObjectStatus.js
```diff
@@ -0,0 +1,71 @@
...
+        className
+    );
+    const StatusTag = typeof link !== 'undefined' ? 'a' : 'span';
+
+    return (
+        <StatusTag
+            {...props}
+            className={objectStatusClasses}
+            href={link}
+            onClick={onClick}
```
jbadan, line 32: I think if it has `onClick` it will also need a `role='button'`, but feel free to correct me @prsdthkr / @meganaconley  / @jacobdevera

meganaconley: If it's a `span` it should have `role='button'` and also support using `space` and `enter` to activate whatever the `onClick` is doing. If it's an `a` then only `enter` needs to be supported and it should not have `role='button'`.

prsdthkr: We will also need the buttons and links to have tab index.
My suggestion:
```js
    let StatusTag = 'span';
    let semanticProps = {};

    if (link?.length) {
        StatusTag = 'a';
        semanticProps.href = link;
    } else if ( typeof onClick === 'function') {
        StatusTag = 'button';
        semanticProps.onClick = onClick;
        semanticProps.style = {
            background: 'transparent'
        };
    }
    return (
        <StatusTag
            {...props}
            className={objectStatusClasses}
            {...semanticProps}
            ref={ref}>
            {children}
        </StatusTag>
    );
```

src/ObjectStatus/ObjectStatus.js
athulvraj: Yes will do it

src/ObjectStatus/__stories__/ObjectStatus.stories.js
```diff
@@ -0,0 +1,186 @@
+/* eslint-disable react/no-multi-comp */
+import ObjectStatus from '../ObjectStatus';
+import React from 'react';
+
+export default {
+    title: 'Component API/ObjectStatus',
```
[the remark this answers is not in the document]
athulvraj: Sure

src/ObjectStatus/ObjectStatus.js
```diff
@@ -0,0 +1,89 @@
...
+
+    if (link?.length) {
+        StatusTag = 'a';
+        semanticProps.href = link;
+    } else if ( typeof onClick === 'function') {
+        StatusTag = 'button';
+        semanticProps.onClick = onClick;
+        semanticProps.style = {
+            background: 'transparent'
+        };
```
prsdthkr, line 36: I realized that this won't play well with `inverted` status. Either we should get the styles for `<button>` into fundamental-styles or use `<span>` with `role='button'`, `onClick` handler, `onKeyPress` handler, and `tabIndex=0`.

PR #1119, opened 2020-07-06
feat: Publish the package with a lib directory

skvale:
### Description

This helps consumers `npm link` the repo locally.

BREAKING CHANGE: All imports from subdirectories will need to include the `lib` in their path now.

### DIFF
--- a/.storybook/custom/components/Import.js
+++ b/.storybook/custom/components/Import.js
@@ -11,7 +11,7 @@ const Import = () => {
         customStyle={{ padding: 10, whiteSpace: 'pre-wrap', fontSize: 14 }}
         language='javascript'
         style={googlecode}>
-        {`import { ${groups[groups.length - 1]} } from 'fundamental-react/${groups[1]}';`}
+        {`import { ${groups[groups.length - 1]} } from 'fundamental-react/lib/${groups[1]}';`}
     </SyntaxHighlighter>)
 }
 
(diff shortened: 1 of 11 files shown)


.storybook/custom/components/Import.js
```diff
@@ -11,7 +11,7 @@ const Import = () => {
         customStyle={{ padding: 10, whiteSpace: 'pre-wrap', fontSize: 14 }}
         language='javascript'
         style={googlecode}>
-        {`import { ${groups[groups.length - 1]} } from 'fundamental-react/${groups[1]}';`}
```
jbadan, line 14: So this isn't your fault, but I just noticed:

`import { ButtonSegmented } from 'fundamental-react/lib/ButtonSegmented';` is not correct, it should be `import { ButtonSegmented } from 'fundamental-react/lib/Button';`

I don't think there's any other components that share a folder

This works for forms because they're inside of a folder in storybook. So we can either make this import component smarter or fix Button to be under a folder by changing https://github.com/SAP/fundamental-react/blob/master/src/Button/__stories__/ButtonSegmented.stories.js#L7 to `title: 'Component API/Button/ButtonSegmented'` and the same for Button.

skvale:
```
import { TabGroup } from 'fundamental-react/lib/TabGroup';
```

should be
```
import { TabGroup } from 'fundamental-react/lib/Tabs';
```

skvale:
Just to confirm, I should include `BREAKING CHANGE: ` in the commit message for this PR right?

jbadan:
@skvale Yeah - the only repo where you can't yet is fundamental-styles (confusing I know)

PR #1121, opened 2020-07-09
feat: upgrade fundamental-styles to 0.11.0-rc.100

jacobdevera:
#### Affected Components
- Shellbar: classes updated (buttons should have focus outline); remove ‘fd-product-menu’ class
- Select: role changed to `combobox`, `aria-haspopup` changed to `listbox`, fixing `aria-disabled` and `aria-readonly`
- Table row selection styling
- Checkbox: `fd-checkbox__text` class added for ellipsis overflow
- Dialog: remove `backdropClassName`, remove overlay classes
- DatePicker: `fd-bar—cosy` renamed to `fd-bar—cozy`; new icon for ‘choose date’ button
- Icon: sizes removed, using font-size
- Popover: `is-expanded` class added


src/utils/constants.js
```diff
@@ -38,12 +38,12 @@ export const FORM_MESSAGE_TYPES = [
...
-    'm',
-    'l',
-    'xl'
-];
+export const ICON_SIZES = {
+    's': '12px',
+    'm': '16px',
+    'l': '20px',
+    'xl': '24px'
+};
```
jacobdevera, line 46: Debating removing these like what was done for styles, but for some reason it feels wrong to not have predefined sizes and having the consumer set it.

src/Table/Table.js
```diff
@@ -74,14 +81,21 @@ const Table = React.forwardRef(({ headers, tableData, className, tableBodyClassN
...
+                            </td>
+                        );
                         displayRows = row.rowData.splice(1, row.rowData.length);
                     }
 
                     return (
                         <tr
                             className={tableRowClasses}
                             {...rowProps}
+                            aria-selected={selectedRows[index]}
```
skvale, line 98: If we know the first item in the row is a checkbox, can we just do
```js
aria-selected={row?.rowData?[0]?.props.checked}
```
Otherwise we'd be having duplicated selected logic here and the consumer, like Table.stories.js, and they could get out of sync.  Such as if they wanted to render a row pre-selected this component would not account for that.

src/Forms/Checkbox.js
```diff
@@ -77,7 +83,7 @@ const Checkbox = React.forwardRef(({
                 className={labelClasses}
                 disabled={disabled}
                 htmlFor={checkId}>
-                {children}
+                {checkboxChildren}
```
skvale, line 86: on a related note, is children a required prop, I saw this warning in our Table stories
```
Warning: Failed prop type: The prop `children` is marked as required in `Checkbox`, but its value is `undefined`.
    in Checkbox (created by storyFn)
    in storyFn
    in ErrorBoundary
```

src/Dialog/Dialog.js
```diff
@@ -161,8 +152,6 @@ Dialog.propTypes = {
     actions: PropTypes.arrayOf(PropTypes.node).isRequired,
     /** Localized text for the heading */
     title: PropTypes.string.isRequired,
-    /** CSS class(es) to add to the dialog backdrop */
-    backdropClassName: PropTypes.string,
```
jbadan, line 165: I think we should still keep the ability to put a class on the backdrop even though `fd-overlay` was removed

src/Forms/Checkbox.js
jbadan: I think it was me that set it to required - from my memory it was to enforce the checkbox having a label

src/Icon/Icon.js
```diff
@@ -34,7 +34,7 @@ Icon.propTypes = {
     /** CSS class(es) to add to the element */
     className: PropTypes.string,
     /** Size of the component: 's', 'm', 'l', 'xl' */
-    size: PropTypes.oneOf(ICON_SIZES)
+    size: PropTypes.oneOf(Object.keys(ICON_SIZES))
```
jbadan, line 37: let's choose a default size

src/utils/constants.js
jbadan: I think icon sizes should not have been removed from fundamental-styles - I think this is fine until they add them back

src/SearchInput/__stories__/__snapshots__/SearchInput.stories.storyshot
```diff
@@ -158,6 +157,7 @@ exports[`Storyshots Component API/SearchInput Disabled 1`] = `
               >
                 <div
                   className="fd-input-group--control fd-input-group"
+                  onClick={[Function]}
```
jbadan, line 160: Is it right for the onClick to be on a different element than the focus and blur?

src/Icon/Icon.js
jbadan: Actually let's not so it just gets the whatever font-size - can we make that prop conditional so we don't have undefined when it's not specified though

src/Forms/Checkbox.js
meganaconley: Unless they change the design for the table with checkbox, we need to mark `children` as not required and provide an aria-label alternative. Or do a custom prop so we require either `children` or an `aria-label`.

src/utils/constants.js
droshev: the values should be in `rem` at least:
```
0.75rem
1rem
1.25rem
1.5rem
```

src/Icon/Icon.js
jbadan: Look at the diff for InputGroup:
```
<span
          className="sap-icon--globe"
          style={
            Object {
              "fontSize": undefined,
            }
          }
        />
```
It should probably not even add the style prop if fontSize is undefined instead of adding that

PR #1122, opened 2020-07-10
feat: Allow a custom flip boundary container for Popover

skvale:
### Description

We're making it possible to define a boundary where the popover should flip within.  This example the popover would flip to the left first, but since the left is outside of the boundary it considers that direction invalid.

popover

------

## You might notice this happening in the DatePicker story

datepicker-shift

It's shifting to the `bottom-start` placement because for a time there is no fallback placement that works.  Since the bottom is off the page and there's not enough room to the top yet.  I don't see a good way around this other than make the DatePicker smaller.


babel-build.config.js
```diff
@@ -0,0 +1,16 @@
+const rootBabel = require('./babel.config');
+
+// TODO: Remove this once https://github.com/storybookjs/storybook/issues/11246 is fixed
+const popper2AliasRemovalPlugin = [
+    'babel-plugin-module-resolver',
+    {
+        alias: {
+            'react-popper-2': 'react-popper'
```
skvale, line 8: I'm jumping through hoops to get the generated output to look like
```js
var _reactPopper = require("react-popper");
```
instead of
```js
var _reactPopper = require("react-popper-2");
```
and still have the examples work in storybook

PR #1124, opened 2020-07-15
fix: prevent disabled and readonly select from opening

jacobdevera:
### Description

- disable Popover triggers on disabled and readonly Select
- move onClick to right component


jacobdevera:
The Fiori spec states to make them non-focusable: https://experience.sap.com/fiori-design-web/ui-element-states/#disabled

PR #1125, opened 2020-07-16
feat: combobox filtering and autocomplete

### LINKED ISSUE #1019: ComboboxInput: should filter based on FormInput input
### Description

ComboBox input should filter the dropdown + enable selection via the FormInput. This is necessary for the replacement of SearchInput (no longer supported in fundamental styles)

### Versions
**fundamental-react:**

---

_**DISCLAIMER:**

prsdthkr:
### Description
In this change, we make the Combobox filterable and enable autocomplete capabilities.
We have tried to align with the specification here: [ARIA 1.1 Combobox with Listbox Popup Examples
](https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html) There is are some minor deviations from the specification because of the way we filter the Combobox options.

QA:
https://deploy-preview-1125--fundamental-react.netlify.app/?path=/story/component-api-comboboxinput--dev

fixes #1019


skvale:
is this still a WIP or ready to be reviewed?

prsdthkr:
> is this still a WIP or ready to be reviewed?

Hi @skvale, feel free to review and [QA](https://deploy-preview-1125--fundamental-react.netlify.app/?path=/story/component-api-comboboxinput--dev). I'm in the process of writing unit tests for the interactions.

devtools/buildVisualStories.js
```diff
@@ -39,7 +39,9 @@ export default {
 };
 
 export const ${componentName} = () => {
-    let storyNames = Object.keys(stories).filter(story => story !== 'default');
+    let storyNames = Object.keys(stories).filter(story => {
+        return story !== 'default' && story !== 'dev';
```
jbadan, line 43: Good idea! Just in case other reviewers are wondering why there's so many visual test changes - this is why. I think it's a good idea to remove the dev stories from the visuals.

jbadan:
I see this a11y violation, but I don't really understand it. I looked up the values for aria-controls, https://www.w3.org/WAI/PF/aria-1.1/states_and_properties#aria-controls and it says "ID reference list: A list of one or more ID references." We use aria-controls like this everywhere so 🤷‍♀️  Maybe this is a case of the tool not understanding grey area.

Screen Shot 2020-07-24 at 7 07 34 AM

jbadan:
This is an issue for fundamental-styles, but I think the list should either be full width or the container should be limited. We should open an issue over there.

Screen Shot 2020-07-24 at 7 11 45 AM

jbadan:
This looks really good!! 

1. Should the Combobox open if I haven't typed anything yet and press the down key - like if I wanted to see all the options or is that not a thing?

2. We should pass the `compact` size to `List`:

Screen Shot 2020-07-24 at 7 16 00 AM

3. We need to pass `autocomplete='nope'` instead of `autocomplete='off'` to `FormInput`, let the constant battle with keeping it disabled begin 😆 

Screen Shot 2020-07-24 at 7 29 36 AM

prsdthkr:
Thanks for the review, Jenna 😃 

* aria-controls a11y issue: this happens mostly because the Listbox that the id is referring to is non-existent until opened. So if you re-run the a11y tests on opening the listbox - the error should disappear. To re-run the tests click on the `✓ Tests completed` thingy below. I don't know if there is a way to overcome this violation.

* autocomplete=nope introduces a new a11y violation 🙄  `Ensure the autocomplete attribute is correct and suitable for the form field`. Is this acceptable?

* > Should the Combobox open if I haven't typed anything yet and press the down key - like if I wanted to see all the options or is that not a thing?

  For this, I'll try to align with [the ARIA example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html). There, the Combobox opens like that only in the 3rd example.

* List width styling issue seems to be with the `fd-list--dropdown` class. Removing it has fixed it for now. I'll create an issue in fd-styles for this.

* > We should pass the compact size to List

  good catch  👍

jbadan:
@prsdthkr I think you're right about the autocomplete issue. Maybe we put a smaller version of this https://hig.concur.com/pages/develop/web/components/support.html#autocomplete-attribute-issue, to explain why that might happen. 
Then we can do something like this:
```
<FormInput
    autocomplete='off'
    {...inputProps}
   {...labelProps}
   ....
/>
```

The consumer then has full control to override our autocomplete attribute to whatever works at the time.

PR #1134, opened 2020-07-24
Fix: focusable disabled button

### LINKED ISSUE #889: Use aria-disabled instead of disabled attribute on Button
### Description
For accessibility purposes, it will be better to use the aria-disabled attribute than the built-in `disabled` attribute to differentiate disabled Buttons.

- Use aria-disabled instead of disabled
- Handle click/key events when disabled
- Ensure styles are same
- Add aria-live region to announce button state change

### References
https://axesslab.com/disabled-buttons-suck/
and
https://a11y-101.com/development/aria-disabled

Mike-Diaz:
### Description
- Remove `disabled` property and use `aria-disabled` instead
`aria-disabled` buttons are still clickable - ensure click/button press handler is removed
Consider using a style to disable clicks? `pointer-events: none;`

- Add `allowFocusOnDisabled` prop
If true, require an additional prop: two messages, one for when a button changes from disabled to enabled, and one for 
when a button changes from enabled to disabled

- Documentation and Testing 
Has been added to storybook along with instructions on how to use. A11y tests added to the component.

Before
Button_not_focusable_no_aria

After
- Focusable
Disabled_Button_Focusable1
- Aria Live
Button_Disabled_Focusable_Aria-Live

### NOTE
[fix: add screen-reader-only styles to button #1293](https://github.com/SAP/fundamental-styles/pull/1293) must be merged with styling added for hidden `screen-reader-only` message, otherwise; the `aria-live` message will be visible next to a disabled button that has `allowFocusOnDisabled` set to **true**. Additionally, tests will not pass until PR#1293 is merged.

Dynamic reading of the helper text to alert users with screen readers the state of the button has changed does not work on IE11 or Edge using JAWS. All functionality works as intended using JAWS/Chrome or VoiceOver/Chrome.

### References
https://axesslab.com/disabled-buttons-suck/
and
https://a11y-101.com/development/aria-disabled

fixes #889

### DIFF
--- a/src/Button/Button.js
+++ b/src/Button/Button.js
@@ -99,9 +99,9 @@ const validateStateTransitionMessage = (props, propName) => {
     if (props.allowFocusOnDisable && (props.disabled !== null || props['aria-disabled']) && !props[propName]) {
         return new Error(`
         Missing or invalid value for ${propName}.
-        If this button can recieve focus when disabled (i.e. allowFocusOnDisable prop is true),
+        If this button can receive focus when disabled (i.e. allowFocusOnDisable prop is true),
         you need to provide localized string values for the enabled and disabled states.
-        These strings are used to annouce the state change by assitive techonolgies.
+        These strings are used to announce the state change by assistive technologies.
         Hence, these should be specified for a better, accessible experience.
         For e.g. 
         <Button
--- a/src/Button/Button.js
+++ b/src/Button/Button.js
@@ -1,6 +1,7 @@
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
 import React from 'react';
+import shortId from '../utils/shortId';
 import { BUTTON_OPTIONS, BUTTON_TYPES } from '../utils/constants';
 import 'fundamental-styles/dist/icon.css';
 import 'fundamental-styles/dist/button.css';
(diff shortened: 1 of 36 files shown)


jacobdevera:
Awesome work! Couple things I noticed:

1. `Emphasized Focusable` disabled button outline is not visible. Definitely overlooked, and I'd try to get the examples in `fundamental-styles` to match the ones here.
2. Just realizing the outline color seems dim, and may not pass color contrast ratio requirements for focus rings ([3:1](https://www.w3.org/WAI/WCAG21/Understanding/focus-visible-enhanced)). Also was overlooked. This is because of the low opacity passed to the disabled button. I couldn't find a way to quickly override this so that the `outline` color is fully opaque. We can try to address this in a follow-up if it requires major work.

Mike-Diaz:
@jacobdevera Updated fundamental-styles as requested. Emphasized Focusable Disabled button now shows focus ring. Additional examples of Focusable Disabled buttons now present in Button States Storybook. As mentioned in spike, aria-disabed buttons do not need to pass color contrast. https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html

src/Button/Button.js
```diff
@@ -37,18 +41,85 @@ const Button = React.forwardRef(({
...
+        } else {
+            disabledProps.disabled = disabledInAnyWay;
+        }
+        return disabledProps;
+    };
+
+    const liveMessageForScreenReaders = disabledInAnyWay ? disabledMessage : enabledMessage;
+
+    const renderButtonStateMessage = () => {
+        let content = '';
```
jacobdevera, line 66: Small nit: this seems to add a new line on every storyshot for every component that uses a button, probably due to the empty string `content`. I wonder if this should be instead be `null`.

Mike-Diaz: Oh holy moly, I was wondering why there were a bunch of blank lines when I updated the story snapshots. Yeah I agree this should be null. I will update and test to make sure there are no adverse effects. Thanks!

src/Button/Button.js
```diff
@@ -37,18 +41,85 @@ const Button = React.forwardRef(({
...
 
 Button.displayName = 'Button';
 
+const validateStateTransitionMessage = (props, propName) => {
+    if (props.allowFocusOnDisable && (props.disabled !== null || props['aria-disabled']) && !props[propName]) {
+        return new Error(`
+        Missing or invalid value for ${propName}.
+        If this button can recieve focus when disabled (i.e. allowFocusOnDisable prop is true),
+        you need to provide localized string values for the enabled and disabled states.
+        These strings are used to annouce the state change by assitive techonolgies.
```
jacobdevera, line 104: nit: `receive`, `announce`, `assistive technologies`
I know these are carryovers from nui-widgets 😅

Mike-Diaz: Didn't think to double check this haha. Will fix grammar.

PR #1141, opened 2020-07-30
fix: allow custom modal manager for DatePicker

meganaconley:
### Description

If a DatePicker component is included in a modal, pressing escape will close both the DatePicker and modal as they have different modal managers. This PR allows for a custom modal manager to be passed down from a parent component.

### DIFF
--- a/src/DatePicker/DatePicker.js
+++ b/src/DatePicker/DatePicker.js
@@ -364,6 +364,7 @@ class DatePicker extends Component {
             todayAction,
             validationState,
             weekdayStart,
+            modalManager,
             ...props
         } = this.props;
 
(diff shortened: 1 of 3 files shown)


src/DatePicker/DatePicker.js
```diff
@@ -364,6 +364,7 @@ class DatePicker extends Component {
             todayAction,
             validationState,
             weekdayStart,
+            modalManager,
```
skvale, line 367: We could add this to the propTypes for DatePicker as well
```js
/** If Popover is to be rendered in a modal, the parent modal manager can be passed as a prop */
modalManager: PropTypes.object,
```

PR #1142, opened 2020-08-01
feat: update fundamental styles to 0.11.3-rc.8

prsdthkr:
### Description
In this change we
* Update `fundamental-styles` to `0.11.3-rc.8` and support the new features and fixes therein.
* This also includes some accessibility upgrades
  * ActionBar has a glyph based back button. Enforce aria-label for it.
  * MultiInput has a glyph based popover trigger button. Enforce aria-label for it.
  * Tokens have a glyph based close/clear button. Allow setting aria-label for it.
  * It wasn't possible to remove MultiInput tokens from the input field using a keyboard. After this change, if a MultiInput token has focus, pressing `backspace` or `delete` should remove it.
    
  * Update TimePicker stories so that there are no accessibility issues in Storybook
  * Support FormLabel with colons feature (https://github.com/SAP/fundamental-styles/pull/1181)
  * Deprecate LocalizationEditor (https://github.com/SAP/fundamental-styles/pull/1261)


package.json
```diff
@@ -51,7 +51,7 @@
     "chain-function": "^1.0.1",
     "classnames": "^2.2.6",
     "detect-browser": "^5.1.0",
-    "fundamental-styles": "0.11.3-rc.7",
+    "fundamental-styles": "0.11.3-rc.8",
```
skvale, line 54: I don't see any changes that would affect us here: https://github.com/SAP/fundamental-styles/compare/v0.11.3-rc.7...v0.11.3-rc.8

src/MultiInput/MultiInput.js
```diff
@@ -179,6 +189,16 @@ class MultiInput extends Component {
             </List>
         );
 
+        let extendedButtonProps = buttonProps;
```
skvale, line 192: We should keep props unchanged.  Assigning `extendedButtonProps['aria-label'] = buttonLabel;` will also assign `aria-label` to buttonProps.

I'm pretty sure doing this in the component accomplishes the same thing without affecting props.
```jsx
                    <Button
                        aria-label={buttonLabel}
                        {...buttonProps}

File: src/ActionBar/ActionBar.js [hunk 3]
```diff
@@ -55,6 +57,14 @@ const ActionBar = React.forwardRef(({
 
     const HeadingTag = `h${headingLevel}`;
 
+    if (!buttonProps) {
+        buttonProps = {};
+    }
+
+    if (!buttonProps['aria-label']) {
+        buttonProps['aria-label'] = backButtonLabel;
```
skvale, line 65: Similar to the comment below, we shouldn't be assigning things to the props objects passed in, does this do the same thing?
```jsx
                    <Button
                        aria-label={backButtonLabel}
                        {...buttonProps}
                        compact
                        glyph='navigation-left-arrow'
                        onClick={onBackClick}
                        option='transparent' />
```
set the aria-label to backButtonLabel, unless it's defined in buttonProps, then use buttonProps['aria-label']

src/Forms/FormLabel.js
```diff
@@ -47,7 +49,9 @@ FormLabel.propTypes = {
     /** Set to **true** if child is InlineHelp component */
     isInlineHelp: PropTypes.bool,
     /** Set to **true** for required input fields */
-    required: PropTypes.bool
+    required: PropTypes.bool,
+    /** Set to **true** to add the `:` character at the end of a label as pseudo element.*/
+    withColon: PropTypes.bool
```
skvale, line 54: Given a choice between `includeColon` or `withColon`, I like `includeColon` better. 

It matches the prop `includeEmptyOption`

PR #1148, opened 2020-08-03
chore: Bump storybook to matching 6.0.0-rc.21 versions

skvale:
### Description
Keep all storybook deps in line with https://github.com/SAP/fundamental-react/pull/1146


prsdthkr:
@skvale @jbadan I think we stopped seeing all `storyDescription`s after this change. Investigating.

PR #1150, opened 2020-08-04
fix: add popoverProps to Select

meganaconley:
### Description
Support additional props to be passed through the `Select` component down to the underlying `Popper`.


src/Select/Select.js
```diff
@@ -135,6 +136,7 @@ const Select = React.forwardRef(({
 
     return (
         <Popover
+            {...popoverProps}
```
skvale, line 139: could we add placement before the spread
```jsx
<Popover
  placement='bottom-start'
  {...popoverProps}
```
I bet it will come up soon that we'll want to be able to control the Select component's placement.  We might be able to make the same argument for `widthSizingType='minTarget'` too.

meganaconley: Good point, I went ahead and added both of those before the spread.

PR #1154, opened 2020-08-07
fix: add prop for Select dropdown styling

meganaconley:
### Description
Add support for an optional prop that allows custom styling on the option dropdown list.

### DIFF
--- a/src/Select/Select.js
+++ b/src/Select/Select.js
@@ -21,7 +21,7 @@ const Select = React.forwardRef(({
     emptyAriaLabel,
     id,
     includeEmptyOption,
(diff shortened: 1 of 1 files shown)


src/Select/Select.js
```diff
@@ -129,6 +130,7 @@ const Select = React.forwardRef(({
 
     const listClassName = classnames(
         'fd-list--dropdown',
+        listBoxClassName,
```
jacobdevera, line 133: Not sure if this affects the styling at all, but should the `listBoxClassName` be last to be consistent with the other components that insert the consumer class name prop at the end?

PR #1156, opened 2020-08-08
fix: MessageStrip Accessibility + useUniqueId hook

### LINKED ISSUE #1153: aria-controls value is hard coded in MessageStrip
### Description
aria-controls value is hard coded in MessageStrip
https://github.com/SAP/fundamental-react/blob/6235d2753bf83fd1e8ad15e4df6eadc3cc138dcc/src/MessageStrip/MessageStrip.js#L58-L65

Figure out the correct semantics of a dismissable alert and add aria attributes accordingly, for accessibility.

### Versions
**fundamental-react:** version 0.11.1-rc.22

prsdthkr:
### Description
In this change, we
* remove the hard-coded `aria-controls` value for the `MessageStrip` close button
* add stories for MessageStrip states with custom children and link
image
* create and use the `useUniqueId` hook to generate IDs in functional components

fixes #1153

### Before
image

### After
image

### Know Issues
https://github.com/SAP/fundamental-react/issues/1155


src/MessageStrip/MessageStrip.js
```diff
@@ -47,17 +48,20 @@ const MessageStrip = (props) => {
         className
     );
 
+    const alertId = otherProps?.id || shortId.generate();
```
jbadan, line 51: Can we create a `useUniqueId()` hook like in widgets -

```
import shortid from './shortId';
import { useState } from 'react';

export default () => {
    const [uniqueId, setUniqueId] = useState();
    if (uniqueId) return uniqueId;
    const id = shortid.generate();
    setUniqueId(id);
    return id;
};
```

src/MessageStrip/MessageStrip.js
```diff
@@ -48,10 +48,10 @@ const MessageStrip = (props) => {
         className
     );
 
-    const alertId = otherProps?.id || shortId.generate();
+    const alertId = otherProps?.id || useUniqueId();
```
skvale, line 51: We can't [conditionally call hooks](https://reactjs.org/docs/hooks-rules.html) like this, or with ternary expressions like others.

We can split it onto separate lines though,
```js
    const generatedId = useUniqueId();
    const alertId = otherProps?.id || generatedId;
```

PR #1160, opened 2020-08-12
chore: change standard-version commit message format

scottheron:
### Description
`-m` or `--message` is deprecated https://github.com/conventional-changelog/standard-version/blob/master/command.js#L33

Instead of using `%s` we can use `{{currentTag}}` -- https://github.com/conventional-changelog/standard-version/blob/ae032bfa9268a0a14351b0d78b6deedee7891e3a/test.js#L1244

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "size": "npm run build && size-limit",
     "size:debug": "npm run build && size-limit --why",
     "start": "npm run storybook",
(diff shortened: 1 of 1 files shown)


jbadan:
@scottheron Looks good! I edited your title/description to remove links to our internal jira - we try to avoid posting internal links on the open source projects :)

PR #1166, opened 2020-08-13
fix: improve screen reader instructions for DatePicker

jacobdevera:
### Description
- Remove `'assertive'` `aria-live` region on header, and instead use polite region for all announcements. This ensures the initially-focused date is read out and the previous/next buttons aren't unexpectedly read when opening the calendar.
- Provide separate instructions for year view and month view. This helps add context when switching to month and year view, and helps work around the fact that the aria-live announcement will not read out if the same instructions are simply re-populated.
- Provide instructions after selecting a first date for range selection.

### Testing
- JAWS and VoiceOver: ensure announcements described above are properly done
- VoiceOver + Chrome has a known issue where it will read polite announcements twice

### DIFF
--- a/src/Calendar/Calendar.js
+++ b/src/Calendar/Calendar.js
@@ -411,7 +411,7 @@ class Calendar extends Component {
 
         this.setState({
             currentDateDisplayed: day,
-            screenReaderText: '', // selected day
+            screenReaderText: isRangeEnabled ? this.props.localizedText.rangeInstructions : '',
             selectedDate: day,
             arrSelectedDates: selectedDates
         }, function() {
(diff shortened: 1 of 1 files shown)


src/Calendar/Calendar.js
```diff
@@ -166,6 +167,7 @@ class Calendar extends Component {
 
         this.setState({
             currentDateDisplayed: newDate,
+            screenReaderText: this.props.localizedText.dayInstructions,
```
meganaconley, line 170: Should this be `monthInstructions` and the one below be `yearInstructions`?

jacobdevera: This is upon selection, so this is when it switches back to the day view.

meganaconley: Ah ok I see it now.

PR #1167, opened 2020-08-13
fix: list selection behavior + accessibility

### LINKED ISSUE #1164: List: selected state for List.Selection
### Description

Add ability to pass `selected` prop to `List.Item` when using `List.Selection`

See: https://fundamental-styles.netlify.app/?path=/docs/components-list-standard--standard#selection
https://ui-fundamentals.slack.com/archives/CFDTYK9QT/p1597309029001900

Relevant line of code: https://github.com/SAP/fundamental-react/blob/master/src/List/_ListItem.js#L30

### Versions
**fundamental-react:** 0.11.1-rc.26

---

_**DISCLAIMER:**

prsdthkr:
### Description
In this change, we
* add `selected` property for `List.Selection` to allow setting the checked state of the checkbox. This property will be internally set by the parent `List.Item` component based on the `List.Item.selected` property.
* add a `selectable` property to `List` to
  * add `fd-list--selection` class to `<ul>`
  * set `<ul>` role='listbox'
* remove `List.Header` and `List.Footer` from the list structure, they should be passed through the new `header` and `footer` List props.
* `List.Header` renders `<h2-6>` depending on `List.headerLevel`
* `List.Footer` renders `<span>`
* add an `id` property to `List` to associate `List.Header` as label
* add `checkBoxAriaLabel ` to `List.Selection` for accessibility
* add `headerClassName` and `footerClassName` props to allow custom classes

### Screen cap

**Before**
list-selection-bug

**After**
list-selection-fixed

### Know issues
* ~deviates from fundamental-styles markup for accessibility reasons~

fixes #1164


src/List/List.js
```diff
@@ -36,11 +40,31 @@ const List = React.forwardRef(({
...
     return (
-        <ul {...props} className={ListClasses}
-            ref={ref}>
-            { React.Children.map(children, child => React.cloneElement(child, { hasByline, navigation, partialNavigation })) }
-        </ul>
+        <>
+            {listHeader && React.cloneElement(listHeader, { id: `${listId}-label` })}
+            <ul
+                {...props}
+                {...(selectable ? { 'role': 'listbox' } : null)}
```
skvale, line 55: Slightly easier to read syntax:
```jsx
role={selectable ? 'listbox' : null}
```

src/List/List.js
```diff
@@ -36,11 +40,31 @@ const List = React.forwardRef(({
...
+                aria-labelledby={listHeader ? `${listId}-label` : null}
+                className={ListClasses}
```
skvale, line 57: quick unrelated change `ListClasses` -> `listClasses`

src/List/List.js
```diff
@@ -36,11 +40,31 @@ const List = React.forwardRef(({
...
+                id={`${listId}-list`}
+                ref={ref}>
+                { React.Children.map(children, child => {
+                    if (!(isHeader(child) || isFooter(child)) ) {
```
skvale, line 61: probably a nitpick
```js
if (!(isHeader(child) || isFooter(child)) ) {
vs
if (!isHeader(child) && !isFooter(child)) {
```
But it brings up a bigger issue to me, we aren't supporting custom components or fragments.  We **should** be able to consume this code and add the right props to the right children

```jsx
const MyOtherItems = () => (
   <React.Fragment>
        <List.Item>
            <List.Text>List Item 2</List.Text>
        </List.Item>
        <List.Item>
            <List.Text>List Item 3</List.Text>
        </List.Item>
        <List.Item>
            <List.Text>List Item 4</List.Text>
        </List.Item>
   </React.Fragment>
);
// ....
const MyListFooter = () => <List.Footer>List Footer</List.Footer>
// ...
<List>
        <List.Header>List Header</List.Header>
        <List.Item>
            <List.Text>List Item 1</List.Text>
        </List.Item>
        <MyOtherItems />
        <MyListFooter />
    </List>
```

src/utils/useUniqueId.js
```diff
@@ -1,7 +1,8 @@
 import shortid from './shortId';
 import { useState } from 'react';
 
-export default () => {
+export default (defaultId) => {
+    if (defaultId) return defaultId;
     const [uniqueId, setUniqueId] = useState();
```
skvale, line 6: I think this would qualify as conditionally calling a hook, breaking the rules of hooks.
If the first call from an initial render of a component didn't have a defaultId, it would call `useState`, if the next call did have defaultId it wouldn't call `useState`

src/List/List.js
prsdthkr: Thanks for pointing this out. I removed the children scanning and made the header and footer as List props.

skvale: That's an interesting idea to get around the problem.  I think it would have to look more like this though
```jsx
<List
  footer='My footer'
  header='My header'>
   // ...
```
and the List component would render them like
```jsx
return (
  <>
    {header && <List.Header id={`${listId}-label`}>{header}</List.Header>}
    // ...
    {footer && <List.Footer>{footer}</List.Footer>}
  </>
```

prsdthkr: This will mean that we cannot get fancy with the header and footer [like here](https://github.com/SAP/fundamental-react/pull/1167/files#diff-e225e0a8c2fdc1854374ffe9ca6096deR150-R153), right?

skvale: I guess it would look like
```jsx
footer = {<React.Fragment><List.Text><b>Total</b></List.Text><List.Text secondary>7</List.Text></React.Fragment>}
```
instead of
```jsx
        footer={
            <List.Footer>
                <List.Text><b>Total</b></List.Text>
                <List.Text secondary>7</List.Text>
            </List.Footer>
        }
```

src/List/List.js
```diff
@@ -53,14 +70,22 @@ List.propTypes = {
     className: PropTypes.string,
     /** Set to **true** to enable compact mode */
     compact: PropTypes.bool,
+    /** The list footer as a String or a React component.*/
+    footer: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
     /** Set to **true** if any list item has a byline. */
     hasByline: PropTypes.bool,
+    /** The list header as a String or a React component.*/
+    header: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
```
skvale, line 78: > React component
> PropTypes.object

It's more accurate to say React node and PropTypes.node

src/List/_ListFooter.js
```diff
@@ -13,9 +13,9 @@ const ListFooter = ({
     );
 
     return (
-        <li {...props} className={ListItemClasses}>
+        <span {...props} className={ListItemClasses}>
```
skvale, line 16: This seems right to change the footer to a span, but do we need to stay in line with fundamental-styles here?

I've looked at a few o their examples and they all still show the footer as a `li`, like here: https://github.com/SAP/fundamental-styles/blob/master/stories/list/standard/standard-list.stories.js#L301

skvale: I guess this is part of 
> deviates from fundamental-styles markup for accessibility reasons

from your description, but is that OK to do?  Do we document our deviations?

prsdthkr: Trying to get this change in fundamental-styles: https://github.com/SAP/fundamental-styles/pull/1397

We can either jump the gun and get this change in fd-react or wait for the fd-styles change and subsequent fd-styles upgrade in fd-react.

prsdthkr: I'm for jumping the gun, as we will eventually be up to spec when fd-styles is upgraded :)

PR #1170, opened 2020-08-19
fix: Checkbox classname

prsdthkr:
### Description
In this change, we update the Checkbox component to apply `className` value to the wrapping `FormItem` component instead of the internal `input`. This is to be consistent with the behavior in other Form components.

### Questions
Is this a breaking change?

### DIFF
--- a/src/Forms/Checkbox.test.js
+++ b/src/Forms/Checkbox.test.js
@@ -88,6 +88,26 @@ describe('<Checkbox />', () => {
                 element.find('.fd-form-label').getDOMNode().attributes['data-sample'].value
             ).toBe('Sample');
         });
+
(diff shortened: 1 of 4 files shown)


src/Forms/Checkbox.test.js
```diff
@@ -88,6 +88,26 @@ describe('<Checkbox />', () => {
...
+
+        test('should set labelClassName on FormLabel', () => {
+            const wrapper = setup({
+                labelClassName: 'wonderful-styles'
+            });
+
+            expect(
+                wrapper.find('FormLabel').getDOMNode().classList
+            ).toContain('wonderful-styles');
+        });
```
jbadan, line 110: Thanks! 🚢

PR #1171, opened 2020-08-19
feat: create Title component

### LINKED ISSUE #1162: Title: new component
### Description
Implement the new component for which the styles were available after this change: https://github.com/SAP/fundamental-styles/pull/1338

meganaconley:
### Description
Create the Title component and implement in other components that use headings.

fixes #1162 

From https://github.com/SAP/fundamental-styles/pull/1338/

Will wait to merge until fundamental-styles is upgraded as a breaking change was merged in just before this new component.


src/Title/Title.js
```diff
@@ -0,0 +1,49 @@
...
+
+Title.propTypes = {
+    /** Determines the heading tag rendered on the page, also known as the semantic level of the heading. */
+    level: PropTypes.oneOf(headingLevels).isRequired,
+    /** Node(s) to render within the component */
+    children: PropTypes.node,
+    /** CSS class(es) to add to the element */
+    className: PropTypes.string,
+    /** If the desired visual appearance is different from the default style given to the semantic level of the header, provide the desired heading level style to apply. */
+    levelStyle: PropTypes.oneOf(headingLevels),
```
skvale, line 44: What if `Title` took number props for `level` and `levelStyle`
```
    level: CustomPropTypes.range(2, 6),
    levelStyle: CustomPropTypes.range(2, 6)
```
It would match up with how we pass the heading tags in with other components and then we could remove the `h${headingLevel}` logic from the scattered components using this and move it here.  Thoughts?

meganaconley: Sure. I think I would have rather had everything use `h#` instead as it seems a bit off to divorce the number from the actual html tag. But I don't think it's worth making a breaking change to update all the other components, so this one might as well match.

src/ActionBar/ActionBar.js
```diff
@@ -55,7 +57,7 @@ const ActionBar = React.forwardRef(({
         actionClassName
     );
 
-    const HeadingTag = `h${headingLevel}`;
+    const titleStyle = headingStyle || headingLevel;
```
skvale, line 60: Kind of minor, but we could simplify this a bit if we just passed `levelStyle={headingStyle}>` into `<Title />` since the same logic happens there if it's not defined
```js
const levelClass = levelStyle ? `h${levelStyle}` : `h${level}`;
```

meganaconley:
@skvale This PR will go in after #1172

Issue #1159, opened 2020-08-11
Select onFocus and onBlur happen on a single click


skvale:
### Description

You can see focus and blur happen on a single click of the Select component

![focus-blur](https://user-images.githubusercontent.com/5314713/89958158-d3371500-dbfe-11ea-9ea9-3c8a28e6f298.gif)

### Versions
**fundamental-react:**

---

_**NOTE:** Where applicable, please include uncropped screen captures._

_**DISCLAIMER:**
After triaging an issue, the fundamental-react team will see if it can be reproduced or confirmed. If more information is needed, the fundamental-react team will contact the author.  Any issues awaiting responses from the author for more than 7 days will be closed.  The author can re-open the issue at a later time if they can present the requested information._

PR #1183, opened 2020-08-25
feat: normalize FormValidationOverlay usage in other components

jbadan:
### Description

1. added unit tests for FormValidationOverlay
1. consistent apply new prop `validationOverlayProps` to all components that consume `FormValidationOverlay`: `Checkbox`, `FormInput`, `FormTextArea`, `Select`, `InputGroup`, `SearchInput`, `Select`, `StepInput`
1. unit tests for  `validationOverlayProp` in all consuming components
1. `validationOverlayProp.formMessageProps` also spread to `FormMessage` when used inside component (outside of `ValidationOverlay` - see `Select` or `SearchInput`). This way, if you are expecting `FormMessage` to receive additional classes, it will whether it's inside the dropdown or as it's own popover. This will be useful in nui-widgets-lab for override Gateway theming colors.
1. `_FormValidationOverlay` (not exported to consumers) now spreads additional props to the wrapping `<div>` element via `wrapperProps`. I couldn't use `...rest` because that creates a rabbithole of problems with the way `Popover` is cloning elements. Possibly worth a followup, but is a ton of work and creates a pretty confusing output just for continuity sake.

BREAKING CHANGES:
1. Removed `fd-list--has-message` class from Select's dropdown - it was adding unnecessary padding. It was a relic from before FormMessage rendered inside a Popover
1. `validationClassName` renamed and moved inside of `validationOverlayProps`, use like `validationOverlayProps.className`

There will be at least 2 additional prs for Select - 1 to add the ability to pass custom classes to each component, the second addressing https://github.com/SAP/fundamental-react/issues/1159

### DIFF
--- a/src/Forms/Checkbox.js
+++ b/src/Forms/Checkbox.js
@@ -29,7 +29,7 @@ const Checkbox = React.forwardRef(({
     name,
     onChange,
     value,
-    validationClassName,
+    validationOverlayProps,
     validationState,
     ...props
 }, ref) => {
@@ -96,7 +96,7 @@ const Checkbox = React.forwardRef(({
 
     return validationState ? (
         <FormValidationOverlay
-            className={validationClassName}
+            {...validationOverlayProps}
             control={checkbox}
             validationState={validationState} />
     ) : checkbox;
@@ -143,8 +143,15 @@ Please ensure you are either using a visible \`FormLabel\` or an \`aria-label\` 
     labelProps: PropTypes.object,
     /** Sets the `name` for the checkbox input */
     name: PropTypes.string,
-    /** Additional classes to apply to validation popover */
-    validationClassName: PropTypes.string,
+    /** Additional props to be spread to the ValidationOverlay */
+    validationOverlayProps: PropTypes.shape({
+        /** Additional classes to apply to validation popover's outermost `<div>` element  */
+        className: PropTypes.string,
+        /** Additional props to be spread to the ValdiationOverlay's FormMessage component */
+        formMessageProps: PropTypes.object,
+        /** CSS class(es) to add to the ValidationOverlay's reference `<div>` element */
+        referenceClassName: PropTypes.string
+    }),
     /** An object identifying a validation message.  The object will include properties for `state` and `text`; _e.g._, \`{ state: \'warning\', text: \'This is your last warning\' }\` */
     validationState: PropTypes.shape({
         /** State of validation: 'error', 'warning', 'information', 'success' */
--- a/src/Forms/Checkbox.js
+++ b/src/Forms/Checkbox.js
@@ -29,6 +29,7 @@ const Checkbox = React.forwardRef(({
     name,
     onChange,
     value,
+    validationClassName,
     validationState,
     ...props
 }, ref) => {
@@ -95,6 +96,7 @@ const Checkbox = React.forwardRef(({
 
     return validationState ? (
         <FormValidationOverlay
+            className={validationClassName}
             control={checkbox}
             validationState={validationState} />
     ) : checkbox;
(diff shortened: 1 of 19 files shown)


src/Forms/Checkbox.js
```diff
@@ -143,8 +143,15 @@ Please ensure you are either using a visible \`FormLabel\` or an \`aria-label\`
     labelProps: PropTypes.object,
     /** Sets the `name` for the checkbox input */
     name: PropTypes.string,
-    /** Additional classes to apply to validation popover */
-    validationClassName: PropTypes.string,
```
skvale, line 147: Should moving this prop to `validationOverlayProps.className` be listed as a breaking change as well?

src/Forms/_FormValidationOverlay.test.js
```diff
@@ -0,0 +1,137 @@
...
+describe('<FormValidationOverlay />', () => {
+    const setup = (props = {}) => mount(
+        <FormValidationOverlay control={(<input onChange={() => {}} value='Test' />)} {...props} />
+    );
+
+    afterEach(() => {
+        document.body.innerHTML = '';
+    });
+
+    const getFormMessage = () => document.body.querySelector('.fd-popover__popper > div > .fd-form-message');
```
skvale, line 15: Just an idea or an alternative, I've seen tests that include
```jsx
// Ensure the popper is accessible with enzyme
jest.mock('react-dom', () => {
    const original = jest.requireActual('react-dom');
    return {
        ...original,
        createPortal: (node) => node,
    };
});
```
so you can still access the elements via `enzyme`

src/Forms/Checkbox.js
jbadan: 100%! Good catch

src/Forms/_FormValidationOverlay.test.js
```diff
@@ -0,0 +1,137 @@
...
+        test('should show the FormMessage onFocus', async() => {
+            let element = setup({ validationState: { state: 'error', message: 'test message' } });
+
+            await act(async() => {
+                element.find('.fd-popover').simulate('focus');
+            });
+
+            const messageNode = getFormMessage();
+
+            expect(messageNode).toBeDefined();
```
skvale, line 103: I'm pretty sure `expect(null).toBeDefined()` passes.  I f`querySelector` can't find an element it returns `null`, so this expect will always pass.

jbadan: Well... crap haha bad unit test!

src/Forms/_FormValidationOverlay.test.js
jbadan: I like this, but I'd like to apply it globally in `config/jest/setup.js` so we don't have some unit tests with it/some without. It makes 450 unit tests fail at the moment so I'll probably skip it on this pr 😆

src/Forms/_FormValidationOverlay.test.js
jbadan: The one after this is going to have the same problem - I'll rework both

src/Forms/_FormValidationOverlay.test.js
```diff
@@ -0,0 +1,116 @@
...
+                });
+            });
+
+            expect(
+                element.find('FormMessage').getDOMNode().attributes['data-sample'].value
+            ).toBe('Sample');
+        });
+    });
+
+    describe('Interactions', () => {
```
jbadan, line 78: @skvale Looks like you solved this exact problem in tooltip in widgets recently - I copied and applied what you did there to get this working 🎉

jbadan:
I forgot to merge this as a breaking change 😞

bcullman:
@jbadan any chance we would use `git --amend` to fix this?

jbadan:
@bcullman We could try... but that prop was added 1 day before and removed 1 day after. We're most likely the only ones that noticed it came and went 😆

PR #1184, opened 2020-08-25
feat: add textClassName prop and unit test

Mike-Diaz:
### Description

This adds the `textClassName` prop inside of `Checkbox` in order for custom classes to be applied to the element. This is necessary for styles to target the class name and be applied directly to the children. This does not change the way the component originally works since the standard class will always be applied.

Linked to PR:  https://github.concur.com/coreui/nui-widgets-lab/pull/124

Added Unit tests


src/Forms/Checkbox.test.js
```diff
@@ -119,6 +119,16 @@ describe('<Checkbox />', () => {
             ).toContain('wonderful-styles');
         });
 
+        test('should set validationClassName on the popover', () => {
```
skvale, line 122: > should set childrenClassName on the popover

src/Forms/Checkbox.js
```diff
@@ -119,6 +125,8 @@ Please ensure you are either using a visible \`FormLabel\` or an \`aria-label\`
             );
         }
     },
+    /** Additional classes to apply to children */
+    childrenClassName: PropTypes.string,
```
jbadan, line 129: I think we have to be a little more careful with the name here - if the child is not a string, it won't add this class. Do we want that? Do we not? I don't know...

src/Forms/Checkbox.test.js
Mike-Diaz: Oh that is super weird. I was sure I changed the title of this test, thanks!

src/Forms/Checkbox.js
Mike-Diaz: I also thought perhaps this was not the best name. Would you have a suggestion of a more semantic, or explicit name? Perhaps `stringChildrenClassName`?

jbadan: I think `textClassName` would be fine

jbadan (approved):

jacobdevera (approved):

PR #1186, opened 2020-08-26
fix: select onfocus onblur callbacks called at appropriate times

### LINKED ISSUE #1159: Select onFocus and onBlur happen on a single click

jbadan:
### Description

`onBlur` is now only triggered when leaving the component - via `onClickOutside` or by leaving the select field **AND** the dropdown list. Previously, `onBlur` was being fired when leaving the select field and navigating inside of the dropdown list items.

 

fixes #1159

### DIFF
--- a/src/Select/Select.test.js
+++ b/src/Select/Select.test.js
@@ -85,33 +85,123 @@ describe('<Select />', () => {
     });
 
     describe('interactions', () => {
-        let onSelect;
-        let element;
-        beforeEach(() => {
-            onSelect = jest.fn();
-            element = mount(
-                <Select onSelect={onSelect} options={options} />
+        describe('onSelect', () => {
+            let onSelect;
+            let element;
+            beforeEach(() => {
+                onSelect = jest.fn();
+                element = mount(
+                    <Select onSelect={onSelect} options={options} />
+                );
+                element.find('.fd-select__button').simulate('click');
+            });
+
+            afterEach(() => {
+                let event = new MouseEvent('mousedown', {});
+                document.dispatchEvent(event);
+            });
+
+            test('should call onSelect when the first checkbox option is clicked', () => {
+                element.find('.fd-list__item').at(0).simulate('click');
+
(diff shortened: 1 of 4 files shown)


src/Select/Select.test.js
```diff
@@ -85,33 +85,123 @@ describe('<Select />', () => {
...
+            });
+
+            expect(focusSpy).toHaveBeenCalledTimes(1);
+
+            expect(
+                focusSpy
+            ).toHaveBeenCalledWith(
+                expect.objectContaining({
+                    target: wrapper.find('.fd-select').getDOMNode()
+                })
```
jbadan, line 136: These onSelect tests remain unchanged - I put them inside their own describe block so we could keep all the interactions in the interaction describe block

src/Select/__stories__/Select.stories.js
```diff
@@ -76,6 +77,8 @@ export const dev = () => (
     <Select
         compact={boolean('compact', false)}
         disabled={boolean('disabled', false)}
+        onBlur={() => console.log('on-blur-callback')}
+        onFocus={() => console.log('on-focus-callback')}
```
jbadan, line 81: I originally tried to use addon-actions, but that slowed storybook down so much. I think there's a bug there

skvale:
Side note  ore related to the FormValidationOverlay, but it weird that the validation shows on mousedown before the click happens to open/close the control.

select-error

jbadan:
@skvale That is weird, but I'm not really sure what else it would look like for Select... it makes sense to be that was for a component with a keyboard-entry input (like combobox). 

Should the Popover FormMessage (trying to differentiate here between the single floating message and the one that is inside the dropdown Popover) only appear onFocus and _not_ onClick (as onClick is always going to trigger the dropdown opening)?
OR
Should  the Popover FormMessage always be open when it exists (ie there is a validation state) and then is only removed when dropdown popover is open?

PR #1188, opened 2020-08-26
chore: cleanup some test warnings

jbadan:
### Description

Cleans up test warnings from a couple components (definitely not all of them!)

Breaking change: removed unused `spinners` prop from `TimePicker` and `Time` components

fixes #issueid

### DIFF
--- a/src/Menu/_MenuItem.js
+++ b/src/Menu/_MenuItem.js
@@ -59,25 +59,20 @@ const MenuItem = ({
                 children.props.className
             );
 
-            const addonChildBefore = addonBefore ? (<span
-                key='addonChildBefore'
-                {...addonProps}
-                className={addonBeforeClassnames} />) : null;
-            const addonChildAfter = addonAfter ? (<span
-                key='addonChildAfter'
-                {...addonProps}
-                className={addonAfterClassnames} />) : null;
-
             return (
                 <>
-                    {React.Children.toArray(children).map((child, key) => {
-                        return React.cloneElement(child, {
-                            'className': childrenClassnames,
-                            key: `menu-item-${key}`,
-                            ...urlProps
-                        },
-                        [addonChildBefore, (<span className='fd-menu__title' key='menuTitle'>{child.props.children}</span>), addonChildAfter]);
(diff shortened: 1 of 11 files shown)


src/Menu/_MenuItem.js
```diff
@@ -59,19 +59,26 @@ const MenuItem = ({
...
-                    {React.Children.toArray(children).map(child => {
+                <>
+                    {React.Children.toArray(children).map((child, key) => {
                         return React.cloneElement(child, {
                             'className': childrenClassnames,
+                            key: `menu-item-${key}`,
                             ...urlProps
                         },
-                        [addonChildBefore, (<span className='fd-menu__title'>{child.props.children}</span>), addonChildAfter]);
+                        [addonChildBefore, (<span className='fd-menu__title' key='menuTitle'>{child.props.children}</span>), addonChildAfter]);
```
skvale, line 79: I get that we're adding keys because this stuff is an array,  but it seems really awkward that it is an array to begin with.

and what's with
```js
addonChildBefore, (<span className='fd-menu__title' key='menuTitle'>{child.props.children}</span>), addonChildAfter]
```
being passed in as the `this` argument of `.map`.  

Anyways, I think this block of code is *trying* to be this:
```jsx
return (
    <>
        {addonBefore && <span {...addonProps} className={addonBeforeClassnames} />}
        <span className='fd-menu__title'>
            {React.Children.map(children, child => {
                return React.cloneElement(child, {
                    className: childrenClassnames,
                    ...urlProps
                });
            })}
        </span>
        {addonAfter && <span {...addonProps} className={addonAfterClassnames} />}
    </>
);
```

jbadan: Yeah I honestly didn't think too hard about it 😆  I'll make this change

jbadan:
@skvale I think the prop was used a loooong time ago, but we might as well mark it as a breaking change. The only thing I could think of is a situation where a consumer was adding it from some legacy implementation and now they'd have to update their snapshots (the same situation where I had to update snapshots in this pr)

PR #1193, opened 2020-08-29
fix: fundamental Switch state doesn't refresh 

### LINKED ISSUE #977: fundamental Switch state doesn't refresh
### Description
// With code below, when clicking the test button, Switch's checked state should be refreshed but in fact it's not.

import { Switch } from "fundamental-react";

const SwitchTest = () => {
  const [value, setValue] = useState(true);
  const handleClick = () => {
    setValue(!value);
  }
  return `<div><div><Button onClick={handleClick}>test</Button></div > <SwitchWraper checked={value} /></div>`;
};

const SwitchWraper = (props: any) => {
  return `<Switch {...props}></Switch>`;
}

export const FundamentalSwitchIssue = () => {
  return `<SwitchTest />`;
}

### Versions
**fundamental-react:**
0.9.0
---

_**DISCLAIMER:**

rjstyles:
### Description
Deriving Switch's checked state from the `checked` prop, which fixes the refresh of checked state on update of the component.

fixes #977

### DIFF
--- a/src/Switch/Switch.js
+++ b/src/Switch/Switch.js
@@ -4,7 +4,7 @@ import FormLabel from '../Forms/FormLabel';
 import keycode from 'keycode';
 import PropTypes from 'prop-types';
 import SwitchItem from './_SwitchItem';
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import 'fundamental-styles/dist/icon.css';
 import 'fundamental-styles/dist/switch.css';
 
@@ -30,6 +30,10 @@ const Switch = React.forwardRef(({
 }, ref) => {
 
     let [isChecked, setIsChecked] = useState(!!checked);
+
+    useEffect(() => {
+        if (checked) setIsChecked(checked);
+    }, [checked]);
 
     const handleChange = (e) => {
         setIsChecked(!isChecked);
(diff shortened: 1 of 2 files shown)


src/Switch/Switch.js
```diff
@@ -31,6 +31,10 @@ const Switch = React.forwardRef(({
 
     let [isChecked, setIsChecked] = useState(!!checked);
 
+    useEffect(() => {
+        if (checked) setIsChecked(checked);
+    }, [checked]);
+
```
jbadan, line 37: @rjstyles Can you add a test for this? Thanks!

skvale: I think we'd want to remove the `if(checked)` so it could set the state to checked=false
```jsx
    useEffect(() => {
        setIsChecked(checked);
    }, [checked]);
```

rjstyles: > @rjstyles Can you add a test for this? Thanks!

Sure, will do that

rjstyles: Yeah @skvale, Thanks for the review, we can do something like below to also handle  if `checked` is `null` or `undefined`

```
useEffect(() => {
    setIsChecked(!!checked);
}, [checked]);
```

@jbadan I will add the tests for this.

dariadomagala:
Hello team! 
Is there a chance that you will do a 0.13.2 patch release with this bug fix? We need just this one fix and switching to 0.14rc breaks too many things

PR #1196, opened 2020-09-03
fix: multiple validationState messages + popover validation message styles

prsdthkr:
## Description
In this change we
* Update ComboboxInput, DatePicker, and MultiInput to not show input aligned semantic messages when the popover is open. This is because the popover is expected to show this message. Hence we do not want to duplicate the semantic message.
* Use appropriate styles for validation messages that are inside popovers

## Before
image

image

image

## After
image
image
image

### DIFF
--- a/src/ComboboxInput/ComboboxInput.js
+++ b/src/ComboboxInput/ComboboxInput.js
@@ -403,7 +403,7 @@ const ComboboxInput = React.forwardRef(({
                 disabled={disabled}
                 onClick={handleInputGroupClick}
                 validationOverlayProps={validationOverlayProps}
-                validationState={validationState}>
+                validationState={!isExpanded && validationState}>
                 <FormInput
                     autoComplete='off'
                     {...inputProps}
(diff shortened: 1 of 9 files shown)


skvale:
This is probably something to bring back to fundamental-styles, but it seems like the box-shadow should be reworkd.
- Why is the box shadow on the multi-input so visible compared to other components
- It seems like the component overlaps the box shadow in the datepicker, but it doesn't overlap where it has border radius so you get this little part visible -- and the border colors are different
Screen Shot 2020-09-03 at 8 47 09 AM

prsdthkr:
@skvale, for now, I've fixed this by reusing `fd-list__message` and associated styles but I've also created an issue in fd-styles asking for new styles for such components https://github.com/SAP/fundamental-styles/issues/1579

skvale:
LGTM, but should we create an issue or follow up work after the fundamental-styles pr?  Something to track that these components have clipped borders

Screen Shot 2020-09-04 at 8 50 27 AM

Screen Shot 2020-09-04 at 8 51 09 AM

skvale (approved):

PR #1203, opened 2020-09-08
feat: update fundamental-styles to 0.12.0-rc.44

jacobdevera:
### Description
BREAKING CHANGE
- Implements breaking changes from fundamental-styles up to 0.12.0-rc.44
    - rc.21 - moving icons from pseudoelements to `<i>` tag in button
        - icons now in a `<i>` tag in `Icon` and button text moved into a `<span>` with class ‘fd-button__text’
- adds `iconBeforeText`, `iconProps`, and `textClassName` prop to `Button`
- adds `ariaHidden` and `ariaLabel` props to `Icon`

### DIFF
--- a/src/ActionBar/__stories__/__snapshots__/ActionBar.stories.storyshot
+++ b/src/ActionBar/__stories__/__snapshots__/ActionBar.stories.storyshot
@@ -20,6 +20,7 @@ exports[`Storyshots Component API/ActionBar Dev 1`] = `
           type="button"
         >
           <i
+            aria-hidden="true"
             className="sap-icon--navigation-left-arrow"
           />
         </button>
@@ -212,6 +213,7 @@ exports[`Storyshots Component API/ActionBar Primary 1`] = `
           type="button"
         >
           <i
+            aria-hidden="true"
             className="sap-icon--navigation-left-arrow"
           />
         </button>
@@ -305,6 +307,7 @@ exports[`Storyshots Component API/ActionBar With a ContextualMenu 1`] = `
                 type="button"
               >
                 <i
+                  aria-hidden="true"
                   className="sap-icon--vertical-grip"
                 />
               </button>
--- a/src/ActionBar/__stories__/__snapshots__/ActionBar.stories.storyshot
+++ b/src/ActionBar/__stories__/__snapshots__/ActionBar.stories.storyshot
@@ -15,10 +15,14 @@ exports[`Storyshots Component API/ActionBar Dev 1`] = `
       >
         <button
           aria-label="Back to home"
-          className="fd-button fd-button--transparent fd-button--compact sap-icon--navigation-left-arrow"
+          className="fd-button fd-button--transparent fd-button--compact"
           onClick={[Function]}
           type="button"
-        />
+        >
+          <i
+            className="sap-icon--navigation-left-arrow"
+          />
+        </button>
       </div>
       <div
         className="fd-action-bar__title"
@@ -36,13 +40,21 @@ exports[`Storyshots Component API/ActionBar Dev 1`] = `
           className="fd-button"
           type="button"
         >
-          Button
+          <span
+            className="fd-button__text"
+          >
+            Button
+          </span>
         </button>
         <button
           className="fd-button fd-button--emphasized"
           type="button"
         >
-          Button
+          <span
+            className="fd-button__text"
+          >
+            Button
+          </span>
         </button>
       </div>
     </div>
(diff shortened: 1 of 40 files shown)


src/ActionBar/__stories__/__snapshots__/ActionBar.stories.storyshot
```diff
@@ -15,10 +15,14 @@ exports[`Storyshots Component API/ActionBar Dev 1`] = `
...
         <button
           aria-label="Back to home"
-          className="fd-button fd-button--transparent fd-button--compact sap-icon--navigation-left-arrow"
+          className="fd-button fd-button--transparent fd-button--compact"
           onClick={[Function]}
           type="button"
-        />
+        >
+          <i
+            className="sap-icon--navigation-left-arrow"
```
jbadan, line 23: Do all these icons need aria-hidden attributes like we do in widgets?

src/Select/Select.js
```diff
@@ -134,7 +133,11 @@ const Select = React.forwardRef(({
             ref={divRef}>
             <div className={selectControlClasses}>
                 <span aria-label={selectAriaLabel} className={textContentClassNames}>{textContent}</span>
-                {!readOnly && <span className={triggerClassNames} />}
+                {!readOnly &&
+                    <span className={triggerClassNames}>
+                        <i className='sap-icon--slim-arrow-down' />
```
jbadan, line 138: Can this use the Icon component + does it need some sort of aria-label?

src/Button/Button.js
```diff
@@ -86,7 +101,9 @@ const Button = React.forwardRef(({
                 ref={ref}
                 selected={selected}
                 type={typeAttr}>
-                {children}
+                {iconBeforeText && glyph && <i aria-hidden='true' className={iconClasses} />}
+                {children && <span className={buttonTextClasses}>{children}</span>}
+                {!iconBeforeText && glyph && <i aria-hidden='true' className={iconClasses} />}
```
jbadan, line 106: Same here, can these use the Icon component + remove import of Icon.css?

jbadan (changes requested):
Can we reuse the Icon component where possible + augment it to have warnings like in widgets for aria-labels/aria-hidden?

src/Icon/Icon.js
```diff
@@ -22,9 +22,13 @@ const Icon = React.forwardRef(({ glyph, size, className, ...props }, ref) => {
         style = { fontSize: ICON_SIZES[size] };
     }
 
+    const Tag = isInButton ? 'i' : 'span';
```
jacobdevera, line 25: I was hesitant to use the `Icon` component because it uses a `span` instead of `i`. Should we go back to fundamental-styles so that all icons use an `i` as in nui-widgets?

jbadan: After the discussion this morning, I think they're in the process of moving all icons to their own `<i>` tag- https://github.com/SAP/fundamental-styles/pull/1604

We can probably just go ahead and use `i` tags since it doesn't cause any styling issues and then styles will catch up. Up to you though

src/Button/__stories__/Button.stories.js
```diff
@@ -53,7 +53,7 @@ export const types = () => (
 
 types.storyName = 'Types';
 
-/** Button can have an icon with text or just and icon. */
+/** Button can have an icon with text or just an icon. Be sure to use an `aria-label` if there is no text. */
```
prsdthkr, line 57: Can we add aria-label to the
```jsx
<Button glyph='alert' option='emphasized' />
```
in this file.

src/Icon/Icon.js
meganaconley: 👍 If we could use the `Icon` component and use `i` tags to get ahead of it where possible that would save us time down the line.

src/Button/Button.js
```diff
@@ -76,6 +83,18 @@ const Button = React.forwardRef(({
...
     };
+
+    const renderIcon = () => {
+        const content = glyph ? (
+            <Icon
+                {...iconProps}
+                ariaHidden
+                glyph={glyph} />
+        ) : null;
+
```
jbadan, line 94: Does it work to just do something like:
```
const renderIcon = () => {
return ( glyph && <Icon props here />
)
```
}

jacobdevera: lint complains `Declare only one React component per file`

Not sure if it's cleaner to disable it / if we want to 🤷‍♂️

jbadan: Eh, I think it's fine the way you have it

jbadan (approved):
Sorry, I didn't realize I had left this blocked

PR #1218, opened 2020-09-22
feat: SearchInput as a functional comp with forwarded ref

### LINKED ISSUE #956: SearchInput: Missing forwardRef support
### Description
SearchInput component is not forwardRef style (similar to Table component), which causes searchList not to be lazy loaded or updated after SearchInput component is rendered. 

Adding example where props are updated later but SearchInput state values are not updated and having their default values.

image

### Versions
**fundamental-react:** 0.8.1

prsdthkr:
### Description
In this change, we
* convert the SearchInput component from a class component to a functional component with forwarded reference
* update tests and snapshots
* add story for substring search input

fixes #956


jbadan (approved):
LGTM 🚢

PR #1220, opened 2020-09-24
feat:  Support CSS modules

skvale:
### Description

Support loading fundamental-react with css-modules as a way to avoid style collisions from multiple instances of the library loaded onto a single page.

Some of the odd class name changes are necessary because fundamental-styles is referencing classes outside of the module scope, such as using `fd-bar` in `dialog.css` or `.token` in `tokenizer.css`

For QA the image_snapshots aren't actually using the css-module styles.  I'd like them to, but that'd make a really big diff in the storysnaps.

**Caveat**

_SideNavList has a class `'level-${level}'` and css selectors in fundamental-styles like 
```css
.fd-nested-list [class*=level]
```
this is looking for a class that contains `level`.  This doesn't work of you've hashed the `level-` class into something that doesn't contain `level` ... so a localIdentName that doesn't have `[local]` (contain the original class in addition to a unique hash) will lose styles.  

This happens in a number of other places in fundamental-styles, here's a few examples:
```sass
[class*=fd-avatar]
.fd-button > [class*=sap-icon]
.fd-calendar__action [class*=sap-icon]
.fd-carousel__button.fd-button[dir=rtl] > [class*=sap-icon]
.fd-product-switch .fd-button.fd-product-switch__control > [class*=sap-icon]
.fd-input-group__addon span[class*=sap-icon--]
.fd-button-split > [class*="--emphasized"]
```

Linking: https://github.com/SAP/fundamental-styles/issues/1711

fixes #issueid


src/Switch/__stories__/__snapshots__/Switch.stories.storyshot
```diff
@@ -212,15 +212,15 @@ exports[`Storyshots Component API/Switch Internal Labels 1`] = `
...
-            className="fd-fd-switch__text"
+            className=""
           />
           <span
             className="fd-switch__handle"
             role="presentation"
           />
           <span
             aria-label="off"
-            className="fd-fd-switch__text"
```
skvale, line 223: the class `fd-fd-switch__text`.  I'm not sure this is meant to be `fd-switch--text` either, to preserve this PR only being about css-modules I'm just planning on taking it out.

jbadan:
Can we add something to the readme to explain to consumers how they can enable this if they want it?

jbadan (approved):
💪  🥇

PR #1221, opened 2020-09-24
fix: datepicker add list styles

prsdthkr:
### Description
DatePicker component needs list styles because it uses the `fd-list--has-message` class.

### DIFF
--- a/src/DatePicker/DatePicker.js
+++ b/src/DatePicker/DatePicker.js
@@ -14,6 +14,7 @@ import { DATEPICKER_TODAY_ACTIONS_TYPES, FORM_MESSAGE_TYPES } from '../utils/con
 import React, { Component } from 'react';
 import 'fundamental-styles/dist/dialog.css';
 import 'fundamental-styles/dist/bar.css';
+import 'fundamental-styles/dist/list.css';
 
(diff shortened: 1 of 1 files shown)


src/DatePicker/DatePicker.js
```diff
@@ -14,6 +14,7 @@ import { DATEPICKER_TODAY_ACTIONS_TYPES, FORM_MESSAGE_TYPES } from '../utils/con
 import React, { Component } from 'react';
 import 'fundamental-styles/dist/dialog.css';
 import 'fundamental-styles/dist/bar.css';
+import 'fundamental-styles/dist/list.css';
```
jbadan, line 17: Gross - isn't the list file huge? Understandable though

jbadan (approved):

prsdthkr:
@jbadan Yeah since there are no styles defined for [these use cases](https://github.com/SAP/fundamental-styles/issues/1579), I had to [resort to reusing](https://github.com/SAP/fundamental-react/pull/1196) the list styles.

jbadan:
@prsdthkr Totally reasonable! I was just being a hater 😆

PR #1222, opened 2020-09-25
feat: make shellbar popovers customizable

### LINKED ISSUE #919: Accept "Placement" props in Shellbar ProfileMenu/ProductMenu
### Description
Shellbar component has default placement of "right" for ProfileMenu popover and placement "bottom-end" for ProductMenu popover. 

There should be "placement" prop which accepts popover placement value and reflect that in Menu component of Shellbar.

### Versions
**fundamental-react:**
0.8.0
---

_**DISCLAIMER:**

prsdthkr:
### Description
 
In this change, we update the Shellbar component to accept props to be spread to the Popovers of the action menu, collapsed mobile menu, notifications menu, product menu, product switch popover, and the profile menu. This will allow consumers to customize these; for example, set the `placement`.

fixes #919

### DIFF
--- a/src/Shellbar/Shellbar.js
+++ b/src/Shellbar/Shellbar.js
@@ -112,6 +112,7 @@ class Shellbar extends Component {
             productSwitchList,
             profile,
             profileMenu,
+            popoverPropsFor,
             className,
             backAction
         } = this.props;
@@ -143,6 +144,7 @@ class Shellbar extends Component {
                     {productTitle && !productMenu && <span className='fd-shellbar__title'>{productTitle}</span>}
                     {productMenu && (
                         <Popover
+                            {...popoverPropsFor?.productMenu}
                             body={
                                 productMenu && (
                                     <Menu>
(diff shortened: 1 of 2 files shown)


src/Shellbar/Shellbar.js
```diff
@@ -490,6 +497,18 @@ Shellbar.propTypes = {
...
+     * product switch popover and the profile menu.
+     * */
+    popoverPropsFor: {
+        actionMenu: PropTypes.object,
+        collapsedMobileMenu: PropTypes.object,
+        notifications: PropTypes.object,
+        productMenu: PropTypes.object,
+        productSwitch: PropTypes.object,
+        profileMenu: PropTypes.object
+    },
```
skvale, line 511: How about separating each of them out?
```
Shellbar.PropTypes = {
        actionMenuPopoverProps: PropTypes.object,
        collapsedMobileMenuPopoverProps: PropTypes.object,
        notificationsPopoverProps: PropTypes.object,
        productMenuPopoverProps: PropTypes.object,
        productSwitchPopoverProps: PropTypes.object,
        profileMenuPopoverProps: PropTypes.object
},
```

prsdthkr: Yeah, I started out with that but thought there are too many popovers. So I think packaging them would be cleaner, no?

skvale: Do you think we would need to make other changes to the popovers or just placement overrides?

prsdthkr: We might need to customize some cosmetic props like `noArrow`, `widthSizingType` or add `popperClassName` and other stuff. It would be a little more future proof.

jbadan (approved):

PR #1224, opened 2020-09-26
feat: datepicker improvements

prsdthkr:
### Description

In this change, we update the Datepicker
* to accept and set a `defaultValue` when `enableRangeSelection` is set. For example, `defaultValue='01/01/2019 - 01/01/2020'`
* add story for above use case
* to allow spreading props into the `<InputGroup.Addon>`, footer, and footer `<Button>`

## GIF
fd-react-datepicker-range-default


jbadan (approved):
LGTM 🚢

PR #1228, opened 2020-09-30
fix: focus manager use with grid manager

prsdthkr:
### Description
In this change, we fix an issue with the focus manager which prevented focus from traveling into a tabbable element positioned after a focus-manged grid inside a popover. Using a grid, with controls in its cells, inside a popover (DatePicker) caused tabbable nodes expected by the focus manager to be changed by the grid manager. This is seen in Datepicker where it was not possible to navigate to the today footer button with keyboard tabbing.

## Before
fd-react-datepicker-tabing

## After
fd-react-datepicker-tabing-fixed

### DIFF
--- a/src/utils/focusManager/focusManager.js
+++ b/src/utils/focusManager/focusManager.js
@@ -15,7 +15,7 @@ export default class FocusManager {
     }
 
     isFocusContained = (e) => {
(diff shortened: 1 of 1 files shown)


jacobdevera (approved):
LGTM

PR #1231, opened 2020-10-03
feat: row navigation in Table

jacobdevera:
### Description
- adds row-level navigation to Table (inspired by https://openui5.hana.ondemand.com/entity/sap.m.Table/sample/sap.m.sample.TableVerticalAlignment)
    - Up and down arrow keys to navigate rows
    - Space to select a row with `onSelectRow`
    - Enter to click the row with `onClickRow`
- replace `richTable` prop with a `selection` prop containing `onClickRow`, `isSelected`, and `onSelectRow` props
- adds screen reader instructions for row navigation

### Known Issues
- `:active` state styling interferes with selected row styling when selecting a row with the mouse. Will need to be fixed in `fundamental-styles`.
- `:active` styling does not trigger on enter key, only on mouse click (https://stackoverflow.com/questions/32467397/how-to-trigger-the-active-pseudoclass-on-keyboard-enter-press-using-only-cs). This is native behavior.


src/Table/Table.js
```diff
@@ -58,81 +68,149 @@ const Table = React.forwardRef(({ headers, tableData, className, compact, conden
...
+
+            case 'keydown':
+                const { row, col } = cell;
+                const key = event.which || event.keyCode;
+
+                const navigatedHorizontally = (key === keycode.codes.left || key === keycode.codes.right) && col > 0;
+                const navigatedVertically = (key === keycode.codes.up || key === keycode.codes.down) && row > 0;
+
+                if (navigatedVertically) {
+                    newInstructionsText += `${localizedText.row} ${row} `;
```
meganaconley, line 164: I think this string needs to have the row number as a variable rather than concatenated. This wouldn't be grammatical in some other languages. Same for column.

src/Table/Table.js
```diff
@@ -58,81 +68,149 @@ const Table = React.forwardRef(({ headers, tableData, className, compact, conden
...
+
+        let newInstructionsText = '';
+
+        switch (event.type) {
+            case 'focus':
+                if (instructionsText.length === 0) {
+                    if (keyboardNavigation !== 'row') {
+                        newInstructionsText += `${localizedText.arrowKeys} `;
+                    } else {
+                        newInstructionsText += `${localizedText.rowArrowKeys} ${localizedText.rowSelection} ${localizedText.rowClick}`;
```
meganaconley, line 151: Similar comment here about concatenating localized strings.

src/Table/Table.js
jacobdevera: Decided to remove the row and column numbers as screen readers still give that information anyway when using the built-in table navigation keys (ctrl+option/alt+arrow on VO and JAWS). And it doesn't seem prudent to try and work in a localization system that supports variables right now.

src/Table/Table.js
jacobdevera: Decided to combine the row instructions into a single prop to prevent concatenation.

meganaconley (approved):
Localized text changes looks good and keyboard handling works well

PR #1234, opened 2020-10-05
feat: Hide the DatePicker button when readOnly is true

skvale:
### Description

A readOnly DatePicker field should not have a button

Screen Shot 2020-10-05 at 3 13 55 PM

Breaking Change: A readOnly DatePicker field does not have an InputGroup Button

IE11
Screen Shot 2020-10-06 at 8 46 22 AM

### DIFF
--- a/src/DatePicker/DatePicker.js
+++ b/src/DatePicker/DatePicker.js
@@ -401,8 +401,6 @@ class DatePicker extends Component {
             footerButtonProps?.className
         );
 
-        const disableButton = disabled || readOnly;
-
         const inputGroup = (
             <InputGroup
                 {...inputGroupProps}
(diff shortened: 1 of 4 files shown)


jbadan:
I hate to make this change more complicated, but do we need to do this in fundamental styles too?

jbadan (approved):

bcullman:
> I hate to make this change more complicated, but do we need to do this in fundamental styles too?
> 
> Edit: there is no readOnly example in fundamental-styles, so we're safe to proceed here.

I came here to ask about this too:

I see it is not documented there, but what is the Fiori3 stance on what readonly fields should look like?

jbadan:
@bcullman It's not defined anywhere in the spec - DatePicker documentation mostly centers around the Calendar in the popover with very little about the actual input field.

skvale:
Are we OK merging this and picking that up as a conversation in a Thursday meeting?

bcullman:
> @bcullman It's not defined anywhere in the spec - DatePicker documentation mostly centers around the Calendar in the popover with very little about the actual input field.

if it's not documented, why do we think the solution is to *hide* the add-ons? I get that this is *A* possible solution, but I am not certain it is the *right* one. We need someone from UX to weigh in/followup on this

skvale:
We talked with Brandon and he thought it made sense to go with this change as it currently is.

PR #1239, opened 2020-10-10
feat: Custom CSS namespace

greg-a-smith:
### Description
Although all the components come with styling (CSS files from `fundamental-styles`), some consumers of this library want/need to provide custom styles.  It can be pretty painful to create styles that override all the default styling from `fundamental-styles` so this will allow consumers to write their own CSS.  Instead of messing with conditional imports of CSS files, I created a `withStyles` HOC and a `CssNamespaceContext` that all components now use.

By default, all components use the `fd` namespace for CSS classnames and will continue to do so with no changes from the consumer.  A new prop has been exposed on all components (`cssNamespace`) to allow the consumer to change the default namespace for CSS classnames.  The remainder of the CSS classnames will be unchanged.  Since all the components use the `CssNamespaceContext`, only top-level components need this new prop.  Child components will automatically get the same value via the context.

Here is a quick example:
```jsx
<Menu>
    <Menu.List>
        <Menu.Item url='#'>
            Option 1
        </Menu.Item>
        <Menu.Item url='#'>
            Option 2
        </Menu.Item>
    </Menu.List>
</Menu>
```
will output the following markup: (using CSS modules)
```html
<nav class="fd-menu-08add8">
  <ul class="fd-menu__list-ebf9a4" role="menu">
    <li class="fd-menu__item-00a544" role="presentation">
      <a class="fd-menu__link-0ca6e5" href="#" role="menuitem">
        <span class="fd-menu__title-7bc5d8">Option 1</span>
      </a>
    </li>
    <li class="fd-menu__item-00a544" role="presentation">
      <a class="fd-menu__link-0ca6e5" href="#" role="menuitem">
        <span class="fd-menu__title-7bc5d8">Option 2</span>
      </a>
    </li>
  </ul>
</nav>
```

By adding the `cssNamespace` prop:
```jsx
<Menu cssNamespace='xxx'>
    <Menu.List>
        <Menu.Item url='#'>
            Option 1
        </Menu.Item>
        <Menu.Item url='#'>
            Option 2
        </Menu.Item>
    </Menu.List>
</Menu>
```
the markup changes to: (CSS modules would need to be defined by the consumer)
```html
<nav class="xxx-menu">
  <ul class="xxx-menu__list" role="menu">
    <li class="xxx-menu__item" role="presentation">
      <a class="xxx-menu__link" href="#" role="menuitem">
        <span class="xxx-menu__title">Option 1</span>
      </a>
    </li>
    <li class="xxx-menu__item" role="presentation">
      <a class="xxx-menu__link" href="#" role="menuitem">
        <span class="xxx-menu__title">Option 2</span>
      </a>
    </li>
  </ul>
</nav>
```

I also added a "No Styles" storybook example for each component.


jbadan:
@greg-a-smith Hope you don't mind, but I did the following:
1.  updated LayoutGrid's Col component to also use the `cssNamespace` prop for the responsive sizes - previously we were getting `<div class='xxx-col fd-col--6'>` now `<div class='xxx-col xxx-col--6'>`
2. edited the storybook custom css overrides so that the docs page for LayoutGrid still has the colors - I think was possibly lost during css modules, but regardless, works now
3. updated storyshots/snapshots
4. fixed netlify build by clearing cache - it's due to upgrading to node 12. Hopefully this will resolve itself soon or else I want to go back to node 10.21.0...

BEFORE:
Screen Shot 2020-10-11 at 9 09 21 AM
Screen Shot 2020-10-11 at 9 10 27 AM

AFTER:
Screen Shot 2020-10-11 at 9 21 03 AM
Screen Shot 2020-10-11 at 9 20 53 AM

greg-a-smith:
@jbadan I don't mind at all -- thank you!  I see the build is still failing with 2 snapshot failures and I can't figure out how to fix that.  I don't see any failures when I run it locally. 🤷‍♂️

jbadan (approved):
LGTM  - huge undertaking! I adjusted visual tests to allow size mismatch and tests are now passing 🚢

greg-a-smith:
@jbadan Thanks for getting this across the 🏁 line!

PR #1240, opened 2020-10-12
fix: Fix breaking datepicker test with merge

skvale:
### Description
Fix the failing build on the master branch after I merged the last PR

fixes #issueid

### DIFF
--- a/src/DatePicker/DatePicker.test.js
+++ b/src/DatePicker/DatePicker.test.js
@@ -658,7 +658,7 @@ describe('<DatePicker />', () => {
 
         test('should disabled the popover when readOnly is true', () => {
             wrapper = mount(<DatePicker readOnly />);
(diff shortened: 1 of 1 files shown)


jbadan (approved):
Should we add a branch setting that requires  your branch to be up to date with master before merge?

🚢

skvale:
> Should we add a branch setting that requires your branch to be up to date with master before merge?

Has it been a problem before?  My thought process was: "I should probably merge the latest in, but I don't want to wait another 4 hours for my build to finish" so, to me, the root-cause was painfully slow Travis builds

jbadan:
@skvale Not sure what to tell you 🤷  We're trying to switch over to travis.com (waiting on red tape of course), but it's also going to be painfully slow as soon as those other repos that hog the builds move over there.

PR #1241, opened 2020-10-13
fix: gridManager cleanup, fixes

jacobdevera:
### Description
- clean up gridManager. The `grid` used to store only HTML elements, but now stores an object with all needed cell properties (row, column, element, focusable elements, editable element) to prevent unnecessarily recalculating properties
- fix some selectors in Calendar to fix keyboard navigation


prsdthkr (approved):
QA LGTM ⌨️

PR #1242, opened 2020-10-15
feat: new innerRefClassName prop as part of validationOverlayProps

jbadan:
### Description
This is useful for passing an additional class from nui-widgets to the wrapper div inside popper.


src/utils/_Popper.js
```diff
@@ -213,7 +213,7 @@ class Popper extends React.Component {
                             x-out-of-boundaries={isReferenceHidden ? 'true' : undefined}
                             // This is needed for fundamental-styles even though popper-2 uses data-placement as well
                             x-placement={placement}>
-                            <div ref={innerRef}>
+                            <div className={classnames(`${cssBlock}__innerRef`, popperProps?.innerRefClassName)} ref={innerRef}>
```
prsdthkr, line 216: Should `innerRefClassName` be a separate prop? It doesn't seem to be a part of popper.js. Which is what we [document](https://github.com/SAP/fundamental-react/blob/select-div/src/Popover/Popover.js#L251) `popperProps` as, right?

jbadan: Yeah, you're right... I'm not sure what I was thinking. Maybe I had popperProper and popoverProps confused 🤷‍♀️  I'll change this. Thanks!

skvale (approved):
LGTM

PR #1244, opened 2020-10-16
feat: Make Calendar disabledDates an array of single dates, use disabledDateRanges for existing functionality

skvale:
### Description

Calendar disabledDates were treated as ranges in fundamental-react but documented as single dates that are disabled in our internal component library that consumes this.
Breaking changes
- remove blockedDates -- These can be replicated with disabledDateRanges
- treat disabledDates as an array of single dates that are disabled -- previous behavior can be replicated with disabledDateRanges

New additions
- new prop to Calendar and DatePicker: disabledDateRanges which is an array of two-tuples, everything between the tuples(inclusive) are disabled dates.  I went with inclusive because the concept of date ranges is already inclusive
- the propType for dates was expanded to almost anything that a moment constructor takes, since they should all go through the moment constructor anyways
- when we use the moment constructor on something that may not be a moment, we need to include the dateFormat or we'll get unexpected dates when a consumer is using 'YYYY.DD.MM' or something

fixes #issueid


prsdthkr (approved):
QA LGTM 👍

skvale:
@prsdthkr Do you think we should get more eyes from the accessibility team for this change?

prsdthkr:
@skvale this seems to be more of an API change than markup changes... so we should be good. This will go down as a breaking change right?

skvale:
yes, this will be a breaking change.  

It looks like we could get 17 rc versions into 0.13.1 if we released now and leave this for 0.14.0-rc.0

PR #1245, opened 2020-10-16
chore: revert to node 10

jbadan:
### Description
I'm tired of having to rebuild netlify for every commit on every pull request. Let's hang out on 10 until we have time to upgrade properly.

fixes #issueid


greg-a-smith (approved):
Looks good. 🚢

PR #1247, opened 2020-10-21
docs: add base font size to html in readme instructions

### LINKED ISSUE #1236: Docs: add configuration note to readme about adding base font size
Because our styles are based in rem, consumers should be adding 
```
html {
    font-size: 16px;
}
```

to their projects to get the correct styling.

jbadan:
### Description
css from fundamental-styles uses rems. 1rem equals the font size of the html element. While most browsers have 16px as the base, we might want to explicitly call it out. I'm not sure if this pr is necessary 🤷‍♂️ 

fixes #1236

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,14 @@ For an existing react application, follow the steps below:
 
 Additionally, edit your webpack configuration to load font and icon fonts - see [file-loader](https://webpack.js.org/loaders/file-loader/).
 
+1. All styles are based on `rem` units. Include the following in your css to ensure components are sized correctly:
+
+```css
+    html {
+        font-size: 16px;
+    }
+```
+
 ## Versioning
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -75,6 +75,14 @@ For an existing react application, follow the steps below:
 
 Additionally, edit your webpack configuration to load font and icon fonts - see [file-loader](https://webpack.js.org/loaders/file-loader/).
 
+1. All styles are based on `rem` units. Include the following in your css to ensure components are sized correctly:
```
prsdthkr, line 78:
```suggestion
All styles are based on `rem` units. Include the following in your CSS to ensure components are sized correctly:
```

prsdthkr (approved):
Minor nitpick. Else, LGTM.

PR #1248, opened 2020-10-21
feat: update icon list with new icons

### LINKED ISSUE #1226: List.Icon doesn't fully support fundamental icons
Dear developers,
When I use List.Icon within a List.Item and set glyph="text", which is part of fundamental icons from the list below, I got a warning in console that says "Failed prop type: Invalid prop `glyph` of value `text` supplied to `List.Icon`, expected one of [_An array of 621 items_]"
But according to fundamental icons, there're 652 icons supported. 
https://sap.github.io/fundamental-styles/components/icon.html
Could you please kindly check the gap?

jbadan:
### Description
Update our icon list with all the new icons - list from https://github.com/SAP/fundamental-styles/blob/master/stories/icon/data.json

fixes #1226


prsdthkr:
@jbadan Do we need to audit if we were using any removed icons in the components?

jbadan:
@prsdthkr That didn't even occur to me. Yes, we should. I'll convert this to a draft.

jbadan:
@prsdthkr Once I sorted the icons, I realized none are actually removed. There's just additions

prsdthkr (approved):

PR #1260, opened 2020-11-06
fix: InputGroup validation changes

### LINKED ISSUE #1207: Combobox with validation states is broken
### Description
Entering text input when combobox has a validation state set is wonky:

### Versions
**fundamental-react:** v0.13.0-rc.19

jacobdevera:
### Description
BREAKING CHANGE
Refactor `_FormValidationOverlay` in `InputGroup` and related components 
- `InputGroup` will now always render with a `_FormValidationOverlay` regardless of whether a validation state is present or not. This prevents the component from throwing away the old `input` element and losing the previous state when the validation state changes.
    - fixes an issue with `ComboboxInput` that makes it impossible to enter input
    - fixes an issue with `DatePicker` that would cause it to lose focus when closing the `Calendar` with the `Esc` key
- only show `_FormValidationOverlay` if `text` is present in `validationState` with a valid string
- remove `validationOverlayProps` from `SearchInput` as this would now be handled by passing them into `inputGroupProps`
- add `formMessageProps` to `SearchInput` to handle passing props into its own `FormMessage`

fixes #1207


meganaconley (approved):
Tested locally, inputs no longer lose focus after validation.

Mike-Diaz (approved):

PR #1266, opened 2020-11-18
feat: update fundamental-styles to 0.13.1-rc.2

meganaconley:
BREAKING CHANGE

### Description
This update includes:

- `Button` no longer has `selected` attribute
- `Tab` aria attributes and role moved from the `a` to the `li`
- Tab-index adjusted on `Breadcrumb` and `role="menu"` added
- Added labels to docs page for better accessibility on examples
- Fixed other accessibility concerns on stories, mostly around `aria-labels` for icons
- Other minor css adjustments that did not require markup changes


src/Button/__stories__/ButtonSegmented.stories.js
```diff
@@ -9,9 +9,12 @@ export default {
...
 
 export const primary = () => (<ButtonSegmented>
-    <Button glyph='survey' />
-    <Button glyph='pie-chart' selected />
-    <Button glyph='pool' />
+    <Button aria-label='Survey' glyph='survey' />
+    <Button
+        aria-label='Pie chart'
+        glyph='pie-chart'
+        selected />
```
jacobdevera, line 16: `selected` was removed

src/Breadcrumb/_BreadcrumbItem.js
```diff
@@ -10,11 +10,17 @@ const BreadcrumbItem = ({ url, name, className, children, cssNamespace, ...props
     const renderLink = () => {
         if (!children && url) {
             return (
-                <a className={classnames(`${cssNamespace}-breadcrumb__link`)} href={url}>{name}</a>
+                <a
+                    className={classnames(`${cssNamespace}-breadcrumb__link`)}
+                    href={url}
+                    tabIndex='0'>
```
jacobdevera, line 16: Is there a reason we need to explicitly set a `tabindex` for an `a`?

jacobdevera (approved):
One q, otherwise looks good in QA

src/Button/__stories__/ButtonSegmented.stories.js
meganaconley: It was removed as an attribute of the underlying `<button>` but not as a prop. It still is used to apply a classname.

src/Breadcrumb/_BreadcrumbItem.js
meganaconley: Removing it seems to still work, but I think it's safest to keep it in case there are some browser/screen reader combos that might not tab to these links.

src/Select/__stories__/Select.stories.js
```diff
@@ -23,25 +23,31 @@ export default {
...
 export const disabled = () => (
     <Select
+        aria-label='Disabled'
         disabled
         options={options}
         placeholder='Select' />
 );
 export const readOnly = () => (
     <Select
+        aria-label='readOnly'
```
Mike-Diaz, line 50: Wasn't sure how the screen reader would sound with the concatenated word `readOnly` but it sounds no different than if it were "Read Only" 👍

Mike-Diaz (approved):
QA with VoiceOver, all new `aria-labels` are awesome! 
Everything else works as expected!

PR #1290, opened 2021-01-20
fix: Select selectedKey behavior

prsdthkr:
### Description
Fixes an issue where the select component did not set the `selectedKey` when new options were set.
This change should cause select to set the `selectedKey` when it or `options` prop change


hertweckhr1 (approved):
LGTM 🎉

jbadan (approved):

PR #1340, opened 2021-07-16
fix: allow lock focus for specific list 

### LINKED ISSUE #1337: Can you support white list node as a new property of Dialog and pass it form dialog to this Focus Lock components
https://github.com/SAP/fundamental-react/blob/0950399f5b31202bc2923faeec9ccb3847355fb0/src/Dialog/Dialog.js#L105-L106

hi, current situation from my side, is our application is render as a part of iframe application. inside iframe, and if fundamental dialog open inside our iframe, the host app will also lost focus,  due to this focus lock will only get focus in side dialog area. so can you support white list, than I can pass our app's root elements as lock's  lock area, then it wouldn't affects other host application.

> whiteList=fn you could whitelist locations FocusLock should carry about. Everything outside it will ignore. For example - any modals.

https://codesandbox.io/s/react-focus-lock-forked-mr03y?file=/index.js:277-345

Vita-Meow:
### Description

fixes #1337

### DIFF
--- a/src/Dialog/Dialog.js
+++ b/src/Dialog/Dialog.js
@@ -103,8 +103,9 @@ class Dialog extends Component {
         }
 
         return ReactDOM.createPortal(
-            <FocusLock as='div' className={backdropClassName} whiteList={allowLockFocusList}
-                lockProps={{ ...rest }}>
+            <FocusLock as='div' className={backdropClassName}
+                lockProps={{ ...rest }}
+                whiteList={allowLockFocusList}>
                 <span data-autofocus tabIndex='-1' />
                 <div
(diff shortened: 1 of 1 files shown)


src/Dialog/Dialog.js
```diff
@@ -102,7 +103,7 @@ class Dialog extends Component {
         }
 
         return ReactDOM.createPortal(
-            <FocusLock as='div' className={backdropClassName}
+            <FocusLock as='div' className={backdropClassName} whiteList={allowLockFocusList}
```
Vita-Meow, line 106: This white list is defined in Focus lock, I can't help it

PR #1352, opened 2021-08-06
fix: not applying defaultChecked prop for RadioItem

### LINKED ISSUE #1351: RadioItem not applying defaultChecked prop
### Description
Dear colleagues,
If you pass the defaultChecked prop to the FormRadioItem it is not applied. That happens because under the hood the prop is not passed to the FormInput. Feel free to check.
I will open a PR to fix this.

Kind regards,
Gianluca Filitz
Working Student - SAP SE

### Versions
**fundamental-react:^0.13.1**

LucaFilitz:
### Description

This line of code passes the defaultChecked prop to the input item which is required so that the prop actually has an effect.

fixes #1351

### DIFF
--- a/src/Forms/FormRadioItem.js
+++ b/src/Forms/FormRadioItem.js
@@ -46,6 +46,7 @@ const FormRadioItem = React.forwardRef(({
             <input
                 {...inputProps}
                 checked={checked}
+                defaultChecked={defaultChecked}
                 className={inputClassName}
                 disabled={disabled}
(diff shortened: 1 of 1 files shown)


droshev (changes requested):
You would need to rename your commit message respecting conventional commits

LucaFilitz (commented):
I recommited with a different commit message

LucaFilitz:
Is there any update on this?

PR #1356, opened 2021-08-12
feat: introduce MessageBox and Bar components

qbalukom:
### Description
bump fundamental-styles to 0.19
remove css modules
introduce MessageBox and Bar components


qbalukom:
Note about the latest commit - it seems switch is not suppoesed to support icons without semantic coloring nor coloring without icons. As such, i removed both from the storybook

src/MessageBox/MessageBox.js
```diff
@@ -0,0 +1,176 @@
...
+                        </>]}
+                        type='header'
+                        { ...headerProps } />
+                    <section className={bodyClasses} {...bodyProps}>
+                        {children}
+                    </section>
+                    <Bar
+                        className={classnames(`${cssNamespace}-message-box__footer`)}
+                        rightComponents={actions.map(child => cloneElement(child, {
+                            className: classnames(`${cssNamespace}-message-box__decisive-button`),
```
dariadomagala, line 126: we should use the className coming from the child and this one should be only used as a fallback.

src/MessageBox/__stories__/MessageBox.stories.js
```diff
@@ -0,0 +1,163 @@
...
+export const sizes = () => {
+    const [showMessageBoxSm, setShowMessageBoxSm] = useState(false);
+    const [showMessageBoxMd, setShowMessageBoxMd] = useState(false);
+    const [showMessageBoxLg, setShowMessageBoxLg] = useState(false);
+    const [showMessageBoxXl, setShowMessageBoxXl] = useState(false);
+
+
+    return (
+        <div className='fddocs-container'>
+            <Button onClick={() => setShowMessageBoxSm(true)}>Small</Button>
```
dariadomagala, line 113: I cannot really see the differences between those sizes

dariadomagala (changes requested):

src/MessageBox/__stories__/MessageBox.stories.js
qbalukom: They effectively only change the left/right margin size. I do myself feel they're not very useful, but fundamental-styles provides them, so...

src/MessageBox/MessageBox.js
qbalukom: i got inspired by an identical solution used in the `Dialog` component. Apparently this class is expected to be there. I could perhaps add user-provided classes so this one gets appended, rather than replaced.

dariadomagala (changes requested):
~The `Switch` component changed significantly with the newest update, and it doesn't work~
I had a wrong version of `fundamental-styles`, now it looks good.

src/MessageBox/__stories__/MessageBox.stories.js
dariadomagala: That's true... It's not what I would expect 😕

dariadomagala (approved):

PR #1364, opened 2021-09-29
chore: change the combobox input behaviour of hiding not matching entries

### LINKED ISSUE #1334: ComboboxInput with pre-selected key hides all remaining entries that is confusing for the user
### Description

If I use the ComboboxInput with the `selectionType='auto-inline'` and `selectedKey='key'` options, all entries besides the pre-selected one are hidden until the user deletes the input text. This seems to be very unintuitive since for the user it looks like he/she has only one option available:
Screen Shot 2021-07-07 at 12 19 55
when in real life he/she has more:
Screen Shot 2021-07-07 at 12 19 45
The same behavior occurs even without the `selectedKey` but not right away. When the user already selected one entry and then decides to change it - he/she cannot see other options, which might confuse him/her a lot.

My ideas would be:
* grey-out entries not matching the query but still display them 
* clear out the filter on each dropdown click (preferably with the matching entries on top)

### Versions
**fundamental-react:** 0.13.1

---

_**DISCLAIMER:**

dariadomagala:
### Description

* Don't hide entries not matching the search query by default
* Add an option to hide entries not matching the search query
* Filter entries not by `startsWith` but by `includes` (that wasn't changed for the auto-completion on a first find, that just simply wouldn't make sense)

fixes #1334


droshev:
this PR seems a `fix:` to me and not a `chore:`

qbalukom (approved):

